Issue #218, opened 2016-03-11
Implement pop-outs


KeithWoods:
Spot tiles, analytics and the blotter need to be able to pop out into a new windows. 

Note here was a previous implement of this, however it was removed in recent refactors. It needs to be re-implemented with a clean abstraction, and such an abstraction should work with the openfin integration.

LeeCampbell:
In the mean time, it might be worth removing the redundant popout buttons. Just looks unfinished. Probably better to not have the feature at all, and then add the buttons back in once it is completed properly.

KeithWoods:
#227 ready for review this is pretty much done

PR #227, opened 2016-03-17
GUI: Region Management & Popouts

KeithWoods:
This PR enables components (tiles, blotter, analytic) to be popped out of the app. 

To enable this a basic concept of region management is introduced. This allows for any model to put it's view into a 'region' which owns displaying of it's component in a certain way. For example as a popout. When running in a container like openfin the region can adapt it's behaviour to accommodate for that host, the models don't need to change.

Additionally this infrastructure allows container hosting components such as the shell to be decoupled from items it hosts. In a larger implementation different modules would inject different components into well known regions, the containing host (the shell) need not change. The models themselves decide where their views, and which specific view, get displayed. 

Finally it allows for a model to have one or more view representations. For example a model could add a  view representation of it self to one or more regions using a view context.


KeithWoods:
Related issue: #218

Issue #106, opened 2015-12-22
Overall application style vs old reactive trader


odeheurles:
I think the app looks too dark and David (the designer who did the initial visual design) had the good idea to put the blotter in white. Also David's design did not use any gradient and everything was flat. If you open both apps side by side you will see that the new app looks A LOT more noisy that it used to be (it looks like a lot more stuff is being displayed, which is not the case).

Here is the link to the previous version: http://adaptiveconsulting.github.io/ReactiveTraderJS/

The design was not correctly implement (the WPF is the closest one from the design), I

Overall I would be in favour of bringing the design closer to what it was before in a few aspects:
 - the layout of the tile is fine I think (better than in the previous reactive trader web version), but the spark line adds too much noise, I would go crazy if I had to look at this all day. I would suggest to remove them
 - the mouse over should not use gradients, let's use the same style as before
 - remove the animation on trade done and remove the gradients on the trade done tile
 - when the tile is in error state use the same style as before (orange tile), see EURJPY trade on [old version](http://adaptiveconsulting.github.io/ReactiveTraderJS/)
 - I think we also should style the blotter as it was before (white instead of dark)

LeeCampbell:
+1

LeeCampbell:
I agree with the points above, but also think that the current font needs to change to look both modern and professional. The current font selection not only (IMO) looks like a font from a Commodore 64 computer game it is internally inconsistent. Prices are in a Serif font, while all other text is in a san-serif font

### Example of live price in serif
![image](https://cloud.githubusercontent.com/assets/981443/13723397/8ccbb736-e89e-11e5-82fc-8b38ce63d047.png)

### Example of traded price in sans-serif
![image](https://cloud.githubusercontent.com/assets/981443/13723401/9eefbed0-e89e-11e5-86be-7276c2e1a2b5.png)

Spacing needs to be looked at. 
![image](https://cloud.githubusercontent.com/assets/981443/13723409/1afb20c8-e89f-11e5-8f06-8c8ade2cb777.png)

I do like the designers font selection, however I believe that a lot of thought went into the https://reactivetrader.azurewebsites.net/ font as it was a sensible choice for cross platform support. It believe it was chosen to still look great (which is does) and also be consistent across platforms.

KeithWoods:
A note to the dev that implements this: 

[...]

PR #241, opened 2016-04-14
Add Footer

fdore:
#106


KeithWoods:
nice work. will merge when the build goes green

Issue #133, opened 2015-12-27
Client should be able to connect to the cloud


odeheurles:
It would be a lot easier for client side developers to be able to run the client and connect it to the services running on the cloud.

Tasks:
 - configure CORS
 - update the documentation (people interested only by the client should know that this is available)
 - add a npm command to launch the client and connect it to the cloud

KeithWoods:
`npm start` (the default use case) should default the backend to the cloud. It should be an edge case to point to your local environment. 

I suggest we have these targets:
* `npm run dev:local`  
* `npm run dev:cloud` 

Point `npm start` to `npm run dev:cloud`

tdeheurles:
@keith these changes will impact the automated deployment. Just let me know if/when you add it please.

KeithWoods:
@bnwan is this now done?

bnwan:
@KeithWoods this has been merge into the dev branch

PR #243, opened 2016-04-14
#133 : added configs to support dev cloud and dev local endpoints

bnwan:


src/client/config/dev-cloud.config.json
```diff
@@ -0,0 +1,4 @@
+{
+  "useEndPoint": true,
```
raybooysen: When would this be false?

src/client/webpack.config.js
```diff
@@ -6,8 +6,11 @@ const CopyWebpackPlugin = require('copy-webpack-plugin');
 const ExtractTextPlugin = require('extract-text-webpack-plugin');
 const isProductionMode = process.env.NODE_ENV == 'production';
 const chalk = require('chalk');
-
 const path = require('path');
+const parseArgs = require('minimist');
```
raybooysen: does this need a package.json update?

src/client/config/dev-cloud.config.json
KeithWoods: I believe in the case of server deployment. There are cases when we want to use window.location, otherwise we have to update the deployment so we write out config for each deployment we push to

src/client/webpack.config.js
bnwan: oops...sorry

src/client/config/dev-cloud.config.json
bnwan: when you want to default to the same hostname as the application, i.e. location.hostname

KeithWoods: perhaps a better name would be `overwriteServerEndpoint`?

raybooysen: Why don't we have a server deployment config?

src/client/webpack.config.js
raybooysen: and a shrinkwrap. :)

src/client/webpack.config.js
```diff
@@ -53,10 +56,11 @@ const webpackConfig = {
     // This is purely for ide object/type discoverability.
     // Until our ide (intellij/webstorm) understands import aliass we feel the benefits of object discoverability outweigh the relative path cost.
     alias: {
-      system: __dirname + '/src/system',
-      services: __dirname + '/src/services',
+      config: path.join(__dirname, 'config', config),
```
KeithWoods: this should probably be called `config.json`

src/client/config/dev-cloud.config.json
KeithWoods: we can, but the url  changes based on the build number so we'd have to update the deployment to write out this file.

KeithWoods: I think an 'overwrite' concept seems reasonable for now and we keep that closer to the dev workflow.

PR #245, opened 2016-04-14
use web font matching designs

qooroo:
Part of #106 

This font is the closest matching one to the visual designs we have out of the available Google web fonts. It's very subtly different - only the '0's are a bit thinner and the end of '3's are a tiny bit longer.

The proprietary design font probably doesn't conform to the FOSS style of this project anyway.

Rubik (Google web font):
image

### DIFF
--- a/src/client/src/index.html
+++ b/src/client/src/index.html
@@ -4,6 +4,7 @@
   <title>Reactive Trader Cloud</title>
   <meta charset="utf-8">
   <link rel="shortcut icon" type="image/x-icon" href="images/favicon.ico"/>
+  <link href='https://fonts.googleapis.com/css?family=Rubik:300' rel='stylesheet' type='text/css'>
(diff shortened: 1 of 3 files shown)


src/client/src/index.html
```diff
@@ -4,6 +4,7 @@
   <title>Reactive Trader Cloud</title>
   <meta charset="utf-8">
   <link rel="shortcut icon" type="image/x-icon" href="images/favicon.ico"/>
+  <link href='https://fonts.googleapis.com/css?family=Rubik:300' rel='stylesheet' type='text/css'>
```
KeithWoods: is it possible to pull it down locally? I've noticed some issues sometimes when reloading when you have to hit the web for such resources.

qooroo:
@KeithWoods yup - see #247

PR #249, opened 2016-04-14
Layout improvement

fdore:
Make layout responsive. Tiles are now stretching and take all the available space in the workspace area. This also means we don't have a gap between the workspace and the analytics component anymore.

#106

screen shot 2016-04-15 at 11 20 51

### DIFF
--- a/src/client/src/ui/regions/views/workspace/workspaceRegionView.jsx
+++ b/src/client/src/ui/regions/views/workspace/workspaceRegionView.jsx
@@ -38,6 +38,6 @@ export default class WorkspaceRegionView extends ViewBase {
           {View}
         </WorkspaceItemContainer>
       );
-    });
+    }).concat(_.times(3, () => <div className='workspace-region__spacer'></div>)); // add empty items at the end so tiles lay out nicely
   }
 }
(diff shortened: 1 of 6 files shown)


src/client/src/ui/regions/views/workspace/workspaceRegionView.jsx
```diff
@@ -38,6 +38,6 @@ export default class WorkspaceRegionView extends ViewBase {
           {View}
         </WorkspaceItemContainer>
       );
-    });
+    }).concat(_.times(3, () => <div className='workspace-region__spacer'></div>)); // add empty items at the end so tiles lay out nicely
```
KeithWoods: this seems a bit magic. does this screw with the layout on different resolutions?

KeithWoods:
looking good

fdore:
Scrollbar is now displayed for the workspace area.

src/client/src/ui/regions/views/workspace/workspaceRegionView.jsx
fdore: Yeah, this is a bit of magic. This is so all the tiles have the same width, even when a row is not full, but yet, tiles can stretch.

PR #252, opened 2016-04-15
make spot tiles styling more in line with design

fdore:
#106


src/client/src/ui/spotTile/views/notionalInput.jsx
```diff
@@ -26,7 +26,7 @@ export default class NotionalInput extends React.Component {
    */
   componentWillMount(){
     const today = new Date;
-    this.SPOTDATE = ['SP.', today.getDate(), MONTHS[today.getMonth()]].join(' ');
+    this.SPOTDATE = [today.getDate(), MONTHS[today.getMonth()]].join(' ');
```
KeithWoods: this should be in the model

PR #253, opened 2016-04-15
Blotter styling

fdore:
Style blotter so its closer to original design
Add styling for scrollbars

#106

### DIFF
--- a/src/client/src/ui/blotter/views/blotter.scss
+++ b/src/client/src/ui/blotter/views/blotter.scss
@@ -5,7 +5,9 @@
   width: 100%;
   overflow: hidden;
   font-family: 'Segoe UI', sans-serif;
-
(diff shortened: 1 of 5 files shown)


src/client/src/ui/blotter/views/blotter.scss
```diff
@@ -5,7 +5,9 @@
   width: 100%;
   overflow: hidden;
   font-family: 'Segoe UI', sans-serif;
-
+  .fixedDataTableCellLayout_wrap1 {
```
bnwan: Not quite sure about the naming of this, ie. ending with "wrap1". What's the reason for this?

fdore: its to overwrite the styling of the grid we are using

bnwan: Ok cool

PR #261, opened 2016-04-19
Fix prod deployment for client

qooroo:


src/client/webpack.config.js
```diff
@@ -43,7 +42,12 @@ const webpackConfig = {
     ]),
     new webpack.optimize.CommonsChunkPlugin(/* chunkName= */'vendor', /* filename= */'vendor.js', function(module){
       return module.resource && module.resource.indexOf('node_modules') !== -1;
-    })
+    }),
+    new webpack.DefinePlugin({
+     __CONFIG__: args.endpoint
```
raybooysen: We were trying to avoid the DefinePlugin for this, why did we have to?

fdore: Whats the reason for avoiding it?

raybooysen: This is a good demo app, and for people unfamiliar with webpack and only reading the bootstrapper.js, it looks like magic that it works and not great for discovering how it all works together.  cc @KeithWoods

KeithWoods: I quote like the alias approach. 

Makes the usage more explicit. I tend to think of the define plugin in the same context as magic global variables. 

With the alias the scope of the configuration is better defined. It's still somewhat magic, but at least the scope is better defined.

fdore: Fair enough. You might still wonder where this config.json file comes from though.

raybooysen: Yeah agreed, feels like the lesser of two evils. :(

qooroo: Just from my perspective: I'm unfamiliar with webpack. In the bootstrapper I see this `__CONFIG__` thing and a search takes me to webpack config which defines it from arguments. Doesn't seem that convoluted.

Comparatively the setting of `isProductionMode` seems more complicated.

And also top result for "webpack environment variables" is this: http://stackoverflow.com/a/30061249

qooroo: eurgh, github doesn't update comments in real time..

KeithWoods: Shall we close this then without the merge. I think what we currently have is sufficient.

qooroo: Prod deployment is broken at the moment in dev because the build script runs `npm run deploy:prod` which breaks trying to find the `config.json` import. If we keep the way things are then I'll add a config file for prod which doesn't override the endpoints. It'll effectively be an empty `config.json` which gets imported into bootstrapper. Happy to go that way?

PR #265, opened 2016-04-19
refactor service status display

fdore:
#106

### DIFF
--- a/src/client/src/services/model/applicationStatusConst.js
+++ b/src/client/src/services/model/applicationStatusConst.js
@@ -1,18 +1,6 @@
-export default class ApplicationStatusConst {
-  static get Unknown():String {
-    return 'Unknown';
-  }
-
-  static get Healthy():String {
-    return 'Healthy';
-  }
-
-  static get Warning():String {
-    return 'Warning';
-  }
-
-  static get Down():String {
-    return 'Down';
-  }
-
+export default ApplicationStatusConst = {
+  Unknown: 'Unknown',
+  Healthy: 'Healthy',
+  Warning: 'Warning',
+  Down: 'Down'
 }
--- a/src/client/src/services/model/applicationStatusConst.js
+++ b/src/client/src/services/model/applicationStatusConst.js
@@ -1,0 +1,18 @@
+export default class ApplicationStatusConst {
+  static get Unknown():String {
+    return 'Unknown';
+  }
+
+  static get Healthy():String {
+    return 'Healthy';
+  }
+
+  static get Warning():String {
+    return 'Warning';
+  }
+
+  static get Down():String {
+    return 'Down';
+  }
+
+}
--- a/src/client/src/ui/footer/model/footerModel.js
+++ b/src/client/src/ui/footer/model/footerModel.js
@@ -23,6 +23,7 @@ export default class FooterModel extends ModelBase {
 
     this.serviceLookup = new ServiceStatusLookup();
     this.isConnectedToBroker = false;
+    this.shouldShowServiceStatus = false;
   }
 
   @observeEvent('init')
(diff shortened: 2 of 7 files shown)


src/client/src/ui/footer/views/footerView.jsx
```diff
@@ -16,19 +18,36 @@ export default class FooterView extends ViewBase {
 
   render() {
     let model:FooterModel = this.state.model;
+    let panelClasses = classnames(
```
KeithWoods: this this needs to go below the `if(!model)` check

src/client/src/ui/footer/views/footerView.jsx
```diff
@@ -16,19 +18,36 @@ export default class FooterView extends ViewBase {
...
+      'footer__service-status-panel',
+      {
+        'hide': !model.shouldShowServiceStatus
+    });
     if (!model) {
       return null;
     }
-
+    let healthy = model.isConnectedToBroker && _.every(model.serviceLookup.services, service => service.isConnected);
+    let down = !model.isConnectedToBroker || _.every(model.serviceLookup.services, service => !service.isConnected);
```
KeithWoods: these 2 bools could go on the model

KeithWoods:
Some minor comments, but looking good.

src/client/src/ui/footer/model/footerModel.js
```diff
@@ -3,10 +3,18 @@ import { CompositeStatusService } from '../../../services';
 import { logger } from '../../../system';
 import { ModelBase } from '../../common';
 import { ServiceStatusLookup } from '../../../services/model';
+import _ from 'lodash';
 
 var _log:logger.Logger = logger.create('FooterModel');
 
-export default class FooterModel extends ModelBase {
+export const FooterConnectionStatus = {
```
KeithWoods: can we put this in a separate file and follow the same pattern we've been using for other status? for exmaple [tradeStatus](https://github.com/AdaptiveConsulting/ReactiveTraderCloud/blob/dev/src/client/src/services/model/tradeStatus.js)

src/client/src/ui/footer/model/index.js
```diff
@@ -1 +1 @@
-export { default as FooterModel } from './footerModel';
+export * from './footerModel';
```
KeithWoods: if we move the `FooterConnectionStatus` to another class this will have to change

src/client/src/ui/footer/model/footerModel.js
```diff
@@ -31,13 +34,29 @@ export default class FooterModel extends ModelBase {
...
   _subscribeToConnectionStatus() {
     this.addDisposable(
       this._compositeStatusService.serviceStatusStream.subscribeWithRouter(
         this.router,
         this.modelId,
         (serviceStatusLookup:ServiceStatusLookup) => {
+          const services = Object.keys(serviceStatusLookup.services)
+                                 .map(key => serviceStatusLookup.services[key]);
           this.serviceLookup = serviceStatusLookup;
+          if (_.every(services, service => service.isConnected)) {
```
DimitarChristoff: seems like this can probably be done in 1 loop rather than `.every` and `.some` - though given how few services they are, it does not matter. the api supports property shortcuts so you could say `_.every(services, 'isConnected')`

src/client/src/services/model/applicationStatusConst.js
```diff
@@ -0,0 +1,18 @@
+export default class ApplicationStatusConst {
```
DimitarChristoff: should this be a simple map and not a class with static getters?

fdore: just trying to be consistent with the other models

Issue #228, opened 2016-04-08
Blotter doesn't flash when a new trade arrives


KeithWoods:
The blotter should flash (as per v1) when a new trade pops in.

KeithWoods:
#106 needs to be done first

PR #268, opened 2016-04-19
highlight new trade in blotter

fdore:
#228

### DIFF
--- a/src/client/src/ui/blotter/model/blotterModel.js
+++ b/src/client/src/ui/blotter/model/blotterModel.js
@@ -53,6 +53,12 @@ export default class BlotterModel extends ModelBase {
     this._regionManagerHelper.popout(850, 280);
   }
 
+  preProcess() {
+    this.trades.forEach((trade:Trade) => {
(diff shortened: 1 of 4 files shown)


src/client/src/ui/blotter/model/blotterModel.js
```diff
@@ -60,12 +60,16 @@ export default class BlotterModel extends ModelBase {
           this.router,
           this.modelId,
           (trades:Array<Trade>) => {
+            _.forEach(this.trades, (trade:Trade) => {
```
KeithWoods: You might want to move this into a method calle `preProcess()` ([doco])(http://esp.readthedocs.org/en/latest/advanced-concepts/complete-event-workflow.html#pre-event-processing). 
Pre processing happens before any change to the model, it allows for any transient state to be reset. 

This way any change to the model will below away the transient status `isNew`. We may not have anything else changing the blotter model ATM, however this could change if we had filtering or such things built in.

KeithWoods:
nice. will merge when the build is done

PR #275, opened 2016-04-20
display application version in splash screen

### LINKED ISSUE #266: Add Adaptive splash screen
rt splash

### LINKED ISSUE #295: Display application version on splash screen

fdore:
Fixes #295

### DIFF
--- a/src/client/webpack.config.js
+++ b/src/client/webpack.config.js
@@ -43,6 +43,9 @@ const webpackConfig = {
     ]),
     new webpack.optimize.CommonsChunkPlugin(/* chunkName= */'vendor', /* filename= */'vendor.js', function(module){
       return module.resource && module.resource.indexOf('node_modules') !== -1;
+    }),
+    new webpack.DefinePlugin({
+      __VERSION__: JSON.stringify(require(path.resolve(__dirname, './package.json')).version)
     })
   ],
   // these break for node 5.3+ when building WS stuff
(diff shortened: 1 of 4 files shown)


src/client/webpack.config.js
```diff
@@ -43,6 +43,9 @@ const webpackConfig = {
     ]),
     new webpack.optimize.CommonsChunkPlugin(/* chunkName= */'vendor', /* filename= */'vendor.js', function(module){
       return module.resource && module.resource.indexOf('node_modules') !== -1;
+    }),
+    new webpack.DefinePlugin({
```
raybooysen: Funny, we just discussed yesterday about Define. :)

KeithWoods: Think this is good evidence option to enhance the config.json so include the version and to a proper solution now as discussed on slack. Can discuss tomorrow after scrum perhaps?

fdore: surely the app version comes from the package.json file

KeithWoods: makes sense

DimitarChristoff:
should this be read from `package.json?` currently tagged as:
```js
  "version": "2.1.0",
```

milestone is kind of different...

src/client/webpack.config.js
KeithWoods: I had a look at seeing if you could write a config module on the fly and registered it in the bundle during build. Doesn't seem to be supported out of the box. You'd have to write a plugin I believe. 

For example 

```
// webpack.config.js
var config = {
  someSetting:'foo'
}
// where plugins are registered
new DymanicModulePlugin('my-dynamic-config', config);
```
```
// consumer.js
import config from 'my-dynamic-config';
console.log(config.someSetting);
```

Shame, but until such exists I guess the options are:
1. global variables are the only option. 
2. write a custom config object to disk at build time and configure a dynamic alias to it.

Point 1 (as per this PR) makes the most sense.

PR #276, opened 2016-04-20
Auto dismiss trade notification

### LINKED ISSUE #223: Once a trade is executed, the tile doesn't automatically flip back to prices
In RT, when you executed a trade, the tile would flip and show a confirmation state. After about 5seconds, it would automatically flip back and show you ticking prices again. This was usefull for performing demos, so you didn't have to go back and click all these small `Done` or `Rejected` links to flip it back.

fdore:
Also, remove obsolete code in spot tile model.

Closes #223

### DIFF
--- a/src/client/src/ui/spotTile/model/spotTileModel.js
+++ b/src/client/src/ui/spotTile/model/spotTileModel.js
@@ -132,7 +132,7 @@ export default class SpotTileModel extends ModelBase {
                 : TradeExecutionNotification.createForSuccess(response.trade);
               if (!response.hasError && response.trade.status === TradeStatus.Done) {
                 this._toastNotificationTimerDisposable.setDisposable(
-                  this._schedulerService.async(DISMISS_NOTIFICATION_AFTER_X_IN_MS, () => this.router.publishEvent(this.modelId, 'tradeNotificationDismissed', {}))
+                  this._schedulerService.async.scheduleFuture('', DISMISS_NOTIFICATION_AFTER_X_IN_MS, () => this.router.publishEvent(this.modelId, 'tradeNotificationDismissed', {}))
                 );
               }
             },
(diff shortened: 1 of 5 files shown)


src/client/src/ui/spotTile/model/spotTileModel.js
```diff
@@ -113,22 +114,29 @@ export default class SpotTileModel extends ModelBase {
...
-            this.notification = response.hasError
-              ? TradeExecutionNotification.createForError(response.error)
-              : TradeExecutionNotification.createForSuccess(response.trade);
-          },
-          err => {
-            this.isTradeExecutionInFlight = false;
-            this._log.error(`Error executing ${request.toString()}. ${err}`, err);
-            this.notification = TradeExecutionNotification.createForError(`Unknown stream error`);
+        this._executionService.executeTrade(request)
+          .do(response => {
```
KeithWoods: not a massive fan of side effects in a `do` method, additionally it's setting local state which is a bit dodgy.  

I think we should move this into the bode of the observer (i.e. the function passed to the subscribe function).

src/client/src/ui/spotTile/model/spotTileModel.js
```diff
@@ -113,22 +114,29 @@ export default class SpotTileModel extends ModelBase {
...
+            if (!response.hasError && response.trade.status === TradeStatus.Done) {
+              this.notificationTimer = setTimeout(() => this.router.publishEvent(this.modelId, 'tradeNotificationDismissed', {}), DISMISS_NOTIFICATION_AFTER_X_IN_MS);
```
KeithWoods: Can we use the rx schedule server rather than setTimeout?

It'll wrap setTimeout in a nicer api that we can easily dispose. for example something like this:

```
this.addDisposable(_scheduleService.async(1000, () => { ... })
```

I can't remember the actual method names, but that's close

PR #284, opened 2016-04-21
fix issue with current implementation of flashing new trades

fdore:
Don't flash trades when receiving state of the world

#228

### DIFF
--- a/src/client/src/services/executionService.js
+++ b/src/client/src/services/executionService.js
@@ -39,7 +39,7 @@ export default class ExecutionService extends ServiceBase {
                   _this._serviceClient
                     .createRequestResponseOperation('executeTrade', executeTradeRequest)
                     .map(dto => {
-                      var trade = _this._tradeMapper.mapFromDto(dto.Trade);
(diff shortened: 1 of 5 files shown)


src/client/src/services/executionService.js
```diff
@@ -39,7 +39,7 @@ export default class ExecutionService extends ServiceBase {
                   _this._serviceClient
                     .createRequestResponseOperation('executeTrade', executeTradeRequest)
                     .map(dto => {
-                      var trade = _this._tradeMapper.mapFromDto(dto.Trade);
+                      var trade = _this._tradeMapper.mapFromTradeDto(dto.Trade);
```
DimitarChristoff: `s/var/const/g` for consistency.

raybooysen: What does that mean?

DimitarChristoff: find `var` and replace with `const` globally in vi :) or `let` though we should be using `const` for all immutables

PR #287, opened 2016-04-21
add font family to body, so torn out items share the same font.

### LINKED ISSUE #285: Torn out Item has incorrect font
Torn out:
image

In-browser:
image

fdore:
Remove page container as it is not needed anymore.

Closes #285


DimitarChristoff:
LGTM.

src/client/src/ui/common/components/pageContainer/pageContainer.jsx
```diff
@@ -1,13 +0,0 @@
-import React from 'react';
-import '../../styles/reactive-trader.scss';
-
-export default class PageContainer extends React.Component {
```
KeithWoods: @fdore I though we'd need something like this to get styles back in our popups?

Issue #270, opened 2016-04-20
Positions tab on analytics should show currency aggregation


ferguskeenan:
![positions currency aggregation](https://cloud.githubusercontent.com/assets/8998351/14669259/fc179768-06df-11e6-8770-18c87284ee95.PNG)

PR #290, opened 2016-04-21
Issue270 positions tab

zeala:
#270


src/client/src/ui/analytics/views/analyticsBarChart.jsx
```diff
@@ -9,19 +10,29 @@ export default class PNLChart extends React.Component{
...
       resultObj.max = element[propName] > resultObj.max ? element[propName] : resultObj.max;
       resultObj.min = element[propName] < resultObj.min ? element[propName] : resultObj.min;
 
       return resultObj;
     }, {min: 0, max: 0});
 
     let ratio = (containerWidth/2) / Math.max(Math.abs(maxMinValues.max), Math.abs( maxMinValues.min));
-    let bars = this.props.series.map((element, index) => {
+    let bars = chartData.map((element, index) => {
+      if (element[propName] === 0 && !this.props.isPnL) return;
```
raybooysen: return what?  What does react do if you return immediately out here?

src/client/src/ui/analytics/views/analyticsBarChart.jsx
```diff
@@ -35,10 +46,16 @@ export default class PNLChart extends React.Component{
...
 
-  render(){
-    let bars = this._createBars();
-    return (
-      <div>{bars}</div>
-    );
+  _getPositionsDataFromSeries(propName){
+    let positionsPerCcyObj = this.props.series.reduce((resultObj, element) => {
+      let ccy = element.symbol.substr(0,3);
+      resultObj[ccy] = resultObj[ccy] ? resultObj[ccy] + element[propName] : element[propName];
```
raybooysen: Can we pop a comment explaining what this does?  It's non-obvious.

src/client/src/ui/analytics/views/pnlBar.jsx
```diff
@@ -22,12 +22,13 @@ export default class PNLBar extends React.Component{
     let indicatorClass = 'analytics__barchart-indicator ';
     indicatorClass += isPositive ? 'analytics__barchart-indicator--positive' : 'analytics__barchart-indicator--negative';
     let amountStr = numeral(baseValue).format();
+    let ccyPrefix = this.props.isPnL ? this.props.model.symbol.substr(0, 3) : '';
```
raybooysen: We do a lot of substring calls on symbol. Can we not model this correctly in the model?  So a proper CurrencyPair object with base and terms properties?

src/client/src/ui/analytics/views/analyticsBarChart.jsx
zeala: if it's not a PnL chart we don't want to draw a bar for that option, so skip it, (we do need to draw a bar, though, it if is* PnL)

DimitarChristoff: so do we wan to `.filter()` instead of map? this just creates a sparse array with an undefined which react does not render but has to iterate through.

src/client/src/ui/analytics/views/analyticsBarChart.jsx
```diff
@@ -9,19 +10,29 @@ export default class PNLChart extends React.Component{
...
 
+  _createBars(){
     const containerWidth = 330;
+
     let propName = this.props.isPnL ? CurrencyPairPosition.basePnlName : CurrencyPairPosition.baseTradedAmountName;
-    let maxMinValues = this.props.series.reduce( (resultObj, element) => {
+    let chartData = this.props.isPnL ? this.props.series : this._getPositionsDataFromSeries(propName);
+
+    let maxMinValues = chartData.reduce( (resultObj, element) => {
       resultObj.max = element[propName] > resultObj.max ? element[propName] : resultObj.max;
```
DimitarChristoff: why not `resultObj.min = Math.max(element[propName], resultObj.max)` and `... Math.min(element[propName], resultObj.min);`

you could even do:

```js
let vals = _.map(this.props.series, propName),
        max = _.max(vals),
        min = _.min(vals);
```    - this will be x2 faster than reduce with an arrow function.

File: src/client/src/ui/analytics/views/analyticsBarChart.jsx [hunk 5]
```diff
@@ -9,19 +10,29 @@ export default class PNLChart extends React.Component{
...
+  }
 
+  _createBars(){
     const containerWidth = 330;
+
     let propName = this.props.isPnL ? CurrencyPairPosition.basePnlName : CurrencyPairPosition.baseTradedAmountName;
-    let maxMinValues = this.props.series.reduce( (resultObj, element) => {
+    let chartData = this.props.isPnL ? this.props.series : this._getPositionsDataFromSeries(propName);
+
+    let maxMinValues = chartData.reduce( (resultObj, element) => {
```
DimitarChristoff: why not `resultObj.min = Math.max(element[propName], resultObj.max)` and `... Math.min(element[propName], resultObj.min);`

you could even do:

```js
let vals = _.map(this.props.series, propName),
        max = _.max(vals),
        min = _.min(vals);
```
- this will be x2 faster than reduce with an arrow function.

src/client/src/services/model/currencyPairPosition.js
```diff
@@ -1,10 +1,12 @@
 export default class CurrencyPairPosition {
   _symbol:string;
+  _ccy:string;
   _basePnl:number;
   _baseTradedAmount:number;
 
   constructor(symbol:string, basePnl:number, baseTradedAmount:number) {
     this._symbol = symbol;
+    this._ccy = symbol.length === 6 ? symbol.substr(0, 3) : '';
```
raybooysen: What is currency?  The base or terms?  I think the naming is slightly off.
Is there ever a case where symbol is not a fully fledged currency pair and is something else?  If it isn't, then this sounds like an error.

raybooysen: Also, we aren't running out of characters on our hard drives. :)  can we call this by it's full name?  so either `.currency` or `.baseCurrency`

KeithWoods: Another note on this. The the service layer has a `CurrencyPair` class and this I believe has base and terms properties already, `CurrencyPairPosition` should use the `CurrencyPair` from the service layer not a string symbol

src/client/src/services/model/currencyPairPosition.js
```diff
@@ -21,6 +25,10 @@ export default class CurrencyPairPosition {
     return this._symbol;
   }
 
+  get currencyPair():string {
```
KeithWoods: this will now `: CurrencyPair`

src/client/src/services/model/currencyPairPosition.js
```diff
@@ -1,10 +1,14 @@
+import CurrencyPair from './currencyPair';
+
 export default class CurrencyPairPosition {
   _symbol:string;
+  _currencyPair:string;
```
KeithWoods: this will now `: CurrencyPair`

src/client/src/services/model/currencyPairPosition.js
```diff
@@ -1,10 +1,14 @@
...
   _basePnl:number;
   _baseTradedAmount:number;
 
   constructor(symbol:string, basePnl:number, baseTradedAmount:number) {
     this._symbol = symbol;
+    this._currencyPair = new CurrencyPair(symbol);
```
KeithWoods: Can we just pass the currency pair via the ctor. The reference data service should have loaded it so we should pass that instance

PR #293, opened 2016-04-22
Improve price generation

### LINKED ISSUE #229: Pricing improvements
Some feedback from @ferguskeenan

* Random price walks seem inconsistent - sometimes appear to just tick down regardless

Chat to Fergus for more info

### LINKED ISSUE #230: Spreads not updating
Some feedback from @ferguskeenan

* spreads don't always look right (ticks cause spread to change but spread doesn't update)

JPWatson:
closes #229
- Bug: Random.Next(-5, 5) is not symmetric (min value is inclusive, max value is exclusive)
- Use mean reversion to stop prices diverging from initial value
- Use normal distribution to approximate prices moves

closes #230 
- Use % spread so that spreads can move (albeit not very often)


JPWatson:
Example price path

image

Issue #83, opened 2015-12-16
Frontend support for WSS


odeheurles:
When running under https, the client should connect to /wws, not /ws

http 80
ws 8080

https 443
wws 443

KeithWoods:
@odeheurles  I don't believe this is done. Adding to our next milestone

tdeheurles:
Nginx was prepared in dev. I will have to add some code to the master branch.

KeithWoods:
@tdeheurles do you have an idea of the status of this?

tdeheurles:
I have no idea for the client.
nginx is setup for wss. But it's not tested for now.

Issue #82, opened 2015-12-15
Implement WebSocket fallback


odeheurles:
- [x]  implement fallback in the UI
- [x] update connection status when using long polling in the UI
- [ ] @tdeheurles : configure nginx routing

ferguskeenan:
Related to #221

fdore:
Currently getting the following error when disabling websocket on the client:
XMLHttpRequest cannot load http://web-dev.adaptivecluster.com:8080/lp/open. Response to preflight request doesn't pass access control check: No 'Access-Control-Allow-Origin' header is present on the requested resource. Origin 'http://web-dev.adaptivecluster.com' is therefore not allowed access. The response had HTTP status code 501.

tdeheurles:
fixed by nsgate 0.2.27

PR #294, opened 2016-04-22
Configure long polling fallback

fdore:
display connection type in footer

#82 #83

### DIFF
--- a/src/client/src/services/model/connectionType.js
+++ b/src/client/src/services/model/connectionType.js
@@ -1,0 +1,5 @@
+export default {
+  Unknown: '',
+  WebSocket: 'web socket',
+  LongPolling: 'long polling'
+};
--- a/src/client/src/system/service/autobahnConnectionProxy.js
+++ b/src/client/src/system/service/autobahnConnectionProxy.js
@@ -9,11 +9,21 @@ export default class AutobahnConnectionProxy {
   connection:autobahn.Connection;
 
   constructor(url:string, realm:string) {
+    const useSecure = location.protocol === 'https:';
     this.connection = new autobahn.Connection({
-      url: url,
       realm: realm,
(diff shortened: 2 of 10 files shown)


src/client/src/services/model/connectionType.js
```diff
@@ -0,0 +1,5 @@
+export default {
```
KeithWoods: @raybooysen do you have comments on this 'enum' styling as opposed to an object with static props?

Am I right in thinking this encourages the use of magic strings and should be avoided?

src/client/src/system/service/autobahnConnectionProxy.js
```diff
@@ -9,11 +9,23 @@ export default class AutobahnConnectionProxy {
...
+    const useSecure = location.protocol === 'https:';
+    const securePort = 443;
+    const defaultPort = 8080;
     this.connection = new autobahn.Connection({
-      url: url,
       realm: realm,
       use_es6_promises: true,
-      max_retries: -1 // unlimited retries
+      max_retries: -1, // unlimited retries,
+      transports: [
```
KeithWoods: Bit surprised to only see 2 options here. Were there more than 2 options in the default list?

KeithWoods:
Nice job. Minor question for ray but looks really good.

src/client/src/system/service/autobahnConnectionProxy.js
fdore: According to the Autobahn documentation, 

> supported transports are WebSocket and HTTP long-poll

cf. http://autobahn.ws/js/reference.html#connection-options

KeithWoods: ok fair enough.

PR #298, opened 2016-04-22
fix typo when checking if trade rejected

fdore:


src/client/src/ui/spotTile/views/tradeNotification.jsx
```diff
@@ -30,7 +30,7 @@ export default class TradeNotification extends React.Component {
         'trade-notification',
         className,
         {
-          'trade-notification--rejected': tradeExecutionNotification.status === TradeStatus.Rejected
+          'trade-notification--rejected': tradeExecutionNotification.status === TradeStatus.Rejected.name
```
KeithWoods: This is the downside of using static props. We should have a discussion on this pattern.

KeithWoods: Actually shouldn't `tradeExecutionNotification.status` be a `TradeStatus`, if so you'd not need the `.name`

fdore: Fair point

Issue #201, opened 2016-03-08
Blotter does not scale when in its own window


tdeheurles:
![image](https://cloud.githubusercontent.com/assets/9496645/13600238/dc38b154-e527-11e5-949d-fce88c238dd3.png)

Not sure it is relevant, just to notice.

LeeCampbell:
Might be related to #216

KeithWoods:
This should be fixed via the new styling work, adding the appropriate label.

fdore:
Blotter is now taking the full space when opening in a new window. The blotter still wont scale when resizing the window though as the resize handler is still attached to the parent window.

KeithWoods:
@DimitarChristoff were you looking at this?

DimitarChristoff:
yes, it's now fixed in react-dimensions and blocked by https://github.com/digidem/react-dimensions/pull/22 being merged in. we can temporarily stick our own dimensions.jsx until the PR goes through

PR #302, opened 2016-04-25
local dimensions fixed for #201 until PR goes in for react-dimensions…

DimitarChristoff:
… makes events fire correctly in tearoffs


KeithWoods:
Can we please move the `Dimensions` decorators to a directory named `src/ui/common/components/dimensions`

src/client/src/system/dimensions.jsx
```diff
@@ -0,0 +1,140 @@
+/*eslint-disable */
```
KeithWoods: I think we need to add the licencing information as per the initial projects licencing agreement. Additionally can we write a note as to why we've forked it locally.

src/client/src/ui/blotter/views/blotterView.jsx
```diff
@@ -1,7 +1,7 @@
 import React from 'react';
 import { Table, Column, Cell } from 'fixed-data-table';
 import { DateCell, NotionalCell } from './';
-import Dimensions from 'react-dimensions';
+import Dimensions from './../../../system/dimensions';
```
KeithWoods: This should look like this once moved correctly:
import { Dimensions } from '...';

src/client/src/system/dimensions.jsx
DimitarChristoff: agreed, will amend.

On Monday, 25 April 2016, Keith Woods <notifications@github.com> wrote:

> In src/client/src/system/dimensions.jsx
> <https://github.com/AdaptiveConsulting/ReactiveTraderCloud/pull/302#discussion_r60889533>
> :
>
> > @@ -0,0 +1,140 @@
> > +/*eslint-disable */
>
> I think we need to add the licencing information as per the initial
> projects licencing agreement. Additionally can we write a note as to why
> we've forked it locally.
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/AdaptiveConsulting/ReactiveTraderCloud/pull/302/files/06adbf723df5998eaf38c3d69099f6f092d910ba#r60889533>
>


-- 
Dimitar Christoff

"JavaScript is to JAVA what hamster is to ham"
@D_mitar - https://github.com/DimitarChristoff

src/client/src/ui/blotter/views/blotterView.jsx
DimitarChristoff: why, we don't export an object with a `Dimensions` property but return the function factory for the decorator (HOC)

PR #303, opened 2016-04-25
Remove react-popout

### LINKED ISSUE #301: Remove react-popout
There are numerous issues with this component:

* approach is incorrect, you don't need a component to open an element. 
* the popout child is bound to a component rendered in a different visual tree
* there are bugs with the onload function (we've forked it)
* seems to be issues with set state (need to verify)

KeithWoods:
Fixes #301.


src/client/src/ui/common/popout/reactPopoutService.js
```diff
@@ -0,0 +1,44 @@
...
+export default class ReactPopoutService {
+  openPopout(options:PopoutOptions, view:React.Component) {
+    let popoutContainer;
+    let windowOptionsString = this._getWindowOptionsString(options.windowOptions);
+    let childWindow = window.open(options.url, options.title, windowOptionsString);
+    let onloadHandler = () => {
+      childWindow.document.title = options.title;
+      popoutContainer = childWindow.document.createElement('div');
+      popoutContainer.id = POPOUT_CONTAINER_ID;
+      childWindow.document.body.appendChild(popoutContainer);
```
DimitarChristoff: maybe we should think about passing a selector for rootNode and document.body as backup. this allows you to have a proper dom present and an element where to install vs `document.body` only.

also, it's possible onload has already fired by the time the fn runs (likely with an empty doc), should check readystate or assign onload handlers.

src/client/src/ui/regions/model/popoutRegionModel.js
```diff
@@ -0,0 +1,44 @@
...
+    let width = modelRegistration.regionSettings && modelRegistration.regionSettings.width
+      ? modelRegistration.regionSettings.width
+      : 400;
+    let height = modelRegistration.regionSettings && modelRegistration.regionSettings.height
+      ? modelRegistration.regionSettings.height
+      : 400;
+    let popoutOptions = new PopoutOptions(
+      modelRegistration.key,
+      '/#/popout',
+      '',
```
DimitarChristoff: shouldn't this be the component title?

src/client/src/ui/common/popout/popoutOptions.js
```diff
@@ -0,0 +1,25 @@
...
+    height: number,
+    resizable: string,
+    scrollable: string
+  };
+  constructor(
+    key:string,
+    url:string,
+    title:string,
+    onClosing,
+    windowOptions: {width: number, height: number, resizable: string, scrollable: string}
```
DimitarChristoff: resizable is not an option that is valid/respected any more.

src/client/src/ui/common/popout/reactPopoutService.js
KeithWoods: Ok good Idea, lets chat in person to to sort it. Might be quicker to code it up together

src/client/src/ui/regions/model/popoutRegionModel.js
KeithWoods: yes, nice catch

src/client/src/ui/common/popout/popoutOptions.js
KeithWoods: not sure what you mean here, lets chat

DimitarChristoff: `resizeable=no` is not respected/used in `window.open` anymore (deprecated), but it was there previously because OpenFin does.

src/client/src/ui/common/popout/reactPopoutService.js
```diff
@@ -0,0 +1,54 @@
...
+    }
+    if (!windowOptions.width) {
+      windowOptions.width = 400;
+    }
+    windowOptions.top = ((window.innerHeight - windowOptions.height) / 2) + window.screenY;
+    windowOptions.left = ((window.innerWidth - windowOptions.width) / 2) + window.screenX;
+
+    let optionsArray = [];
+    for (let key in windowOptions) {
+      windowOptions.hasOwnProperty(key) && optionsArray.push(key + '=' + windowOptions[key]);
```
DimitarChristoff: most of these window features options rely on `yes` or `no` values - so it needs to be something like `+ Boolean(windowOpetions[key]) ? 'yes' : 'no'` - see https://developer.mozilla.org/en-US/docs/Web/API/Window/open

src/client/src/ui/regions/model/popoutRegionModel.js
```diff
@@ -0,0 +1,45 @@
...
+    var title = options.regionSettings && options.regionSettings.title ? options.regionSettings.title : '';
+    let popoutOptions = new PopoutOptions(
+      modelRegistration.key,
+      '/#/popout',
+      title,
+      () => this._popoutClosed(modelRegistration.model),
+      {
+        width: width,
+        height: height,
+        resizable: 'no',
```
DimitarChristoff: this is better as boolean so openfin can also work / consume the options when abstraction is added

src/client/src/ui/common/popout/popoutOptions.js
KeithWoods: ok sure thanks for the note

src/client/src/ui/common/popout/reactPopoutService.js
KeithWoods: ok thanks will update

src/client/src/ui/common/popout/reactPopoutService.js
```diff
@@ -0,0 +1,54 @@
...
+    let popoutContainer;
+    let windowOptionsString = this._getWindowOptionsString(options.windowOptions);
+    _log.debug(`Opening child window url:${options.url},title:${options.title}`);
+    let childWindow = window.open(options.url, options.title, windowOptionsString);
+    let onloadHandler = () => {
+      _log.debug(`Popout window loading`);
+      childWindow.document.title = options.title;
+      popoutContainer = childWindow.document.createElement('div');
+      popoutContainer.id = POPOUT_CONTAINER_ID;
+      childWindow.document.body.appendChild(popoutContainer);
```
DimitarChristoff: our body route has `div#root` in it, should we not append to that like we do in the main app? else, DOM ends up as `<div id=root></div><div id=${POPOUT_CONTAINER_ID}>...</div>`

src/client/src/ui/regions/model/popoutRegionModel.js
KeithWoods: As it stands this object is meant to mirror the options passed to `window.open`, we could make it more general as you suggest, probably makes more sense since open fin or any container for that matter will be different.

src/client/src/ui/common/popout/reactPopoutService.js
KeithWoods: Possibly but it seems odd that this service know of that. Ideally it'd have it's own popout page that loads a subset of the scripts but we don't have that. I think that'd be the ideal solution.

I suggest we leave it for now.

src/client/src/ui/common/popout/reactPopoutService.js
DimitarChristoff: it will scale better. the OpenFin window open api accepts far more options.

eg payload for a openfin window:

```js
    {
        url: '/#popout',
        title: 'Something',
        name: 'unique id',
        defaultWidth: 332,
        defaultHeight: 191,
        resizable: false,
        scrollable: false,
        maximizable: false,
        frame: false,
        autoShow: true,
        cornerRounding: {
           width: 5,
           height: 5
        }  
      }
```

src/client/src/ui/spotTile/model/spotTileModel.js
```diff
@@ -96,7 +96,7 @@ export default class SpotTileModel extends ModelBase {
   @observeEvent('popOutTile')
   _onPopOutTile() {
     this._log.info(`Popping out tile`);
-    this._regionManagerHelper.popout(332, 190);
+    this._regionManagerHelper.popout(`${this.currencyPair.symbol} Spot`, 332, 155);
```
DimitarChristoff: these hard values will not scale well for responsive design - the view should pass these.

DimitarChristoff:
LGTM - one thing that react-popout was solving was ability to `setState({tearoff: true})` which meant components could have a default tearoff state when they get installed at view / persisted. 

also, need to remove react-popout from `package.json`.

src/client/src/ui/spotTile/model/spotTileModel.js
KeithWoods: There is a to do on that method to change this


Sent from Outlook Mobile






On Tue, Apr 26, 2016 at 6:34 AM -0700, "Dimitar Christoff" <notifications@github.com> wrote:





> @@ -96,7 +96,7 @@ export default class SpotTileModel extends ModelBase {
>    @observeEvent('popOutTile')
>    _onPopOutTile() {
>      this._log.info(`Popping out tile`);
> -    this._regionManagerHelper.popout(332, 190);
> +    this._regionManagerHelper.popout(`${this.currencyPair.symbol} Spot`, 332, 155);

these hard values will not scale well for responsive design - the view should pass these.

---
You are receiving this because you authored the thread.
Reply to this email directly or view it on GitHub:
https://github.com/AdaptiveConsulting/ReactiveTraderCloud/pull/303/files/ebceeaf1d5e0ff6ffdb3c5c8fe31d926d5e31f4e#r61085523

src/client/src/ui/regions/model/popoutRegionModel.js
KeithWoods: I think we can make this change as we implement the open fin stuff

KeithWoods:
So in summary:
* Remove `react-popout` from packages.json - will push an update
* Update the windows options api as we implement openfin
* Pop a story in the backlog to let the views handle the widths and heights not the models

KeithWoods:
@DimitarChristoff react-popout removed, think this can now be merged.

PR #306, opened 2016-04-25
Merge latest master

JPWatson:
I hadn't spotted the dev branch


KeithWoods:
we should try do this as a rebase not a merge given we've been following the rebase workflow

JPWatson:
I could revert the PR on master and I'll do a new PR to dev?

KeithWoods:
Master is locked so we can't do a force push to it. You could back out the changes and check the deletion in as a separate commit, then make a new PR against dev. This is prob the better approach. I might just have a look if we can lock down master so we can only push via PRs

PR #315, opened 2016-04-29
Issues/235 charti q

zeala:


src/client/src/ui/spotTile/model/spotTileModel.js
```diff
@@ -99,6 +102,14 @@ export default class SpotTileModel extends ModelBase {
     this._regionManagerHelper.popout(`${this.currencyPair.symbol} Spot`, 332, 155);
   }
 
+  @observeEvent('openChartIQ')
+  _onOpenChartIQ(){
+    this._log.info(`Open ChartIQ`);
+
+    this._executionService._openFin.getChartIQInstance(null, null, this._startChartIQTimer.bind(this));
+    this._executionService._openFin.viewChartIQ(this.currencyPair.symbol);
```
KeithWoods: this looks dodgy, referencing a 'private' field of a service

src/client/src/ui/spotTile/model/spotTileModel.js
```diff
@@ -213,4 +224,13 @@ export default class SpotTileModel extends ModelBase {
         }
     }
   }
+
+  _startChartIQTimer(){
+    this.chartIQisOpening = true;
+    setTimeout(this._stopChartIQTimer.bind(this), 2000);
```
KeithWoods: should probably use the schedule service for this as it wraps it up in a disposable we can manage locally

src/client/src/ui/spotTile/views/spotTileView.jsx
```diff
@@ -28,6 +28,10 @@ export default class SpotTileView extends ViewBase {
     let spotDateClass = classnames('spot-tile__delivery', {'hide': model.hasNotification});
     let notification = this._tryCreateNotification();
     let priceComponents = this._createPriceComponents();
+    let showChartIQIcon = model._executionService._openFin.available;
+    let chartIQIconClassName =  !showChartIQIcon ? 'spot-tile__icon--hidden' :
```
KeithWoods: can we use `classnames` for this rather than the ternary operators?

src/client/src/services/executionService.js
```diff
@@ -57,4 +57,9 @@ export default class ExecutionService extends ServiceBase {
       }
     );
   }
+
+  openChartIQ(currencyPair, chartIQTimerHandler){
```
KeithWoods: why does the execution service car about these charts?

src/client/src/system/openFin/openFin.js
```diff
@@ -57,4 +58,50 @@ export default class OpenFin {
         return disposables;
       });
   }
+
+  viewChartIQ(symbol): void {
+    this.getChartIQInstance(symbol, this.refreshChartIQ, this.startChartIQ);
```
KeithWoods: should this be `this._getChartIQInstance(...`?

src/client/src/system/openFin/openFin.js
```diff
@@ -57,4 +58,50 @@ export default class OpenFin {
...
+    let interval = 5;
+    let chartIqAppId = 'ChartIQ';
+    console.log(`Publishing message to chartIQ with symbol ${symbol} and interval ${interval}`);
+    fin.desktop.InterApplicationBus.publish('chartiq:main:change_symbol', { symbol: symbol, interval: interval });
+  }
+
+  startChartIQ(symbol){
+    let interval = 5;
+    let chartIqAppId = 'ChartIQ';
+    console.log('Chart IQ was not running, so starting new app with symbol = ' + symbol + ' and interval ' + interval);
```
KeithWoods: should be using the logger

src/client/src/ui/spotTile/model/spotTileModel.js
```diff
@@ -213,4 +223,9 @@ export default class SpotTileModel extends ModelBase {
         }
     }
   }
+
+  _startChartIQTimer(){
+    this.chartIQisOpening = true;
+    this._schedulerService.async.scheduleFuture('', 2000, ()=> this.chartIQisOpening = false);
```
KeithWoods: this needs to happen on the esp dispatch loop else the view won't know there was a change. something like this:
```
this._schedulerService.async.scheduleFuture('', 2000, ()=> { 
  this.router.runAction(theModelId, () => {
    this.chartIQisOpening = false; 
  }
});
```
or preferably (and handling the disposable):

```
// note syntax might be off, just coded it from memory
this.addDisposable( 
  Observable.timer(2000, this._schedulerService.async)
    .subscribeWithRouter(
      this.router, 
      this.modelId, 
      ()=> {
        this.chartIQisOpening = false
      }
);
```
[more info](https://github.com/Reactive-Extensions/RxJS/blob/master/doc/api/core/operators/timer.md)

src/client/src/system/openFin/openFin.js
zeala: There is a lot of cleaning up that was done on a different branch, including deleting the unused code, string templates and logging. I'll remove the newly added console.log statements from this branch and will add it to the logger in the notifications branch

src/client/src/system/openFin/openFin.js
```diff
@@ -84,8 +85,52 @@ export default class OpenFin {
...
+  
   get _currentWindow() {
     return fin.desktop.Window.getCurrent();
   }
+  
+  viewChartIQ(symbol): void {
+    this.getChartIQInstance(symbol, this._refreshChartIQ, this._startChartIQ);
+  }
+
+  getChartIQInstance(symbol, callback, errorHandler):fin.desktop.Application{
```
KeithWoods: we should wrap this up in RX:

```
getChartIQInstance(symbol) { // note removal of callbacks callback and errorHandler
 return Rx.Observable.create(observer => {
  let chartIqApp = _.find(apps, ((app) => {
    return app.isRunning && app.uuid === chartIqAppId;
  }));
  if(chartIqApp) {
    observer.onNext(symbol);
    observer.onCompleted();
  } else {
    o.onError(symbol);
  } 
 });
}

```

Then clean up the calling code so we don't have to use 'bind'. Additionally, we can signal via this stream when it completes so the calling code doesn't have to poll.

Happy to talk you through it

src/client/src/system/openFin/openFin.js
```diff
@@ -84,8 +85,51 @@ export default class OpenFin {
...
+  
+  viewChartIQ(symbol){
+    return Rx.Observable.create(observer => {
+      let chartIqAppId = 'ChartIQ';
+      fin.desktop.System.getAllApplications(function(apps) {
+        let chartIqApp = _.find(apps, ((app) => {
+          return app.isRunning && app.uuid === chartIqAppId;
+        }));
+        if(chartIqApp) {
+          observer.onNext(symbol);
```
KeithWoods: it's a bit odd we call back with the symbol we are given, what's the idea here?

src/client/src/ui/spotTile/model/spotTileModel.js
```diff
@@ -99,6 +106,27 @@ export default class SpotTileModel extends ModelBase {
...
+        .subscribeWithRouter(
+          this.router,
+          this.modelId,
+          (symbol:string) => {
+            this._openFin.refreshChartIQ(symbol);
+          },
+          symbol => {
+            this._log.info('chartIQ not running, starting a new instance');
+            this._openFin.startChartIQ(symbol);
+            this._startChartIQTimer();
```
KeithWoods: Why do we need this timer?

src/client/src/ui/spotTile/model/spotTileModel.js
```diff
@@ -213,4 +227,16 @@ export default class SpotTileModel extends ModelBase {
         }
     }
   }
+
+  _startChartIQTimer(){
```
KeithWoods: given this is a hack, lets just move it to the view. And we then don't need RX, the whole rx approach seems suitable when there was data being pushed from open fin, but since that isn't the case and it's just a visual hack, can we just do it right in the view?

src/client/src/ui/spotTile/model/spotTileModel.js
```diff
@@ -99,6 +106,13 @@ export default class SpotTileModel extends ModelBase {
     this._regionManagerHelper.popout(`${this.currencyPair.symbol} Spot`, 332, 155);
   }
 
+  @observeEvent('openChartIQ')
+  _onOpenChartIQ(){
```
KeithWoods: can we rename all usages of 'chart iq' with the actually functionality? Like `displayCurrencyChart`?

src/client/src/ui/spotTile/model/spotTileModel.js
```diff
@@ -99,6 +104,12 @@ export default class SpotTileModel extends ModelBase {
     this._regionManagerHelper.popout(`${this.currencyPair.symbol} Spot`, 332, 155);
   }
 
+  @observeEvent('displayCurrencyChart')
+  _onDisplayCurrencyChart(){
+    this._log.info(`Display currency chart - ChartIQ`);
```
KeithWoods: does this line pass the linter? Thought it blows up when you use a template string without any template variables

src/client/src/ui/spotTile/views/spotTileView.jsx
```diff
@@ -28,6 +37,16 @@ export default class SpotTileView extends ViewBase {
     let spotDateClass = classnames('spot-tile__delivery', {'hide': model.hasNotification});
     let notification = this._tryCreateNotification();
     let priceComponents = this._createPriceComponents();
+    let showChartIQIcon = model._executionService._openFin.available;
```
KeithWoods: this is pretty hacky, the model should do the proxying. And even they it shouldn't access 'private' variables of other objects

src/client/src/ui/spotTile/views/spotTileView.jsx
```diff
@@ -66,6 +87,15 @@ export default class SpotTileView extends ViewBase {
     );
   }
 
+  _displayCurrencyChart(){
+    router.publishEvent(this.props.modelId, 'displayCurrencyChart', {});
+    //set the currencyChartIsOpening to display the icon spinner
+
+    let timeoutId = setTimeout(()=> this.setState({currencyChartIsOpening: false}), 2000);
+    this.setState({currencyChartTimeoutId: timeoutId});
```
KeithWoods: why are we calling set state twice?

src/client/src/ui/spotTile/views/spotTileView.jsx
```diff
@@ -66,6 +87,15 @@ export default class SpotTileView extends ViewBase {
     );
   }
 
+  _displayCurrencyChart(){
+    router.publishEvent(this.props.modelId, 'displayCurrencyChart', {});
+    //set the currencyChartIsOpening to display the icon spinner
```
KeithWoods: comment in an odd spot

src/client/src/ui/spotTile/spotTileFactory.js
```diff
@@ -19,19 +20,22 @@ export default class SpotTileFactory {
...
   _schedulerService:SchedulerService;
+  _openFin:OpenFin;
 
   constructor(
     router:Router,
     pricingService:PricingService,
     executionService:ExecutionService,
     regionManager:RegionManager,
-    schedulerService: SchedulerService
+    schedulerService: SchedulerService,
```
KeithWoods: do we still need the schedule service? think it can be removed from here

src/client/src/bootstrapper.js
```diff
@@ -103,7 +103,7 @@ class Bootstrapper {
     let spotTileLoader = new SpotTileLoader(
       espRouter,
       this._referenceDataService,
-      new SpotTileFactory(espRouter, this._pricingService, this._executionService, regionManager, this._schedulerService)
+      new SpotTileFactory(espRouter, this._pricingService, this._executionService, regionManager, this._schedulerService, this._openFin)
```
KeithWoods: do we still need the schedule service? think it can be removed from here

KeithWoods: actually looks like we do

src/client/src/ui/spotTile/model/spotTileModel.js
KeithWoods: looks like it's used on the execution path

src/client/src/ui/spotTile/model/spotTileModel.js
```diff
@@ -73,6 +79,7 @@ export default class SpotTileModel extends ModelBase {
     this.executionConnected = false;
     this.isTradeExecutionInFlight = false;
     this._regionManagerHelper = new RegionManagerHelper(RegionNames.workspace, regionManager, this);
+    this.openFinAvailable = openFin.available;
```
KeithWoods: looks like `open.available` is a bug, will always be true. 

Can you remove this property on the openFin object (any anywhere using it) and use `openFin.isRunningInOpenFin`?

src/client/src/ui/spotTile/views/spotTileView.jsx
```diff
@@ -66,6 +87,12 @@ export default class SpotTileView extends ViewBase {
     );
   }
 
+  _displayCurrencyChart(){
+    router.publishEvent(this.props.modelId, 'displayCurrencyChart', {});
+    let timeoutId = setTimeout(()=> this.setState({currencyChartIsOpening: false}), 2000);
```
KeithWoods: just thinking you might want to run the hack code first then publish the event to launch open fin.

KeithWoods:
Cool think it's ready to go, just finally can you squish the commits?

PR #317, opened 2016-05-04
Issues/236 notifications

zeala:


src/client/src/notificationBootstrapper.js
```diff
@@ -0,0 +1,22 @@
...
+import React from 'react';
+import TradeNotificationView from './ui/notification/tradeNotificationView';
+
+export default class NotificationBootstrapper{
+
+  run(){
+    window.onNotificationMessage = (message) => this.handleNotificationMessage(message);
+  }
+
+  dismissNotification(){
```
KeithWoods: should be called `_dismissNotification`

src/client/src/notificationBootstrapper.js
```diff
@@ -0,0 +1,22 @@
...
+    window.fin.desktop.Notification.getCurrent().close();
+  }
+
+  handleNotificationMessage(message){
```
KeithWoods: should be called `_handleNotificationMessage`

src/client/src/notificationBootstrapper.js
```diff
@@ -0,0 +1,22 @@
+import ReactDOM from 'react-dom';
+import React from 'react';
+import TradeNotificationView from './ui/notification/tradeNotificationView';
+
+export default class NotificationBootstrapper{
+
+  run(){
+    window.onNotificationMessage = (message) => this.handleNotificationMessage(message);
```
KeithWoods: is this some openFin magic?

src/client/src/services/blotterService.js
```diff
@@ -2,18 +2,30 @@ import Rx from 'rx';
...
 
-  constructor(serviceType:string, connection:Connection, schedulerService:SchedulerService, referenceDataService:ReferenceDataService) {
+  constructor(serviceType:string,
+              connection:Connection,
+              schedulerService:SchedulerService,
+              referenceDataService:ReferenceDataService,
+              openFin:OpenFin) {
     super(serviceType, connection, schedulerService);
     this._tradeMapper = new TradeMapper(referenceDataService);
+    this.openFin = openFin;
```
KeithWoods: should be `_openFin`

src/client/src/services/blotterService.js
```diff
@@ -2,18 +2,30 @@ import Rx from 'rx';
...
   }
 
+  createNotification(trade:Trade){
+    if ((trade.status.name == 'Done' || trade.status.name == 'Rejected')){
```
KeithWoods: should use triple equals

src/client/src/services/model/trade.js
```diff
@@ -23,7 +24,7 @@ export default class Trade {
     tradeDate:Date,
     valueDate:string,
     status:TradeStatus,
-    isNew:boolean
+    isStateOfTheWorld:boolean
```
KeithWoods: I know this was already here but I think it's modeled incorrectly, the fact that trade is part of the initial batch has nothing to do with trade economics. 

Really this state should exist in a wrapper payload

src/client/src/services/model/trade.js
```diff
@@ -35,7 +36,8 @@ export default class Trade {
     this._tradeDate = tradeDate;
     this._valueDate = valueDate;
     this._status = status;
-    this._isNew = isNew;
+    this._isNew = !isStateOfTheWorld;
```
KeithWoods: shouldn't we delete `_isNew` now?

src/client/src/services/model/tradeNotification.js
```diff
@@ -0,0 +1,35 @@
+import { Trade, TradeStatus, Direction } from './';
+import numeral from 'numeral';
+import moment from 'moment';
+
+export const BOUGHT = 'Bought';
```
KeithWoods: how come we export these strings?

src/client/src/services/model/trade.js
zeala: No, isNew is used to flash the newly added row in the blotter. It's used for a different purpose than the isStateOfTheWorld, but the initial value is set by the value of the 'isStateOfTheWorld'

src/client/src/services/model/tradeNotification.js
```diff
@@ -0,0 +1,35 @@
...
+
+  constructor(trade: Trade){
+    this.spotRate = trade.spotRate;
+    this.notional = numeral(trade.notional).format('0,000,000[.]00');
+    this.direction = trade.direction == Direction.Buy ? BOUGHT : SOLD;
+    this.baseCurrency = trade.currencyPair.base;
+    this.tradeId = trade.tradeId;
+    this.termsCurrency = trade.currencyPair.terms;
+    this.tradeDate = trade.tradeDate;
+    this.tradeStatus = trade.status == TradeStatus.Done ? DONE : REJECTED;
```
KeithWoods: I think we should just take the status object and leave the display text upto the JSX. 

Another approach would be to add decorators to the status object that allow us to centralise the display strings.

src/client/src/system/openFin/openFin.js
```diff
@@ -120,4 +121,32 @@ export default class OpenFin {
...
+    if (!this.isRunningInOpenFin) return;
+
+    let tradeNotification = new TradeNotification(trade);
+
+    let notification = new fin.desktop.Notification({
+      url: '/notification.html',
+      message: tradeNotification,
+      onMessage: () => {
+        let win = fin.desktop.Window.getCurrent();
+        win.getState(state => {
```
KeithWoods: we have similar code in `maximise()` can we pull this out to a common method?

src/client/src/ui/blotter/model/blotterModel.js
```diff
@@ -69,6 +69,9 @@ export default class BlotterModel extends ModelBase {
               else {
                 this.trades.unshift(trade);
               }
+              if (!trade.isStateOfTheWorld){
+                this._blotterService.createNotification(trade);
```
KeithWoods: we should have a separate notification service OR just directly call openFin here.

I initially thought we showed in app notifications, but since we don't perhaps just directly calling openFin here makes sense for now.

src/client/src/ui/notification/tradeNotificationView.jsx
```diff
@@ -0,0 +1,56 @@
+import React from 'react';
+import './notification.scss';
+import classnames from 'classnames';
+
+const DONE = 'Done';
```
KeithWoods: I believe this is defined in `TradeNotification`, we should lift it from both places, or just put it here as suggested before.

src/client/src/ui/notification/notification.scss
```diff
@@ -0,0 +1,50 @@
...
+  padding: 0;
+  font-size: 12px;
+}
+
+.notification__title{
+  margin-bottom: 4px;
+}
+
+.notification__status-done{
+  display: none;
```
fdore: should be notification__status--done

src/client/src/ui/notification/notification.scss
```diff
@@ -0,0 +1,50 @@
...
+}
+
+.notification__status-rejected{
```
fdore: should be notification__status--rejected

src/client/src/services/model/trade.js
zeala: @fdore  - could you please take a look? I don't think that the current implementation will support an initial payload, as all trades, even at the initial subscription, are processed individually?

src/client/src/ui/blotter/model/blotterModel.js
zeala: openFin is injected in services that use it, we don't inject it in the models/views directly. can you suggest a way to access/call openFin directly from the model?

src/client/src/services/blotterService.js
```diff
@@ -9,12 +8,16 @@ var _log:logger.Logger = logger.create('BlotterService');
 
 export default class BlotterService extends ServiceBase {
 
-  constructor(serviceType:string, connection:Connection, schedulerService:SchedulerService, referenceDataService:ReferenceDataService) {
+  constructor(serviceType:string,
+              connection:Connection,
+              schedulerService:SchedulerService,
+              referenceDataService:ReferenceDataService,
+              openFin:OpenFin) {
```
KeithWoods: guessing we don't need `openFin` here anymore

src/client/src/services/blotterService.js
```diff
@@ -9,12 +8,16 @@ var _log:logger.Logger = logger.create('BlotterService');
...
     super(serviceType, connection, schedulerService);
     this._tradeMapper = new TradeMapper(referenceDataService);
   }
 
-  getTradesStream() : Rx.Observable<Array<Trade>> {
+  getTradesStream() : Rx.Observable<TradesUpdate> {
```
KeithWoods: you should import this type, if and when we turn on flow it'll expect it imported

src/client/src/services/mappers/tradeMapper.js
```diff
@@ -10,12 +10,18 @@ export default class TradeMapper {
   }
 
   mapFromDto(dto:Object) : Array<Trade> {
-    return _.map(
+    let tradesUpdate = new TradesUpdate(dto.IsStateOfTheWorld, dto.IsStale);
```
KeithWoods: this should take the trades in it's ctor, typically service domain entities are 'immutable' so should not be altered after creation.

src/client/webpack.config.js
```diff
@@ -34,24 +34,34 @@ const webpackConfig = {
...
     new webpack.optimize.DedupePlugin(),
     new HtmlWebpackPlugin({
       template: './src/index.html',
       hash: true,
       filename: 'index.html',
-      inject: 'body'
+      inject: 'body',
+      excludeChunks: ['notification']
     }),
+    new HtmlWebpackPlugin({
```
KeithWoods: this is really cool

PR #344, opened 2016-05-16
use absolute links in readme.md for github webpage publishing

qooroo:


fdore:
I'm not sure we should use absolute paths as the docs should depend on the branch. We should be able to use paths relative to the docs folder?

PR #350, opened 2016-05-18
Implement Openfin popout

fdore:
- implemented new service to tear out windows when running in openfin. We now have two services (BrowserPopoutService and OpenfinPopoutService) which implement the same interface. The actual implementation is abstracted with a PopoutServiceFactory.

- refactored shell so it is now container agnostics. The shell is wrapped into a Chrome component which works out itself if it is running in openfin or not.

#337


src/client/src/system/openFin/openFin.js
```diff
@@ -22,16 +22,15 @@ export default class OpenFin {
...
-    this._currentWindow.close();
+  close(window = this._currentWindow){
+    window.close(true, () => _log.info('Window closed with success.'), err => _log.error(`Failed to close window. Error: ${err}`));
   }
-  
-  minimise(e){
-    this._currentWindow.minimize();
+
+  minimize(window = this._currentWindow){
+   window.minimize(() => _log.info('Window minimized with success.'), err => _log.error(`Failed to minimize window. Error: ${err}`));
```
KeithWoods: you can dump `err` as the last param and it'll get passed to the debug console, i.e. 
```
log.error(`blah ${err}`, err);
```

src/client/src/ui/common/popout/popoutServiceFactory.js
```diff
@@ -0,0 +1,13 @@
...
+import OpenfinPopoutService from './openfinPopoutService';
+
+
+export default class PopoutServiceFactory {
+  constructor(openfin) {
+    this._openfin = openfin;
+  }
+
+  getService() {
+    return window.fin ? new OpenfinPopoutService(this._openfin) : new BrowserPopoutService();
```
KeithWoods: there is a prop on `OpenFin` called `isRunningInOpenFin`  we should point to that for consistency.

src/client/src/ui/regions/model/popoutRegionModel.js
```diff
@@ -1,15 +1,14 @@
...
 import { createViewForModel } from '../';
 
 export default class PopoutRegionModel extends RegionModel {
-  _reactPopoutService:ReactPopoutService;
 
-  constructor(modelId:string, regionName:string, router:Router) {
+  constructor(modelId:string, regionName:string, router:Router, openfin) {
     super(modelId, regionName, router);
-    this._reactPopoutService = new ReactPopoutService();
+    this._reactPopoutService = new PopoutServiceFactory(openfin).getService();
```
KeithWoods: should prob rename `_reactPopoutService` now

KeithWoods: I'd also be tempted to inline `PopoutServiceFactory` given we just new it up and throw it away. Perhaps a 'private' method in this class would be cleaner

src/client/src/ui/shell/model/shellModel.js
```diff
@@ -2,56 +2,35 @@ import { Router, observeEvent } from 'esp-js/src';
...
 
 export default class ShellModel extends ModelBase {
   _connection:Connection;
-  _openFin:OpenFin;
   sessionExpired:boolean;
   wellKnownModelIds:WellKnownModelIds;
-  isRunningInOpenFin:boolean;
 
-  constructor(modelId:string, router:Router, connection:Connection, openFin) {
+  constructor(modelId:string, router:Router, connection:Connection) {
```
KeithWoods: nice, this class looks loads cleaner

Issue #46, opened 2015-12-01
WPF Credit checker integration


odeheurles:
same story as excel integration, the existing code is there and we should keep the same protocol to not have to rebuild/redeploy the WPF app.

https://github.com/AdaptiveConsulting/ReactiveTraderJS/blob/master/src/ts/OpenFin/OpenFin.ts

KeithWoods:
dependent on #217

fdore:
so it seems that the uuid of RT needs to be reactive-trader-js for both apps to be able to communicate, which is not ideal if we support multiple environments...

fdore:
New version allowing any uuid available here http://adaptiveconsulting.github.io/ReactiveTraderCloud/install/LimitChecker/LimitChecker.application

PR #352, opened 2016-05-18
Credit check

fdore:
#46

### DIFF
--- a/src/client/src/system/openFin/openFin.js
+++ b/src/client/src/system/openFin/openFin.js
@@ -16,7 +16,10 @@ export default class OpenFin {
   constructor() {
     this.tradeClickedSubject = new Rx.Subject();
     this.limitCheckId = 1;
-    this._initializeLimitChecker();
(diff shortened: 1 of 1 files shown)


KeithWoods:
nice work. 

@fdore has this been checked with the excel sheet that open fin chats to?

fdore:
@KeithWoods What do you mean exactly? @zeala is currently looking at the excel integration part.

KeithWoods:
I believe when running in open fin and you book a trade there is functionality to chat to an excel spreadsheet to check if there is credit for the transaction.

fdore:
Do you mean that app http://adaptiveconsulting.github.io/ReactiveTraderJS/install/LimitChecker/LimitChecker.application as opposed to Excel?

KeithWoods:
oh, yeah sorry must be that and not excel

fdore:
Yeah, that's the app I have tested against.

Issue #45, opened 2015-12-01
Excel integration


odeheurles:
See existing code: https://github.com/AdaptiveConsulting/ReactiveTraderJS/blob/master/src/ts/OpenFin/OpenFin.ts

We need to use the exact same protocol on the OpenFin bus: we don't want to have to change the Excel spreadsheet / code.

KeithWoods:
dependent on #217

PR #354, opened 2016-05-18
issue #45 - excel integration

zeala:


src/client/src/ui/blotter/model/blotterModel.js
```diff
@@ -41,6 +41,8 @@ export default class BlotterModel extends ModelBase {
     _log.info(`Blotter starting`);
     this._subscribeToConnectionStatus();
     this._regionManagerHelper.addToRegion();
+
+    setTimeout(this.subscribeToOpenFinEvents.bind(this), 1000);
```
zeala: @KeithWoods do we have an event that is dispatched when the DOM is ready? OpenFin has an internal listener for this even when it sets up the desktop API. the 'init' event is dispatched too early, so I can't use that one.

fdore:
As a general comment, I don't think it is the models' responsibility to integrate with openfin. I would rather move the integration to the service layer.

src/client/src/ui/spotTile/model/spotTileModel.js
```diff
@@ -161,8 +162,32 @@ export default class SpotTileModel extends ModelBase {
       this._log.warn(`Ignoring execute request as we can't trade with pricing and execution down`);
     }
   }
+  
+  _subscribeToClosePositionRequests(){
```
fdore: I would move this logic to the execution service, given we don't seem to interact with the spot tile at all.

src/client/src/ui/blotter/model/blotterModel.js
```diff
@@ -55,6 +56,13 @@ export default class BlotterModel extends ModelBase {
     this._regionManagerHelper.popout('Blotter', 850, 280);
   }
 
+  subscribeToOpenFinEvents(){
```
fdore: I would move that to the blotter service

src/client/src/services/mappers/priceMapper.js
```diff
@@ -29,6 +30,15 @@ export default class PriceMapper {
     );
   }
 
+  static mapToStopPriceDto(spotPrice:SpotPrice):Object{
```
fdore: should be mapToSpotPriceDto

Issue #357, opened 2016-05-18
Docking and snapping of spot tiles


odeheurles:
OpenFin have a script which manages snapping and docking of windows which can be found here: https://github.com/openfin/snap-and-dock

The window manager does not support windows which are resizable, so ideally we should prevent the docking of tiles with the main window (ie. only allow tiles to be docked to each other). We will also need a button on the tiles to undock them (only visible when you mouse over a tear-out window).

We know that there are edge cases that the window management script does not handle, this is fine - we will just have to be aware of them when we do a demo.

fdore:
Spot tiles can now be docked to other tiles.

PR #359, opened 2016-05-19
Openfin docking and snapping of spot tiles

fdore:
#357


src/client/src/system/openFin/dockingManager.js
```diff
@@ -0,0 +1,843 @@
+// Source: https://github.com/openfin/snap-and-dock
```
KeithWoods: perhaps this would make more sense in a `lib` folder that sits beside `node_modules` then consume it via an `import`. Just makes it more obvious that it's not ours and we shouldn't touch it.

What do you think?

fdore: Was kind of thinking that as well.. will update

Issue #367, opened 2016-05-20
IE issues


qooroo:
- [x] Console error: `SCRIPT5007: Unable to get property 'contains' of undefined or null reference app.js (4,17460)`
ESP subsequently stops the app: `Event router halted due to previous error [TypeError: Unable to get property 'contains' of undefined or null reference]`

- [x] Tile layout issue:
![image](https://cloud.githubusercontent.com/assets/1429601/15428747/22ea7bf4-1e95-11e6-965a-c9604aa56e1c.png)

- [x] Analytics chart colouring:
![image](https://cloud.githubusercontent.com/assets/1429601/15428763/37c96346-1e95-11e6-85b1-9893213596ed.png)

- [x] PnL / Positions bars not showing in analytics panel

- [ ] Popouts don't work

- [ ] Blotter dates render as NaN
![image](https://cloud.githubusercontent.com/assets/1429601/15429436/de1e01c2-1e98-11e6-9601-d8836f15032f.png)

fdore:
the console error seems related to chartgradient js which I dont think we even use...

PR #368, opened 2016-05-20
#367 added a polyfill, fixed price layout

zeala:


src/client/src/services/mappers/tradeMapper.js
```diff
@@ -56,4 +56,22 @@ export default class TradeMapper {
         throw new Error(`Unknown trade status ${statusDto}`);
     }
   }
+
+  _parseDate(dateStr):Date{
```
fdore: lets fix the date format on the server

fdore:
@zeala Can you run npm shrinkwrap and commit the updated shrinkwrap file?

src/client/src/ui/spotTile/views/spotTileView.jsx
```diff
@@ -66,14 +77,14 @@ export default class SpotTileView extends ViewBase {
...
-            <i className='popout__controls spot-tile__icon--tearoff glyphicon glyphicon-new-window'
+            <i className={newWindowClassName}
                onClick={() => router.publishEvent(this.props.modelId, 'popOutTile', {})}/>
           </div>
           {notification}
           {priceComponents}
           <NotionalInput
               className={notionalInputClass}
-              notional={model.notional}
+              notional={model.notional}l
```
fdore: you have a 1 hanging there

src/client/src/ui/analytics/views/analyticsView.jsx
```diff
@@ -43,12 +44,24 @@ export default class AnalyticsView extends ViewBase {
     if (!model.isAnalyticsServiceConnected)
       return <span />;
 
+    let browser = new userAgentParser().getBrowser().name;
```
fdore: I would pbly add a canPopout method in the popoutService so we dont have to repeat that logic in every view, and then put that on the view model instead of the view itself

zeala: Yes, i was going to do that today. now that we've got a bit more time to do it properly

src/client/src/ui/common/modelBase.js
```diff
@@ -36,6 +37,11 @@ export default class ModelBase extends DisposableBase {
     });
   }
 
+  isIE(){
```
fdore: can you rename that to canPopout?

PR #370, opened 2016-05-20
Persist open windows

fdore:
#338


src/client/src/ui/analytics/model/analyticsModel.js
```diff
@@ -56,6 +58,10 @@ export default class AnalyticsModel extends ModelBase {
     _log.info(`Analytics model starting`);
     this._subscribeToConnectionStatus();
     this._regionManagerHelper.addToRegion();
+
+    if (this._regionManager.shouldPopoutFromRegion(this._regionName, this._modelId)) {
+      this.router.publishEvent(this._modelId, 'popOutAnalytics', {});
```
zeala: It looks like this could be moved into the base model class, so we don't have duplications in all three models.

fdore: created new PR https://github.com/AdaptiveConsulting/ReactiveTraderCloud/pull/377

PR #374, opened 2016-05-23
sort trade descendingly on load

### LINKED ISSUE #369: On load, blotter should order the trades by Id descending

fdore:
Fixes #369

### DIFF
--- a/src/client/src/ui/blotter/model/blotterModel.js
+++ b/src/client/src/ui/blotter/model/blotterModel.js
@@ -64,8 +64,8 @@ export default class BlotterModel extends ModelBase {
           (tradesUpdate:TradesUpdate) => {
             let trades = tradesUpdate.trades;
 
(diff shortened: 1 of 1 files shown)


src/client/src/ui/blotter/model/blotterModel.js
```diff
@@ -64,8 +64,8 @@ export default class BlotterModel extends ModelBase {
           (tradesUpdate:TradesUpdate) => {
             let trades = tradesUpdate.trades;
 
-            if (tradesUpdate.isStateOfTheWorld){
-              this.trades = this.trades.concat(tradesUpdate.trades);
+            if (tradesUpdate.isStateOfTheWorld) {
+              this.trades = this.trades.concat(_.sortBy(tradesUpdate.trades, 'tradeId').reverse());
```
zeala: Is it possible that if multiple batches of trades are  received the final sort order might be incorrect?

PR #382, opened 2016-05-24
Use DateTime to store dates

JPWatson:


src/server/Adaptive.ReactiveTrader.Server.Common/TradeMapper.cs
```diff
@@ -16,7 +16,7 @@ public static TradeDto ToDto(this Trade t)
                 Direction = t.Direction,
                 SpotRate = t.SpotRate,
                 TradeDate = t.TradeDate,
-                ValueDate = t.ValueDate,
+                ValueDate = "SP. " + t.ValueDate.Day + " " + t.ValueDate.ToString("MMM"),
```
fdore: that should stay a date - formatting is done on the client

JPWatson: done

PR #401, opened 2016-06-01
add LimitChecker ClickOnce installer

fdore:
#46


fdore:
There is no CI for this branch, as this is the website content - so we need to ignore the checks

PR #434, opened 2016-06-13
lift canPopout method from base model 

### LINKED ISSUE #433: Lift canPopout() from ModelBase
Looks like some specific edgecase code has found it's way into ModelBase:

```
  canPopout(){
    let browser = new userAgentParser().getBrowser().name;
    return browser.indexOf('IE') !== -1;
  }
```

This should be lifted out of there, it's not in any way common to all sub types of ModelBase. 

Lets wrap the above concept into a static class named `Environment`, it should live in `system`. That can have a static property or method called `IsRunningInIe`. Then we can put specific properties in the models that need it, e.g.:

```
get canPopout() {
  return Environment.IsRunningInIe;
}
```

fdore:
Fixes #433

### DIFF
--- a/src/client/src/system/environment.js
+++ b/src/client/src/system/environment.js
@@ -1,0 +1,14 @@
+import userAgentParser from 'ua-parser-js';
+
+let isRunningInIE = null;
+
+export default class Environment {
+
+  static get isRunningInIE() {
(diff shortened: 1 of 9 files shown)


KeithWoods:
Thanks for fixing this.

PR #451, opened 2016-06-28
Update to .NET Core 1.0 RTM

qooroo:
note that the dotnet cli tooling is still in preview as `1.0.0-preview2-003121`


qooroo:
closes #450

PR #526, opened 2017-01-15
Upgrade to .NET Core 1.1

### LINKED ISSUE #525: Upgrade to .NET Core 1.1
The current websocket implementation is hacked together as .NET core websockets were not available on Unix in 1.0.

This is now available in 1.1 so we should upgrade.

Tasks:
- [x] Upgrade to dotnetcore 1.1 SDK
- [x] Upgrade dependent libraries
- [x] Remove PCL imports
- [ ] Update dotnet base image
- [ ] Update server image

See:
https://github.com/dotnet/corefx/pull/9258
https://github.com/dotnet/corefx/pull/9412

qooroo:
closes #525

now using the WebSocket implementation from the framework 🎉


deploy/config
```diff
@@ -58,8 +58,8 @@ nginxContainer_minor="0"
 nginxContainer="weareadaptive/nginx:$nginxContainer_major.$nginxContainer_minor"
 
 # dotnet
-dotnetversion="dotnet-dev-1.0.0-preview2-003121"
-dotnetContainer_major="0"
+dotnetversion="dotnet-dev-1.0.0-preview2.1-003177"
```
tdeheurles: @qooroo it's the container that was named server before ?

qooroo: No this is the .Net core base container. The server one is based from this.

We probably don't need this base container now that there's an official .Net core image from MS.

tdeheurles: I don't think using an official image is better. We don't know how they will manage that.

tdeheurles:
It's fine to merge for the devops part.
I let others validate the .NET one

deploy/config
qooroo: Yeah OK, I guess it doesn't cost us very much to maintain this.

PR #529, opened 2017-02-07
add kubernetes services to expose broker and eventstore

qooroo:


deploy/kubernetes/deploy
```diff
@@ -63,6 +66,8 @@ cd ..
   # service
   ./kubectl create -f manifests/broker-svc.yml 
   ./kubectl create -f manifests/eventstore-svc.yml 
+  ./kubectl create -f manifests/brokergate-svc.yml
```
JPWatson: do these break if you have generated the svc manifest before?

qooroo: nope should be fine

PR #537, opened 2017-02-26
use docker-compose for local docker run

tdeheurles:
Running reactive-trader locally will be done by docker-compose instead of custom code.


tdeheurles:
The version of docker-compose on CircleCi is an old one, the docker-compose file version 2 is not supported.

tdeheurles:
Docker-Compose file v1 doesn't support network. The network was the resource used to fix local build. So for now it doesn't work.

Supporting both solutions could an option

Issue #555, opened 2017-06-05
Upgrade to Webpack 3


qooroo:

bhavesh-desai-scratch:
https://github.com/AdaptiveConsulting/ReactiveTraderCloud/pull/570#pullrequestreview-52943761

PR #570, opened 2017-07-27
update webpack

bhavesh-desai-scratch:
This PR upgrades our build to webpack 3.

We can further improve the build as time goes on. We need to simply the build and make it a bit cleaner in regards to prod and dev configuration.


qooroo:
#555

PR #574, opened 2017-08-06
Rewrite rtc deployment

tdeheurles:


deploy/minikube/Start-Minikube.ps1
```diff
@@ -0,0 +1,98 @@
...
+    $web_client.DownloadFile($minikube_url, $config_minikube)
+    if ($?) {
+        Write-Host "minikube.exe downloaded`n"
+    } else {
+        Throw "Unable to download minikube from $minikube_url"
+    }
+}
+
+# Create Hyper-V switch
+$ethernet_adapter_name = "Ethernet"
```
qooroo: as discussed, this breaks on wifi

deploy/minikube/Start-Minikube.ps1
```diff
@@ -0,0 +1,98 @@
...
+
+# Remove kubectl message
+& $config_minikube config set WantKubectlDownloadMsg false
+
+# start
+Write-Host -f "Cyan" "Starting minikube:"
+& $config_minikube start `
+    --vm-driver="hyperv" `
+    --hyperv-virtual-switch="$($external_vm_switchs.Name)" `
+    --memory 6144
```
qooroo: also prefer to lower this if possible - perhaps query the current host free memory amount first?

tdeheurles: perhaps query the current host free memory amount first?
==> good idea. Must no be too hard to implement

PR #581, opened 2017-08-17
enable client to connect to local server

MaciejAdaptive:


src/client/config/local.config.json
```diff
@@ -1,4 +1,5 @@
 {
   "overwriteServerEndpoint": true,
-  "serverEndPointUrl": "localhost"
-}
\ No newline at end of file
+  "serverEndPointUrl": "localhost",
+  "serverPort": 8000
```
tdeheurles: What is this configuration ?

We have different way to access backend. If you develop without docker, if you develop with docker if you develop with minikube.
The production way is to access backend on port 80/443. It's also the way to access minikube (local kubernetes).

We need a hangout with front/backend/devops to me.

MaciejAdaptive: In local, if broker port is not provided (8000), default port (80) is used (see autobahnConnectionProxy.ts). This change affects local env only. I don't know anything yet about docker side, I'm open to any suggestions. 

As far as I know, FE team is connected to cloud while working locally.

tdeheurles: We will need to find a better naming convention in the future as we have multiple way to dev in local (installed on machine / docker / minikube / ...)

But this is an issue we did have in the past.

To me it's ok for this.

PR #604, opened 2017-09-26
Refactoring to replace inheritance with composable functions

celikmus:
Being replaced with: https://github.com/AdaptiveConsulting/ReactiveTraderCloud/pull/627

I tidied up appBootstrapper and services to move away from inheritance from ServiceBase and DisposableBase... We don't really need inheritance for those, and the code flows easier as functions. As ServiceBase is not used anymore, I have removed it altogether.

While I removed inheritance, we still use ServiceClient by instantiating in factory functions. This means I haven't altered any Rx behaviour, in fact no behaviour at all. I have fixed the impacted testcases, while there are some existing failing tests for ServiceClient... I am planning to tackle with ServiceClient next, and I will tackle those there...

I used prettier on the files that I worked on... I have also taken out some tsLint rules; which reduced majority of the pre-existing tsLint errors.


tdeheurles:
good for deployment part

src/client/src/services/compositeStatusService.ts
```diff
@@ -1,100 +1,86 @@
...
-    this.pricingService = pricingService
-    this.referenceDataService = referenceDataService
-    this.blotterService = blotterService
-    this.executionService = executionService
-    this.analyticsService = analyticsService
-    this.serviceStatusStream = this.createServiceStatusStream()
-    this.currentServiceStatusLookup = new ServiceStatusLookup()
+): Object {
+  const disposables = new Subscription()
+  const addDisposable = disposable => {
```
fdore: we are not using esp-js anymore, so we dont need that anymore.

src/client/src/system/service/connection.ts
```diff
@@ -37,7 +46,9 @@ export default class Connection extends DisposableBase {
     this.autoDisconnectDisposable = new SerialSubscription()
     this.connectionUrl = ''
     this.connectionType = ConnectionType.Unknown
-    this.addDisposable(this.autoDisconnectDisposable)
+    const prevProto = Object.getPrototypeOf(this.autoDisconnectDisposable)
+    prevProto.dispose = prevProto.unsubscribe
```
fdore: was it because of esp-js?

silviuadaptive:
Replaced with https://github.com/AdaptiveConsulting/ReactiveTraderCloud/pull/627

PR #627, opened 2018-01-29
Celikmus refactor to replace inheritance with composable functions

silviuadaptive:
This is a copy of the other PR because I don't have access to his repo. The only difference in this PR is that I updated master.

From original PR:

src: https://github.com/AdaptiveConsulting/ReactiveTraderCloud/pull/604

I tidied up appBootstrapper and services to move away from inheritance from ServiceBase and DisposableBase... We don't really need inheritance for those, and the code flows easier as functions. As ServiceBase is not used anymore, I have removed it altogether.

While I removed inheritance, we still use ServiceClient by instantiating in factory functions. This means I haven't altered any Rx behaviour, in fact no behaviour at all. I have fixed the impacted testcases, while there are some existing failing tests for ServiceClient... I am planning to tackle with ServiceClient next, and I will tackle those there...

I used prettier on the files that I worked on... I have also taken out some tsLint rules; which reduced majority of the pre-existing tsLint errors.


src/client/src/types/referenceDataService.ts
```diff
@@ -0,0 +1,7 @@
+import { Observable } from 'rxjs/Rx'
```
zeala: So, we've got 2 files with the same name now in the project - referenceDataService.ts ?

src/client/src/services/referenceDataService.ts
```diff
@@ -1,88 +1,105 @@
...
-    return this.currencyPairCache[symbol]
+  const referenceDataStreamConnectable = referenceDataStream().publish()
+  const currencyPairCache = {
+    hasLoaded: false
   }
 
-  getCurrencyPairUpdatesStream() {
-    return this.referenceDataStreamConnectable
+  const addDisposable = disposable => {
+    // esp-js is expecting a dispose method
```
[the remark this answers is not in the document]
silviuadaptive: Fixed.

src/client/src/types/referenceDataService.ts
silviuadaptive: I would suggest to keep it like this now and fix it later. Are you happy with that? I'm doing another story that will clean up many things.

zeala: sounds good

silviuadaptive: thanks

src/client/src/services/pricingService.ts
```diff
@@ -1,61 +1,79 @@
...
-            },
-            (err) => { o.error(err) },
-            () => { o.complete() },
-          )
+const createSpotPriceStream = (serviceClient, request: Request) => {
+  const observable = Observable.create(clientObserver => {
+    log.debug(`Subscribing to spot price stream for [${request.symbol}]`)
+    let lastPrice = null
+    const innerObserver = {
+      next: price => {
```
evanrs: No need to use a fat arrow, there's no prototypical scope to retain.
```js
const innerObserver = {
  next(price) {
    lastPrice = price
    clientObserver.next(price)
  }
  …
  …
}
```

src/client/src/services/pricingService.ts
```diff
@@ -1,61 +1,79 @@
...
+      complete: () => {
+        clientObserver.complete()
+      }
+    }
+    const retryWithPolicyErrorCb = (err, willRetry) => {
+      if (willRetry && lastPrice !== null) {
+        // TODO: remove is stale price fully supported
+      }
+    }
+    const retryWithPolicyArgs = [
```
evanrs: Does this make it easier to read? It looks like indirection-by-alias to me.

src/client/src/services/pricingService.ts
silviuadaptive: Ok, I will do that. I think is more about a style preference. The compiled code in [production is exactly the same](https://babeljs.io/repl/#?babili=false&browsers=&build=&builtIns=false&code_lz=MYewdgzgLgBAlmMBTATgeQEYVQN1TAXhgG8AoGGZADygAoAHFOYJAShPIpgBsBDaAApMWhGI2ZJOFYNzhIwUTNhR4UAOmp1xLVpwC-pA6VCRYCZOiy5UAJlFkKmgFxjhSQgD4OXHvyhCJUW1JHxk5BSVrdU0GN10KAz0gA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&lineWrap=true&presets=es2015%2Creact%2Cstage-2&prettier=false&targets=&version=6.26.0&envVersion=).

src/client/src/services/pricingService.ts
silviuadaptive: Yes, I think is easier to read the first code vs the second one. What do you think?

```js
    const subscription = serviceClient
      .createStreamOperation(getPriceUpdatesOperationName, request)
      // we retry the price stream forever, if it errors (likely connection down) we pump a non tradable price
      .retryWithPolicy(...retryWithPolicyArgs)
      .map(adaptDTO)
      .subscribe(innerObserver)
```

```js
    const subscription = serviceClient
      .createStreamOperation(getPriceUpdatesOperationName, request)
      // we retry the price stream forever, if it errors (likely connection down) we pump a non tradable price
      .retryWithPolicy(		
         RetryPolicy.indefiniteEvery2Seconds,		
         getPriceUpdatesOperationName,		
         Scheduler.async,		
         (err, willRetry) => {		
           if (willRetry && lastPrice !== null) {		
             // TODO: remove is stale price fully supported		
           }		
         },		
       )
      .map(adaptDTO)
      .subscribe(innerObserver)
```

src/client/src/appBootstrapper.tsx
```diff
@@ -16,94 +15,65 @@ import {
...
+    <Provider store={store}>
+      <OpenFinProvider openFin={openFin}>
+        <ShellContainer />
+      </OpenFinProvider>
+    </Provider>,
+    document.getElementById('root')
+  )
 }
 
 const runBootstrapper = location.pathname === '/' && location.hash.length === 0
```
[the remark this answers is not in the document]
silviuadaptive: I'm not sure why in this particular project is needed, but sometimes you want to test the whole app, therefore you need tear down and tear up. (is not doing this at the moment, obviously)
Another reason maybe has to do with the openFin, but again, I'm just guessing. Somebody with more experience might be able to help you better.

src/client/src/services/pricingService.ts
evanrs: Fair point! Thanks for including the example 😉

src/client/src/services/pricingService.ts
evanrs: I have an easier time parsing a sequence of expressions than I do with the mental book-keeping that comes with assignments.

It was also the topic of the day yesterday, indirection that it is—the forms it takes and its trade offs.

I feel variable names are often too hopeful in what they will communicate—and the transparency which we hope they'll have. The names you've chosen are clear and concise, but for me the entailments are lost—I have to jump back and forth. It feels a lot like a computer paging when it runs out of memory 😅

silviuadaptive: Depends at what level you read. I'm thinking that if you want to know what it does you should read like this:

```
   const subscription = serviceClient
      .createStreamOperation(getPriceUpdatesOperationName, request)
      .retryWithPolicy(...retryWithPolicyArgs)
      .map(adaptDTO)
      .subscribe(innerObserver)
```

If you are curious to know how certain thing is implemented, you click on it. I know it has some trade-offs, but normally you want to know what is doing from a higher point of view. I'm happy to change it if you find it too hard to read.

From the previous example you only lose this information `indefiniteEvery2Seconds`, but that could be added to the `retryWithPolicyArgs` variable.

Let me know your thoughts :)

Issue #629, opened 2018-01-30
Make RXJS code more readable and easier to follow/debug


silviuadaptive:
Files that might need some work:

 - src/client/src/system/service/serviceClient.ts
 - src/client/src/services/referenceDataService.ts
 - src/client/src/services/blotterService.ts
 - src/client/src/system/service/serviceObservableExtensions.ts

Other stuff to do:

 - Add comments to explain a bit what the epics do. For example `referenceServiceEpic` is not clear what is doing except dispatching an init type action.
 - https://github.com/AdaptiveConsulting/ReactiveTraderCloud/pull/627#discussion_r164693080

zeala:
@silviuadaptive - this is a bit too generic and needs a discussion. Re comments in the code -  the code should be self-documented. We don't keep comments in the code as a rule with some exceptions of implementing unclear workarounds.

I suggest splitting this issue into documenting the streaming services in a Confluence page and a separate one for improving typings in the code.

I am closing this issue.

zeala:
Re-opening the issue after clarifying the requirements.

dwalker11:
Hey @zeala, out of curiosity do we have any documentation on the overall architecture of this application?

zeala:
closing as fixed

PR #634, opened 2018-02-01
Rxjs refactor: reference data service

silviuadaptive:
Related to the issue: #629

### DIFF
--- a/src/client/src/ui/blotter/blotterOperations.ts
+++ b/src/client/src/ui/blotter/blotterOperations.ts
@@ -1,14 +1,15 @@
-import * as _ from 'lodash'
 import { createAction } from 'redux-actions'
 
 import { regionsSettings } from '../../regions/regionsOperations'
 import { ACTION_TYPES as REF_ACTION_TYPES } from '../../referenceOperations'
+import { Trade } from '../../types'
+import * as keyBy from 'lodash.keyby'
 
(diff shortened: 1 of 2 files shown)


src/client/src/ui/blotter/blotterOperations.ts
```diff
@@ -19,27 +20,33 @@ export const blotterServiceEpic = (blotterService$, openFin) => (action$, store)
...
-      if (payloadTrades.length === 1) {
-        const tradeKey = payloadTrades[0]['tradeId']
-        if (state.trades[tradeKey]) {
-          newState.trades = _.omit(state.trades, tradeKey)
-        }
-      }
-      const trades = _.mapKeys(_.values(payloadTrades), 'tradeId')
+    case ACTION_TYPES.BLOTTER_SERVICE_NEW_TRADES:
+      const newTradesById = keyBy(action.payload.trades, `tradeId`)
       return {
```
zeala: this logic might still be required

silviuadaptive: they had that comment because they are doing different things if one trade is added vs multiple trades are added. The updated code can handle 1 or multiple trades with the same logic. 
Not sure why when they have only one trade, they remove it from the state and then add it again, instead of just replacing it directly.

PR #644, opened 2018-02-06
RT - work on normalising application state, code cleanup, typings, code reuse

zeala:


src/client/src/redux/actions/spotTileActions.ts
```diff
@@ -0,0 +1,113 @@
+import { createAction } from 'redux-actions'
```
bhavesh-desai-scratch: We need to break this file up. It has Epics as well as actions

src/client/src/redux/reducers/combineReducers.ts
```diff
@@ -0,0 +1,30 @@
+import { combineReducers } from 'redux'
```
bhavesh-desai-scratch: We need to name this file better. Maybe root reducer

src/client/src/redux/reducers/currencyPairsReducer.ts
```diff
@@ -0,0 +1,20 @@
+import { CurrencyPair } from '../../types/index'
+import { ACTION_TYPES } from '../actions/referenceDataActions'
+
+interface CurrencyPairReducerState {
+  [id: string]: CurrencyPair
+}
+
+export const currencyPairReducer = (state: CurrencyPairReducerState = {}, action): CurrencyPairReducerState => {
+  switch (action.type) {
```
bhavesh-desai-scratch: Better to use a map then a switch

src/client/package.json
```diff
@@ -87,6 +87,7 @@
     "redux-observable": "0.17.0",
     "redux-persist": "5.6.5",
     "redux-persist-crosstab": "3.7.0",
+    "reselect": "3.0.1",
```
dwalker11: 👍

PR #651, opened 2018-02-08
SpotTileData - action/epics code restructure

zeala:
Moved reducer/actions/epics inside spotTile directory
Split actions file into epics and actions
Explicitly assigning SpotTile props to improve clarity


zeala:
@dwalker11 - a good question. yes, we are going to keep reducer, actions, epics separate, but they should be close to their functional areas. This PR is addressing the comments from @bhavesh-desai-scratch about that. For now, it will make it easier to see each individual area in the application. Once we sort out issues with code clarity and structure we can discuss/see what's best.

dwalker11:
@zeala, understood. I'll implement that pattern into any changes I make.

PR #683, opened 2018-02-26
AgGrid blotter with simplified filtering

zeala:
The following features are included in this PR
- AgGrid-based blotter - styled
- Simplified filtering - only the quick filter/search 
- Blotter toolbar/status bar
- Blotter is resizable in a popup (OpenFin)
- Split view (blotter & pricing view)


zeala:
closing as invalid, will create a new one

PR #685, opened 2018-02-27
AgGrid blotter with quick filter

zeala:
The following features are included in this PR
- AgGrid-based blotter - styled
- Simplified filtering - only the quick filter/search 
- Blotter toolbar/status bar
- Blotter is resizable in a popup (OpenFin)
- Split view (blotter & pricing view)


src/client/src/ui/blotter/BlotterContainer.tsx
```diff
@@ -29,16 +28,15 @@ class BlotterContainer extends React.Component<BlotterContainerProps, {}> {
...
       <div className="shell_workspace_blotter">
-        <Blotter trades={trades}
-                 currencyPairs={this.props.currencyPairs}
-                 canPopout={Environment.isRunningInIE()}
-                 size={this.props.size}
-                 isConnected={this.props.isConnected}
-                 onPopoutClick={this.props.onPopoutClick(openFin)}/>
+        <AgGridBlotter rows={ gridRows }
+                       onPopoutClick={popoutClick}
+                       canPopout={true}/>
```
dwalker11: I see that you've removed the check for IE11, is it no longer necessary to restrict popouts on IE11 because we are using ag-grid?

zeala: this is a good spot. no, this is not intended and will cause a regression. will fix. thanks

Issue #670, opened 2018-02-22
EUR/JPY should ALWAYS times out and then the execution succeeds


amilanan:
Timeout message is not displayed on EUR/JPY tile.

zeala:
fixed

PR #686, opened 2018-02-27
Show error tile on execution error

zeala:
fix for #670


src/client/src/ui/spotTile/reducer.ts
```diff
@@ -25,7 +25,7 @@ export const spotTileDataReducer = (state: any = {}, action) => {
       return updateSpotTile(state, payload.CurrencyPair, { isTradeExecutionInFlight: true })
 
     case ACTION_TYPES.TRADE_EXECUTED:
-      const symbol = payload.trade.symbol
+      const symbol = payload.trade.symbol || payload.trade.CurrencyPair
```
bhavesh-desai-scratch: seems a bit strange that an action can have two types of payload.

zeala: yes, very. but this will require a more detailed investigation

PR #693, opened 2018-03-05
688 - addressing spot date formatting

jem77:


src/client/src/ui/notification/TradeNotification.tsx
```diff
@@ -13,6 +15,8 @@ export interface TradeNotificationProps {
 export default class TradeNotification extends React.Component<TradeNotificationProps, {}> {
   render() {
     const trade = this.props.message
+    const formattedValueDate = trade ? `SP. ${moment(trade.valueDate).format(SPOT_DATE_FORMAT)}` : ''
+
```
zeala: we should probably move it into a util method and use it in all (now 3) places. thanks

PR #712, opened 2018-04-12
Prepare RT for create-react-app

bhavesh-desai-scratch:
## Purpose 
Maintaining Reactive traders build is becoming cumbersome. We should leverage the excellent work done by create-react-app and delete most our build. This change will make it easier to keep our tools up to date and have the support of the opensource community as time passes.

### Key changes

- Make sure all code is under source
- We are going to stop using sass and CSS pre-processors shortly when we reskin the application. 
- Create-react-app advocates compiling SASS separately to the main build. 
- To make things easier, I have moved all the SCSS files to a single folder
- We will be replacing these styles with a JS in CSS solution shortly


src/client/package.json
```diff
@@ -58,6 +58,7 @@
   },
   "license": "MIT",
   "dependencies": {
+    "@types/openfin": "29.0.0",
```
fdore: would move to devDependencies for consistency

src/client/src/system/openFin/openFin.ts
```diff
@@ -47,7 +46,7 @@ export default class OpenFin {
           currentWindow.restore(() => currentWindow.bringToFront(
             () => log.info('Window brought to front.'),
             err => log.error(err),
-          ), err => log.error(err))
+          ), () => log.error())
```
fdore: why did you remove the error?

src/client/src/ui/styles/scss/blotter.scss
```diff
@@ -1,5 +1,5 @@
-@import '../spotTile/variables';
-@import './variables';
+@import 'variables';
+@import 'variables';
```
fdore: duplicate import

tdeheurles:
Nothing for Devops, fine for me

src/client/src/system/openFin/openFin.ts
bhavesh-desai-scratch: Does not exist according to TS definitions

bhavesh-desai-scratch: Fixed

fdore: It does according to the docs http://cdn.openfin.co/jsdocs/stable/fin.desktop.Window.html#restore

PR #713, opened 2018-04-18
prepare config for create-react-app

bhavesh-desai-scratch:
We need to change the way our config works so that is more in line with how create-react-app does it.

1. We must move the config to source as it is actually javascript code.
2. The build process knows nothing about these config files. It just uses environment variables.
3. We should change the build to just use env files once we start using create-react-app


bhavesh-desai-scratch:
No longer needed

PR #716, opened 2018-04-30
Strongly type core connection classes and cleanup

bhavesh-desai-scratch:
* Core classes compile under strictest TS settings
* Remove redundant code
* Introduce interfaces for testing
* Patch bad Autobahn TS files
* Correct imports for next version of TS
* Type the logger and imporove output


bhavesh-desai-scratch:
no good

PR #717, opened 2018-04-30
Strongly type cleanup service classes

bhavesh-desai-scratch:
- Core classes compile under strictest TS settings
- Remove redundant code
- Introduce interfaces for testing
- Patch bad Autobahn TS files
- Correct imports for next version of TS
- Type the logger and improve output

### DIFF
--- a/src/client/package.json
+++ b/src/client/package.json
@@ -46,6 +46,9 @@
     ]
   },
   "dependencies": {
+    "@types/enzyme": "^3.1.10",
+    "@types/enzyme-adapter-react-16": "^1.0.2",
+    "@types/react-test-renderer": "^16.0.1",
     "ag-grid": "15.0.0",
     "ag-grid-react": "15.0.0",
     "animate.css": "3.4.0",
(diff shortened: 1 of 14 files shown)


src/client/package.json
```diff
@@ -46,6 +46,9 @@
     ]
   },
   "dependencies": {
+    "@types/enzyme": "^3.1.10",
+    "@types/enzyme-adapter-react-16": "^1.0.2",
+    "@types/react-test-renderer": "^16.0.1",
```
fdore: would move to devDependencies

src/client/package.json
```diff
@@ -76,6 +81,7 @@
     "redux-persist-crosstab": "3.7.0",
     "reselect": "3.0.1",
     "rxjs": "5.5.9",
+    "source-map-explorer": "^1.5.0",
```
fdore: ditto

src/client/src/system/logger.ts
```diff
@@ -1,26 +1,23 @@
...
-import Guard from './guard'
-
-const levels = {
-  verbose: 0,
-  debug: 1,
-  info: 2,
-  warn: 3,
-  error: 4,
+enum levels {
+  verbose = 'debug',
```
fdore: should that be verbose = 'verbose'?

PR #719, opened 2018-05-03
Upgrade rxjs in compat mode

bhavesh-desai-scratch:


keshaldesai:
Error with the spotTile reducer when it receives a case of 'TRADE_EXECUTED'

keshaldesai:
The error notification for when trades timeout is broken.

PR #721, opened 2018-05-16
cleanup execution service

bhavesh-desai-scratch:
- Stop using inner subscriptions
- move openfin out of the execution service
- type the limit checker
- use race for timeout

### DIFF
--- a/src/client/src/services/executionService.ts
+++ b/src/client/src/services/executionService.ts
@@ -1,14 +1,12 @@
-import { Observable, of, race, timer } from 'rxjs'
-import { map, mapTo, mergeMap, take, tap } from 'rxjs/operators'
+import { merge, Observable, of, timer } from 'rxjs'
+import { map, mapTo, mergeMap, take, takeUntil, tap } from 'rxjs/operators'
 import { logger } from '../system'
 import { ServiceClient } from '../system/service'
-import { TradeSuccessResponse } from '../types'
 import {
   createExecuteTradeResponse,
   createExecuteTradeResponseForError,
   ExecuteTradeRequest,
-  ServiceConst,
-  TradeErrorResponse
+  ServiceConst
 } from '../types'
 import { mapFromTradeDto } from './mappers'
 import { TradeRaw } from './mappers/tradeMapper'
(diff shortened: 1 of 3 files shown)


src/client/src/services/executionService.ts
```diff
@@ -32,80 +23,70 @@ const EXECUTION_REQUEST_TIMEOUT_MS = 30000
...
 
   executeTrade(executeTradeRequest: ExecuteTradeRequest) {
-    return new Observable<ExecuteTradeResponse>(obs => {
-      log.info('executing: ', executeTradeRequest)
-      const disposables = new Subscription()
+    return this.limitChecker(executeTradeRequest).pipe(
+      tap(() => log.info('executing: ', executeTradeRequest)),
+      take(1),
+      mergeMap(limitBreached => {
+        if (!limitBreached) {
```
fdore: should that not be the opposite?

src/client/src/services/executionService.ts
```diff
@@ -32,80 +23,70 @@ const EXECUTION_REQUEST_TIMEOUT_MS = 30000
 export default class ExecutionService {
   constructor(
     private readonly serviceClient: ServiceClient,
-    private readonly openFin: OpenFin
+    private readonly limitChecker: (
```
fdore: 👍

src/client/src/services/executionService.ts
```diff
@@ -32,80 +23,70 @@ const EXECUTION_REQUEST_TIMEOUT_MS = 30000
...
-          }
-          const request = this.serviceClient
-            .createRequestResponseOperation<
-              RawTradeReponse,
-              ExecuteTradeRequest
-            >(
-              ServiceConst.ExecutionServiceKey,
-              'executeTrade',
+        // After a longer period of time we know a trade is not coming back
+        const lastTimeout = timer(EXECUTION_REQUEST_TIMEOUT_MS).pipe(
```
fdore: I would pbly share the timer...

fdore: Would pbly share the timer..

bhavesh-desai-scratch: Each request needs it own timer, right?

src/client/src/services/executionService.ts
bhavesh-desai-scratch: Your right. Inverted the naming

src/client/src/services/executionService.ts
bhavesh-desai-scratch: Will leave in as discussed

PR #723, opened 2018-05-17
Clean up folder structure and simplify retry's

bhavesh-desai-scratch:


src/client/src/services/connectionStatusService.ts
```diff
@@ -1,9 +1,12 @@
...
-import { ConnectionEventType } from '../system/service/ConnectionFactory'
-import { ConnectionStatus } from '../system/service/connectionStatus'
-import { ConnectionType } from './../system/service/connectionType'
+import {
+  ConnectionEventType,
+  ConnectionStatus,
+  ConnectionType,
+  ServiceStub
+} from '../system'
+import { ConnectionEvent } from '../system'
```
dwalker11: Is there a reason for not combining these separate imports from '../system' into one statement?

PR #741, opened 2018-06-22
ARTP-89 Cleanup reducers

keshaldesai:
The reducers for spot tiles and regions are now in a split reducer pattern.

### DIFF
--- a/src/client/src/ui/common/regions/reducer.ts
+++ b/src/client/src/ui/common/regions/reducer.ts
@@ -1,33 +1,49 @@
 import { ACTION_TYPES as CONNECTION_ACTION_TYPES, DisconnectAction } from '../../../operations/connectionStatus'
 import { ACTION_TYPES, Region, RegionActions } from './actions'
 
-const changeRegionTearOffStatus = (state: RegionState, payload: Region, status: boolean) => ({
-  ...state,
-  [payload.id]: {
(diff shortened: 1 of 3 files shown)


src/client/src/ui/common/regions/reducer.ts
```diff
@@ -1,34 +1,50 @@
...
     case ACTION_TYPES.REGION_ADD:
-      return {
-        [action.payload.id]: action.payload,
-        ...state
-      }
     case ACTION_TYPES.REGION_ATTACH_WINDOW:
-      return changeRegionTearOffStatus(state, action.payload, false)
     case ACTION_TYPES.REGION_TEAROFF_WINDOW:
-      return changeRegionTearOffStatus(state, action.payload, true)
+      return { ...state, [action.payload.id]: regionReducer(state[action.payload.id], action) }
```
bhavesh-desai-scratch: Favour spreading code on to multiple lines if it makes it more readable.

bhavesh-desai-scratch: You do not need all these case statements here. Just send them down to the sub reducer.

keshaldesai: If we don't check for an appropriate action, the reducer would pass on actions that did not pertain to the regionReducer.

PR #745, opened 2018-06-26
ARTP-91 Migrate to portal tearoffs

keshaldesai:
- Create an environment provider 
- Create a desktop portal component
- Create a browser portal component
- Migrate blotter to use portals


src/client/src/main.tsx
```diff
@@ -11,6 +11,9 @@ import { OpenFin } from './services/openFin'
 import { AutobahnConnectionProxy, logger } from './system'
 import { User } from './types'
 import { OpenFinProvider, ShellContainer } from './ui/shell'
+import { EnvironmentProvider } from './ui/shell/EnvironmentProvider'
+import { BrowserWindow } from './ui/tearoff/BrowserPortal'
```
bhavesh-desai-scratch: We should name these better

src/client/src/main.tsx
```diff
@@ -11,6 +11,9 @@ import { OpenFin } from './services/openFin'
...
+import { DesktopWindow } from './ui/tearoff/DesktopPortal'
```
bhavesh-desai-scratch: we should be exposing this via an index.ts file

src/client/src/main.tsx
```diff
@@ -29,19 +41,18 @@ const appBootstrapper = () => {
...
   const store = configureStore(applicationDependencies)
   window.store = store
+
   ReactDOM.render(
     <Provider store={store}>
-      <OpenFinProvider openFin={openFin} isRunningInFinsemble={isRunningInFinsemble}>
-        <ShellContainer />
-      </OpenFinProvider>
+      <EnvironmentProvider value={environmentContext}>
+        <OpenFinProvider openFin={openFin}>
```
bhavesh-desai-scratch: We should use a ternary to do this rather then adding the wrapper when we do not need it.

src/client/src/ui/common/regions/actions.ts
```diff
@@ -3,9 +3,9 @@ import { RegionSettings } from '../../../types'
 
 export interface Region {
   id: string
-  isTearedOff: boolean
-  container: React.ComponentClass
-  settings: RegionSettings
+  isTearedOff?: boolean
+  container?: React.ComponentClass
```
bhavesh-desai-scratch: We should remove this? Unless this is a WIP step.

src/client/src/ui/shell/Shell.tsx
```diff
@@ -7,25 +7,54 @@ import { WorkspaceContainer } from '../workspace/'
...
+
+  popout = () => {
+    this.setState({ tornOff: true }, () => this.props.onPopout(blotterRegion))
+  }
+
+  popIn = () => {
+    this.setState({ tornOff: false }, () => this.props.onPopin(blotterRegion))
+  }
+
   static contextTypes = {
```
bhavesh-desai-scratch: Why do we still need this?

src/client/src/ui/tearoff/BrowserPortal.tsx
```diff
@@ -0,0 +1,201 @@
+import * as React from 'react'
```
bhavesh-desai-scratch: These two implementations need to be cleaned up. Factor out what is common and put them in another file

src/client/src/ui/common/regions/actions.ts
keshaldesai: WIP step

src/client/src/ui/tearoff/Portal.tsx
```diff
@@ -0,0 +1,150 @@
...
+  }
+
+  render() {
+    if (!this.state.mounted) {
+      return false
+    }
+    const { environment, config, desktopConfig, browserConfig } = this.props
+
+    const element = environment.isRunningDesktop ? (
+      <div id="popout-content-container">
```
bhavesh-desai-scratch: You can avoid repeating the ```poput-content-container``` by placing the ternary inside of it

src/client/src/ui/shell/Shell.tsx
```diff
@@ -6,84 +6,102 @@ import SidebarRegionContainer from '../sidebar'
...
+const portalProps = {
+  blotterRegion: {
+    title: 'Blotter',
+    config: {
+      name: 'blotter',
+      width: 850,
+      height: 450,
+      url: 'about:Blotter'
+    },
+    browserConfig: { center: 'screen' as 'screen' }
```
bhavesh-desai-scratch: I think this has a default so you can remove it for now

PR #767, opened 2018-07-25
Theming

el-dav:


src/client/.env.local
```diff
@@ -0,0 +1 @@
+REACT_EDITOR=code
```
evanrs: I object ☝️😛

zeala:
Hi, I've noticed that the split-view/resizer is missing in this branch. Is is still wip?
(I'll remove this image later, so it doesn't add unnecessary noise to the branch changes)
split-view

PR #773, opened 2018-08-01
[ARTP-151] Restyle blotter

keshaldesai:


src/client/src/ui/blotter/components/BlotterHeader.tsx
```diff
@@ -0,0 +1,96 @@
+import { GridApi } from 'ag-grid'
+import React, { Component } from 'react'
+import { styled } from 'rt-util'
+import { columnDefinitions } from './blotterUtils'
+import BlotterToolbar from './toolbar/BlotterToolbar'
+
+interface BlotterHeaderProps {
```
el-dav: just opinion but since these interfaces are only used for the one component they could be called 'Props' and 'State'

src/client/src/ui/blotter/components/BlotterHeader.tsx
```diff
@@ -0,0 +1,96 @@
...
+  align-items: center;
+  padding: 8px 8px 16px 20px;
+`
+
+const BlotterControls = styled('div')`
+  cursor: pointer;
+  color: ${({ theme: { text } }) => text.textSecondary};
+`
+
+const BlotterRight = styled('div')`
```
el-dav: Bhav added a Flex component to rt-components which could be useful here

src/client/src/ui/blotter/components/toolbar/BlotterToolbar.tsx
```diff
@@ -17,25 +18,28 @@ interface BlotterToobarState {
   themeName: string
 }
 
+const BlotterToolbarStyle = styled('div')`
```
el-dav: could use Flex component

src/client/src/ui/blotter/components/toolbar/QuickFilter.tsx
```diff
@@ -11,34 +11,80 @@ interface QuickFilterState {
   quickFilterText: string
 }
 
+const FilterIcon = () => (
```
el-dav: could use a filter icon from font-awesome and then use styled so it can adapt to the theme?

src/client/package.json
```diff
@@ -54,8 +54,9 @@
     ]
   },
   "dependencies": {
-    "ag-grid": "15.0.0",
-    "ag-grid-react": "15.0.0",
+    "@types/lodash.keyby": "^4.6.4",
```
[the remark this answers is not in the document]
keshaldesai: I agree with your suggestion. I think we need update our lodash package so we can do this.

PR #778, opened 2018-08-10
[ARTP-169] - Up Node version to LTS

tcdev00:
- Upgraded node version to LTS 8.11.3.
- Separated the npm install layer to quickly build the client.
- Added docker ignore file.

**Note:** To use node LTS alpine it will be needed to install python in the container. Alpine has been avoided in this task.

### DIFF
--- a/deploy/config
+++ b/deploy/config
@@ -41,9 +41,9 @@ eventstoreContainer_minor="0"
 eventstoreContainer="weareadaptive/eventstore:$eventstoreContainer_major.$eventstoreContainer_minor"
 
 # node
-vNode="8.11.3-alpine"
+vNode="8.11.3"
 nodeContainer_major="8"
-nodeContainer_minor="11.3-alpine"
+nodeContainer_minor="11.3"
 nodeContainer="node:$nodeContainer_major.$nodeContainer_minor"
 
 # nginx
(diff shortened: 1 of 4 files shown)


deploy/config
```diff
@@ -41,10 +41,10 @@ eventstoreContainer_minor="0"
...
 
 # node
-vNode="6.2.1"
-nodeContainer_major="6"
-nodeContainer_minor="2"
-nodeContainer="weareadaptive/node:$nodeContainer_major.$nodeContainer_minor"
+vNode="8.11.3"
+nodeContainer_major="8"
+nodeContainer_minor="11.3"
+nodeContainer="node:$nodeContainer_major.$nodeContainer_minor"
```
keshaldesai: I'm not sure what the best practice for this is in Docker, but you might be able to just pull `node:8`.
This would keep us up to date with bug fixes and features without breaking the API.

tcdev00: It makes a lot of sense, LTS will grant bug fixes, as long as we don't break the API I'm happy by going node:8.

tdeheurles: @tcdev00 this have been tested/validated ?

tcdev00: yes @tdeheurles , I have tested node:8 in local and in circleci and works pretty well.
Initially I tried to use an alpine image but it was failing in both ends. The non alpine version works properly.

PR #779, opened 2018-08-10
[ARTP-163] Style shell to match design

evanrs:


evanrs:
For various reasons we have merge conflict due to our branching structure and will orphan this branch.

PR #784, opened 2018-08-13
[ARTP-184] Expand theme type definition

evanrs:


src/client/src/rt-theme/themes.ts
```diff
@@ -109,15 +153,4 @@ export function createTheme({ primary, secondary }: PaletteMap, accents: Palette
   return _.merge(theme, overrides)
```
bhavesh-desai-scratch: Does a spread no work here?

src/client/src/rt-theme/themes.ts
```diff
@@ -1,53 +1,100 @@
...
+  shell: ColorPair
+  component: ColorPair
+
+  button: TouchableIntents
+
+  // Known extensible properties
+  backgroundColor?: Color
+  textColor?: Color
+
+  // TODO: remove after theme migration
```
bhavesh-desai-scratch: Can we date and initial any checked in TODO's plz

evanrs: How has this never occurred to me?? 

Yes, definitely.

src/client/src/rt-theme/themes.ts
evanrs: Will replace with a customizer fn.

PR #787, opened 2018-08-14
[ARTP-185] Move theme management to rt-theme

evanrs:


src/client/src/main.tsx
```diff
@@ -43,21 +45,22 @@ const appBootstrapper = () => {
...
+  const store = (window.store = configureStore(applicationDependencies))
 
   ReactDOM.render(
     <Provider store={store}>
       <EnvironmentProvider value={environmentContext}>
-        <ThemeProvider>
-          {openFin.isRunningInOpenFin ? (
-            <OpenFinProvider openFin={openFin}>
+        <ThemeState.Provider name="light">
+          <DeprecatedUITheme.ThemeProvider>
```
bhavesh-desai-scratch: Is this really deprecated? I see how you this PR lets you change themes but it does not change the way we consume themes.

src/client/src/rt-theme/ThemeState.tsx
```diff
@@ -0,0 +1,96 @@
+import React from 'react'
+import { ThemeProvider } from './ThemeProvider'
+import { Theme, themes } from './themes'
+
+type ThemeName = 'light' | 'dark'
+
+interface ThemeSelector {
+  name?: ThemeName
```
bhavesh-desai-scratch: Are you sure this should be optional?

src/client/src/rt-theme/ThemeState.tsx
```diff
@@ -0,0 +1,96 @@
...
+  renderThemeStateManager = (context: ThemeStateValue | null) => {
+    return <ThemeStateManager context={context} {...this.props} />
+  }
+}
+
+class ThemeStateManager extends React.Component<ThemeStateProps & { context: ThemeStateValue }, ThemeStateValue> {
+  static getDerivedStateFromProps(props, state) {
+    const { name, context } = props
+
+    if (context == null && name == null) {
```
bhavesh-desai-scratch: This looks like it should be derived at compile time using TS instead of having the guard

src/client/src/rt-theme/ThemeState.tsx
bhavesh-desai-scratch: It should not be optional in the provider. For the consumer it is optional.

src/client/src/rt-theme/ThemeState.tsx
```diff
@@ -0,0 +1,96 @@
...
+  render() {
+    return <ContextConsumer children={this.renderThemeStateManager} />
+  }
+
+  renderThemeStateManager = (context: ThemeStateValue | null) => {
+    return <ThemeStateManager context={context} {...this.props} />
+  }
+}
+
+class ThemeStateManager extends React.Component<ThemeStateProps & { context: ThemeStateValue }, ThemeStateValue> {
```
bhavesh-desai-scratch: We should avoid naming things with words like "manager". Its a little wooly.

src/client/src/rt-theme/ThemeState.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+const { Provider: ContextProvider, Consumer: ContextConsumer } = React.createContext(null)
+
+/**
+ * Set default theme and allow descendants to update selected theme.
+ *
+ * @example
+ *  <ThemeState name={light | dark} onChange={this.handleThemeChange} />
+ */
+class ThemeStateProvider extends React.Component<ThemeStateProps> {
```
bhavesh-desai-scratch: The provider is an immediate consumer of the data it is supposed to be providing 😕

src/client/src/rt-theme/ThemeState.tsx
keshaldesai: I agree with this

src/client/src/rt-theme/ThemeState.tsx
```diff
@@ -0,0 +1,96 @@
...
+    return <ContextConsumer children={this.renderThemeStateManager} />
+  }
+
+  renderThemeStateManager = (context: ThemeStateValue | null) => {
+    return <ThemeStateManager context={context} {...this.props} />
+  }
+}
+
+class ThemeStateManager extends React.Component<ThemeStateProps & { context: ThemeStateValue }, ThemeStateValue> {
+  static getDerivedStateFromProps(props, state) {
```
keshaldesai: Type these

src/client/src/rt-theme/ThemeState.tsx
```diff
@@ -0,0 +1,96 @@
...
+        if (this.props.onChange) {
+          this.props.onChange(this.state.name)
+        }
+      })
+    }
+  }
+
+  render() {
+    const { context, children } = this.props
+    const { name } = this.state
```
keshaldesai: It seems like we're losing the type here

src/client/src/shell/Shell.tsx
```diff
@@ -67,18 +61,11 @@ class Shell extends React.Component<Props, State> {
...
-                <SplashScreen />
-              </SplashScreenContainer>
-            )}
-          </Transition>
-        }
+        {loaded ? null : (
+          <SplashScreenContainer>
+            <SplashScreen />
+          </SplashScreenContainer>
+        )}
```
keshaldesai: `!loaded && <SplashScreenContainer>...`

src/client/src/rt-storybook/Story/Story.tsx
```diff
@@ -1,9 +1,54 @@
...
+
+    return (
+      <ThemeState.Provider name="light">
+        <StyledStory>
+          <Toolbar>
+            <ThemeState.Consumer>
+              {({ name, setTheme }) => (
+                <IconButton
+                  onClick={() => setTheme({ name: name === 'dark' ? 'light' : 'dark' })}
+                  type={name || 'primary'}
```
keshaldesai: Where is this being used?

src/client/src/rt-theme/ThemeState.tsx
evanrs: `null` is a valid value, at which point you'd use the previous or root.

src/client/src/rt-theme/ThemeState.tsx
evanrs: Yup, make your suggestion.

src/client/src/rt-theme/ThemeState.tsx
evanrs: Yes, it is. That's the point.

src/client/src/rt-theme/ThemeState.tsx
evanrs: They are typed — if the class is typed these should be too. Or, is React's type definition / typescript not capable?

src/client/src/rt-theme/ThemeState.tsx
evanrs: The class is typed, this should be typed automatically.

src/client/src/shell/Shell.tsx
evanrs: No. 🙂

src/client/src/rt-storybook/Story/Story.tsx
evanrs: It's copied from `<ThemeSwitcher/>`, and `ThemeSwicher` was removed.

evanrs:
@bhavesh-desai-scratch like we discussed the advantage is theming per tear-off, or component. 

This simply provides the communication channel for that selection. Such that the selector in our header — or a toggle anywhere — can set the theme of the nearest theme-able portion.

evanrs:
More importantly than any of that — it isolates theming functionality to `rt-theme`, eliminating cross imports from our modules.

src/client/src/rt-theme/ThemeState.tsx
bhavesh-desai-scratch: Can you see a use case where you would need multiple providers for this code?

src/client/src/rt-theme/ThemeState.tsx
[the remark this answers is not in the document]
evanrs: It seems like you're referring to a different code point. Otherwise, can you clarify what you're saying?

src/client/src/rt-theme/ThemeState.tsx
evanrs: [addressed in offline chat]

src/client/src/rt-theme/ThemeState.tsx
```diff
@@ -0,0 +1,96 @@
...
+        name: name || context.name
+      }
+    }
+
+    return null
+  }
+
+  state = {
+    name: null,
+    setTheme: ({ name }: ThemeSelector) => {
```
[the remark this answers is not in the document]
evanrs: It's not, really. There just aren't many use-cases for it.

The facility provided is a stable object reference to use in context. Reference checks determine whether our context consumers will update.

src/client/src/rt-theme/ThemeState.tsx
```diff
@@ -0,0 +1,96 @@
...
+      })
+    }
+  }
+
+  render() {
+    const { context, children } = this.props
+    const { name } = this.state
+
+    return (
+      <ContextProvider value={this.state}>
```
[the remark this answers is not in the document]
evanrs: See above 👆— this is the state of the context, and a stable object reference that only reflects state changes.

PR #788, opened 2018-08-15
[ARTP-177] Reskin OpenFin notifications

keshaldesai:
Temporarily using the old theme method. Will update to new theme method in another branch.


src/client/.babelrc
```diff
@@ -1,12 +0,0 @@
-{
```
evanrs: Thanks for cleaning up after me! 😅

src/client/src/ui/spotTile/components/spotTile.stories.tsx
```diff
@@ -124,70 +141,79 @@ const onNotificationDismissedClick = action('Notification dismissed')
...
-        width: '320px',
-        height: '150px'
-      }}
-    >
-      <SpotTile currencyPair={currencyPair} spotTileData={spotTileData} executeTrade={executeTrade} />
-    </div>
+    <Centered>
+      <div
+        style={{
+          width: '320px',
```
evanrs: Assume I add a bunch of comments regarding `rem`s — but no need to change, they're not critical here.

src/client/src/shell/notification/TradeNotification.tsx
```diff
@@ -1,61 +1,56 @@
-import classnames from 'classnames'
+import moment from 'moment'
 import React from 'react'
 import { Direction, Trade, TradeStatus } from 'rt-types'
-import '../../ui/styles/css/index.css'
+import { Bottom, Close, CloseContainer, MetaContainer, MetaTitle, Notification, Status, Top, Traded } from './Styled'
```
evanrs: warning, _nit_, 100% _nit_: 
I think the file pattern we resolve with @bhavesh-desai-scratch would make it lower cased, so `./styled`

evanrs: Because it exports multiple, and not a class as its default.

src/client/src/shell/notification/Styled.tsx
```diff
@@ -0,0 +1,50 @@
+import { Flex } from 'rt-components'
+import { styled } from 'rt-util'
+
+export const Notification = styled('div')`
+  height: 100%;
+  background-color: ${({ theme: { background } }) => background.backgroundSecondary};
```
evanrs: Are we not confident about using the new theme structure?

evanrs: It's been merged and ready.

src/client/src/notificationBootstrapper.tsx
```diff
@@ -1,15 +1,24 @@
...
 
 const dismissNotification = () => window.fin.desktop.Notification.getCurrent().close()
 
-const handleNotificationMessage = (message: Trade) => {
+interface Message {
+  tradeNotification: Trade
+  theme: Themes
+}
+
+const handleNotificationMessage = ({ tradeNotification, theme }: Message) => {
```
evanrs: I'm having trouble seeing where the theme is added. The value would only be `light` or `dark`, right?

src/client/src/shell/notification/TradeNotification.tsx
keshaldesai: You're right!

src/client/src/shell/notification/Styled.tsx
keshaldesai: Most of these commits were before I considered the new structure stable. I will have another task to migrate in.

src/client/src/notificationBootstrapper.tsx
keshaldesai: `shell/openFin/openFin.ts`

Look for openTradeNotification.

PR #789, opened 2018-08-16
convert JS files to TS

bhavesh-desai-scratch:
- At some point, we need to run TS through Babel
- Upgraded to TS 3.0.1
- We should now import styled via rt-theme to get full typing


src/client/src/ui/StatusBar/styled.tsx
```diff
@@ -88,10 +84,4 @@ export const ServiceRoot = styled.div`
   background-color: ${({ index = 0, theme }) =>
     // it'd be nice if this were selected from the original palette …
     darken(index / 50, theme.backgroundColor)};
-
-  ${Icon} {
```
evanrs: Don't remove functionality — provide the solution.

Additionally it is on the docket to get support for component selectors in TS — the better approach here is to make note of that.

src/client/src/ui/StatusBar/StatusBar.tsx
```diff
@@ -1,16 +1,27 @@
...
-import { ThemeProvider } from 'rt-theme'
+import { styled, ThemeProvider } from 'rt-theme'
 
+import { GlobalState } from 'combineReducers'
 import Icon from './Icon'
-import { Root, Body, Fill, ExpandToggle, ServiceList, ServiceRoot, ServiceName, NodeCount } from './styled.js'
+import { Body, ExpandToggle, Fill, NodeCount, Root, ServiceList, ServiceName, ServiceRoot } from './styled'
 
 export const SERVICES = ['blotter', 'reference', 'execution', 'pricing', 'analytics'].map(serviceType => ({
-  serviceType
```
evanrs: You've broken functionality — remove this change.

Or, revisit the use to understand the change you'd like to make.

The correct value for these keys is `null`.

bhavesh-desai-scratch: What does it break?  They had a value of undefined before this change.

src/client/src/ui/StatusBar/styled.tsx
bhavesh-desai-scratch: I have not removed this functionality. We will get TS running through babel at a later point and we can go back to doing things this way. I have added the margins back in the status bar file.

src/client/src/ui/StatusBar/StatusBar.tsx
bhavesh-desai-scratch: isConnected = null | true | false 

makes no sense

PR #797, opened 2018-08-21
Chore/cleanup module structure

bhavesh-desai-scratch:
Stops us importing like this:

`import x from 'somewhere/deep/inside'`

Move connection status and service status into the shell


src/client/src/StoreTypes.ts
```diff
@@ -1,6 +1,7 @@
 import { Action } from 'redux'
 import { Epic } from 'redux-observable'
 import { ApplicationDependencies } from './applicationServices'
-import { GlobalState } from './combineReducers'
+import rootReducer from './combineReducers'
 
+export type GlobalState = ReturnType<typeof rootReducer>
```
keshaldesai: This feels weird to just to get the type.

PR #813, opened 2018-08-29
[ARTP 216] Fix for broken storybook views

keshaldesai:
Bug: Price button, price movement, and notional input broken in storybook.

Cause: In the migration to the new theming method, spot tile styling was not selected for.

Fix: Create a `Story` component with a `ThemeProvider` that selects for spot tile styles from the parent theme.


src/client/src/ui/spotTile/components/notifications/NotificationContainer.tsx
```diff
@@ -34,7 +34,7 @@ export default class NotificationContainer extends PureComponent<Props> {
       )
     }
     if (!lastTradeExecutionStatus || !hasNotification(lastTradeExecutionStatus)) {
-      return () => null as SFC<void>
+      return (() => null) as SFC<object>
```
bhavesh-desai-scratch: I thought about this. We should type the functions return type instead of doing this.

PR #815, opened 2018-08-29
[ARTP-218] Blotter pop out flashing content before styles loaded

keshaldesai:
Bug: When popping out the blotter, the blotter grid's header titles were visible before the styles were fully loaded.

Cause: Styles not fully loaded before content is visible.

Fix: Delay adding the content to the pop out DOM to allow styles to load.

### DIFF
--- a/src/client/src/rt-components/tear-off/Portal.tsx
+++ b/src/client/src/rt-components/tear-off/Portal.tsx
@@ -71,9 +71,7 @@ class NewPortal extends React.Component<PortalProps & { environment: Environment
       childHead.innerHTML = parentHead.innerHTML.replace(/\/static/g, window.location.href + 'static')
       this.externalWindow.document.title = title
 
-      setTimeout(() => {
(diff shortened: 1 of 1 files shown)


src/client/src/rt-components/tear-off/Portal.tsx
```diff
@@ -63,14 +63,13 @@ class NewPortal extends React.Component<PortalProps & { environment: Environment
...
-      this.externalWindow.document.title = title
-
       const parentHead = document.head
       const childHead = this.externalWindow.document.head
 
       childHead.innerHTML = parentHead.innerHTML.replace(/\/static/g, window.location.href + 'static')
+      this.externalWindow.document.title = title
 
-      this.externalWindow.document.body.appendChild(this.container)
+      setTimeout(() => this.externalWindow.document.body.appendChild(this.container), 200)
```
bhavesh-desai-scratch: Add a comment here so people know why you have done this

PR #825, opened 2018-09-05
Add overlay for Analytics

bhavesh-desai-scratch:


src/client/public/index.html
```diff
@@ -29,6 +29,11 @@
   <noscript>
     You need to enable JavaScript to run this app.
   </noscript>
+  <style>
+    #root {
+      height: 100vh;
+    }
+  </style>
```
evanrs: Style tag in `index.html` needs justifying comment — I've removed it for now.

PR #833, opened 2018-09-07
fix chrome styles

bhavesh-desai-scratch:


src/client/src/rt-components/open-fin/Openfin.stories.tsx
```diff
@@ -0,0 +1,26 @@
...
+import { styled } from 'rt-theme'
+import { OpenFinChrome, OpenFinHeader } from './OpenFinChrome'
+
+const stories = storiesOf('Openfin', module)
+
+stories.add('Openfin chrome', () => (
+  <Story>
+    <Container>
+      <OpenFinChrome>
+        <OpenFinHeader close={() => {}} minimize={() => {}} maximize={() => {}} />
```
keshaldesai: Use https://github.com/storybooks/storybook/tree/master/addons/actions

PR #836, opened 2018-09-10
Move roots into route components

evanrs:


src/client/src/MainRoute.tsx
```diff
@@ -0,0 +1,90 @@
+import React from 'react'
```
keshaldesai: Please git move these files so we don't lose history

src/client/src/index.tsx
```diff
@@ -1,12 +1,18 @@
...
 
-if (parsed.notification) {
-  run()
-} else {
-  import('./main').then(mod => mod.run())
-}
+ReactDOM.render(
+  <BrowserRouter>
+    <Switch>
+      <Route path="/notification" component={NotificationRoute} />
```
keshaldesai: 👍

src/client/src/applicationServices.ts
```diff
@@ -39,7 +45,7 @@ export function createApplicationServices(user: User, autobahn: AutobahnConnecti
   return {
     referenceDataService,
 
-    openFin,
+    openFin: openfin,
```
keshaldesai: Keep `openFin` consistent

src/client/src/index.tsx
```diff
@@ -1,12 +1,18 @@
...
+      <Route component={MainRoute} />
```
bhavesh-desai-scratch: Nice

src/client/src/MainRoute.tsx
```diff
@@ -0,0 +1,90 @@
...
+      this.store.dispatch(ConnectionActions.disconnect())
+      log.warn(`Application has reached disconnection time at ${APPLICATION_DISCONNECT}`)
+    })
+  }
+
+  render() {
+    return (
+      <React.Fragment>
+        {/* The below style tags are required to preload bold and bold-italic fonts */}
+        <span style={{ fontWeight: 900 }} />
```
bhavesh-desai-scratch: Could we not put this in our index.html template?

I know this was here before btw :)

src/client/src/MainRoute.tsx
bhavesh-desai-scratch: Man. I'm so bad at doing this.

src/client/src/MainRoute.tsx
evanrs: Oh fair point!

src/client/src/index.tsx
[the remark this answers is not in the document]
evanrs: Imports are now without side effect.

We can add `react-loadable` if we can show value for async splitting — I could not find significant value when examined. The bundle will be a cache read for electron/chrome.

I could see value once styleguide is introduced as it may entail a number of additional content assets — though the effect will likely be under 25kb, which remains insignificant.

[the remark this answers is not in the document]
bhavesh-desai-scratch: Just checked. It is cached

src/client/src/MainRoute.tsx
evanrs: Usually git manages this pretty well — in this case it appears the resulting file is so different git didn't not regard it as the same.

I think we're past the point of amending this unless I soft revert and squash into a single commit.

src/client/src/index.tsx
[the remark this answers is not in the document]
evanrs: Nice, thanks for verifying!

src/client/package.json
```diff
@@ -91,7 +90,7 @@
     "reselect": "3.0.1",
     "ress": "^1.2.2",
     "rxjs": "6.1.0",
-    "rxjs-spy": "^7.0.1",
```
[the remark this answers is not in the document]
evanrs: 👍

PR #839, opened 2018-09-10
ARTP-241 Analytics Double Minus

ianchan035:
Fixed the double minus at Analytics

### DIFF
--- a/src/client/src/ui/analytics/components/Analytics.tsx
+++ b/src/client/src/ui/analytics/components/Analytics.tsx
@@ -82,14 +82,13 @@ export default class Analytics extends React.Component<Props> {
 
 function lastPositionWithDirection(lastPos: number) {
   let formattedLastPos = numeral(lastPos).format()
-  const color = ''
+  let color = ''
   if (lastPos > 0) {
     color = 'green'
     formattedLastPos = '+' + formattedLastPos
   }
   if (lastPos < 0) {
     color = 'red'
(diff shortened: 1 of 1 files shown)


src/client/src/ui/analytics/components/Analytics.tsx
```diff
@@ -103,7 +103,7 @@ function lastPositionWithDirection(lastPos: number) {
   }
   if (lastPos < 0) {
     color = 'red'
-    formattedLastPos = '-' + formattedLastPos
+    formattedLastPos = formattedLastPos
```
keshaldesai: Nice fix, I think you can just remove this line.

ianchan035: Done !

src/client/src/ui/analytics/components/Analytics.tsx
```diff
@@ -96,14 +82,14 @@ export default class Analytics extends React.Component<Props> {
 
 function lastPositionWithDirection(lastPos: number) {
   let formattedLastPos = numeral(lastPos).format()
-  let color = ''
+  const color = ''
```
keshaldesai: Only use const if the value is constant. Because we change the value later, this must be let.

ianchan035: Oh I didn't mean to commit that

PR #840, opened 2018-09-10
Fix openfin chrome

bhavesh-desai-scratch:


ianchan035:
openfinchrome tearoff

PR #847, opened 2018-09-11
ARTP-255 Fix /notification path

evanrs:
Adds SPA path rule to web nginx.conf to fix `/notification` path 404.


evanrs:
To restore web-dev comments will be addressed post-merge.

PR #848, opened 2018-09-11
Workspace fixes

keshaldesai:
This PR fixes:

- Tile shifting
- Font shift in tearoffs (font preload again)
- Analytics labels changing colors when resizing

### DIFF
--- a/src/client/src/rt-components/tear-off/TearOff.tsx
+++ b/src/client/src/rt-components/tear-off/TearOff.tsx
@@ -37,7 +37,10 @@ export default class TearOff extends React.PureComponent<Props, State> {
     if (tornOff) {
       return (
         <Portal onUnload={this.popIn} {...portalProps}>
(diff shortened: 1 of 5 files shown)


src/client/src/rt-components/tear-off/TearOff.tsx
```diff
@@ -37,7 +37,10 @@ export default class TearOff extends React.PureComponent<Props, State> {
     if (tornOff) {
       return (
         <Portal onUnload={this.popIn} {...portalProps}>
-          <TearOffContainer>{render(this.popOut, tornOff)}</TearOffContainer>
+          <TearOffContainer>
+            <span style={{ fontWeight: 900, fontStyle: 'italic' }} />
```
evanrs: It's curious that this isn't handled here:
https://github.com/AdaptiveConsulting/ReactiveTraderCloud/blob/master/src/client/public/index.html#L47

PR #852, opened 2018-09-12
Chore/performance

keshaldesai:
This patch centers around performance improvements:

- Block Resizer from calling `setState` on every `mouseup`
- Remove logger as `...Array.from()` on each log had a high performance cost


src/client/src/MainRoute.tsx
```diff
@@ -50,7 +50,7 @@ export default class MainRoute extends React.Component {
 
     timer(APPLICATION_DISCONNECT).subscribe(() => {
       this.store.dispatch(ConnectionActions.disconnect())
-      log.warn(`Application has reached disconnection time at ${APPLICATION_DISCONNECT}`)
+      console.warn(LOG_NAME, `Application has reached disconnection time at ${APPLICATION_DISCONNECT}`)
```
bhavesh-desai-scratch: Use the logger!

bhavesh-desai-scratch: Not sure why you have chosen not to use the logger.

src/client/src/rt-components/resizer/Resizer.tsx
```diff
@@ -78,7 +78,12 @@ export default class Resizer extends Component<Props, State> {
     document.removeEventListener('touchend', this.handleStop)
   }
 
-  handleStop = () => this.setState({ dragging: false })
+  handleStop = () => {
```
bhavesh-desai-scratch:
```
 handleStop = () => {
    if (this.state.dragging) {
         this.setState({ dragging: false })
    }
  }
```

bhavesh-desai-scratch: No need for an early return here

src/client/src/rt-system/ServiceStub.ts
```diff
@@ -4,9 +4,8 @@ import { filter, mergeMap, switchMap } from 'rxjs/operators'
 
 import './AutoBahnTypeExtensions'
 import { ConnectionEvent, ConnectionEventType, ConnectionOpenEvent } from './connectionStream'
-import logger from './logger'
 
-const log = logger.create('Connection')
+const LOG_NAME = 'Connection:'
```
bhavesh-desai-scratch: Why are you removing the logger?

PR #854, opened 2018-09-13
ARTP-250 Add styleguide

evanrs:
### Desktop
styleguide 1040

### Mobile
styleguide iphone x


src/client/.prettierrc
```diff
@@ -1,5 +1,6 @@
 {
   "semi": false,
   "singleQuote": true,
-  "printWidth": 120
+  "printWidth": 120,
+  "trailingComma": "all"
```
keshaldesai: Changes like this should really be in their own PR if possible.

src/client/package.json
```diff
@@ -54,6 +54,12 @@
       "git add"
     ]
   },
+  "prettier": {
+    "semi": false,
+    "singleQuote": true,
+    "printWidth": 120,
+    "trailingComma": "all"
+  },
```
keshaldesai: Why does the prettier config also ned to be here?

src/client/src/routes/StyleguideRoute/StyleguideRoute.tsx
```diff
@@ -0,0 +1,73 @@
...
+    ColorSpectrum,
+    CoreBranding,
+    FontFamilies,
+    Atoms,
+    Ending: () => <SectionBlock mh={5} intent="inverted" />,
+  },
+  (value, key) => _.kebabCase(key),
+)
+
+export class StyleguideRoute extends React.Component {
```
keshaldesai: If you wanted, this looks like it could become a functional component.

src/client/src/routes/StyleguideRoute/StyleguideRoute.tsx
```diff
@@ -0,0 +1,73 @@
...
+import FloatingTools from './components/FloatingsTools'
+import { Block, SectionBlock } from './styled'
+
+import Atoms from './sections/Atoms'
+import ColorSpectrum from './sections/ColorSpectrum'
+import CoreBranding from './sections/CoreBranding'
+import FontFamilies from './sections/FontFamilies'
+import Introduction from './sections/Introduction'
+
+const sections = _.mapKeys(
```
keshaldesai: There is a lot of dependency on lodash in this file. Are there any areas where it might be excessive?

src/client/src/routes/StyleguideRoute/components/ButtonGrid.tsx
```diff
@@ -0,0 +1,157 @@
...
+import _ from 'lodash'
+import React from 'react'
+
+import { Button, ButtonGroup } from 'rt-styleguide'
+import { resolvesColor, styled } from 'rt-theme'
+
+import { PassThroughProps } from '../tools'
+
+export default class ButtonGrid extends React.Component {
+  state = {}
```
keshaldesai: Remove

src/client/src/routes/StyleguideRoute/components/ButtonGrid.tsx
```diff
@@ -0,0 +1,157 @@
...
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
+import _ from 'lodash'
+import React from 'react'
+
+import { Button, ButtonGroup } from 'rt-styleguide'
+import { resolvesColor, styled } from 'rt-theme'
+
+import { PassThroughProps } from '../tools'
+
+export default class ButtonGrid extends React.Component {
```
keshaldesai: Can be a functional component.

src/client/src/routes/StyleguideRoute/components/ColorBlocks.tsx
```diff
@@ -0,0 +1,151 @@
...
+import { PassThroughProps } from '../tools'
+
+const { spectrum } = colors
+
+const SWATCH_KEYS = {
+  light: [95, 9, 8, 7, 6, 5, 4, 3, 2, 1].map(v => `L${v}`),
+  dark: [1, 2, 3, 4, 5, 6, 7, 8, 9, 10].map(v => `D${v}`),
+}
+
+export interface ColorBlocksProps extends PassThroughProps {}
```
keshaldesai: If there is nothing added to the extended interface, why extend?

src/client/src/routes/StyleguideRoute/components/ColorBlocks.tsx
```diff
@@ -0,0 +1,151 @@
...
+
+export const ColorBlocks: React.SFC<ColorBlocksProps> = props => {
```
keshaldesai: Can be functional.

src/client/src/routes/StyleguideRoute/elements.tsx
```diff
@@ -0,0 +1,97 @@
...
+import { resolvesColor } from 'rt-theme'
+
+import { mapMarginPaddingProps, mapTextProps, MarginPaddingProps, TextProps } from './styled'
+
+export { Paragraph } from './styled'
+
+export const Hashable = ({ is: Element = 'div', ...props }) => <Element {...props} id={_.kebabCase(props.children)} />
+
+interface HeaderProps extends MarginPaddingProps, TextProps {}
+export const Header: Styled<HeaderProps> = styled(Hashable)`
```
keshaldesai: General comment for all the places you have this:

`Component: Styled<Props>` --> `Component = styled<Props>('div')`

You shouldn't need the `Styled` interface.

src/client/src/routes/StyleguideRoute/mapProps.tsx
```diff
@@ -0,0 +1,115 @@
+/**
+ * Copied from @evanrs/map-props
+ * https://github.com/evanrs/map-props
+ */
```
keshaldesai: I don't think this is appropriate.

src/client/src/routes/StyleguideRoute/styled/Paragraph.tsx
```diff
@@ -0,0 +1,17 @@
+import { styled, Styled } from 'rt-theme'
+
+import { mapMarginPaddingProps, MarginPaddingProps } from './mapMarginPaddingProps'
+import { mapTextProps, TextProps } from './Text'
+
+export interface ParagraphProps extends TextProps, MarginPaddingProps {}
```
keshaldesai: `type ParagraphProps = TextProps & MarginPaddingProps`

src/client/src/routes/StyleguideRoute/styled/SectionBlock.tsx
```diff
@@ -0,0 +1,125 @@
+import _ from 'lodash'
+import React from 'react'
+import { css, styled, Styled } from 'rt-theme'
+
+import { Block, BlockProps } from '../styled'
+import { mapMarginPaddingProps, MarginPaddingProps } from './mapMarginPaddingProps'
+
+export interface SectionProps extends BlockProps, MarginPaddingProps {
+  mh?: any
```
keshaldesai: Is this really an `any`?

src/client/src/routes/StyleguideRoute/styled/mapMarginPaddingProps.ts
```diff
@@ -0,0 +1,95 @@
+import _ from 'lodash'
+import { css } from 'rt-theme'
+
+import { mapProps } from '../tools'
+
+// export type PropValue = number | string | boolean
+export type PropValue = any
```
keshaldesai: You should be able to type this.

src/client/src/routes/StyleguideRoute/styled/mapMarginPaddingProps.ts
```diff
@@ -0,0 +1,95 @@
...
+  py?: PropValue
+  pt?: PropValue
+  pr?: PropValue
+  pb?: PropValue
+  pl?: PropValue
+}
+
+export interface MarginPaddingProps extends MarginProps, PaddingProps {}
+
+export const { marginPaddingProps, mapMarginPaddingProps } = (() => {
```
keshaldesai: Some comments might help the user understand what you're trying to accomplish.

src/client/src/routes/StyleguideRoute/styled/rules.ts
```diff
@@ -0,0 +1,44 @@
+import { css } from 'rt-theme'
+
+export { mapMarginPaddingProps, MarginPaddingProps } from './mapMarginPaddingProps'
+
+export const preventStutter = css`
+  -webkit-backface-visibility: hidden;
```
keshaldesai: I think that even though these properties are unsupported or experimental, the preprocessor will auto-prefix.

src/client/src/rt-styleguide/components/Button.tsx
```diff
@@ -111,7 +111,7 @@ export class ButtonGroup extends React.Component<ButtonStyleProps> {
   }
 }
 
-const BaseElement = styled.button<ButtonStyleProps>`
+const StyledBase = withProps({ role: 'button' })(styled.div`
```
keshaldesai: Are we adding `recompose` back just for this one case?

src/client/src/rt-theme/globals.ts
```diff
@@ -53,24 +53,21 @@ export default injectGlobal`
...
-  ::-webkit-scrollbar-corner {
-    background-color: rgba(0,0,0,0);
+    @media all and (max-width: 400px) {
+      font-size: 14px;
+    }
+
+    @media all and (max-width: 320px) {
+      font-size: 12px;
+    }
+  }
```
keshaldesai: 👍

src/client/src/rt-theme/index.ts
```diff
@@ -2,9 +2,9 @@ export { flush, hydrate, cx, merge, getRegisteredStyles, injectGlobal, keyframes
...
 
-export { colors } from './colors'
+export { colors, CorePalette, CorePaletteMap, ColorPalette, ColorPaletteMaps } from './colors'
 export { themes, Theme } from './themes'
-export { resolvesColor } from './tools'
+export { resolvesColor, Selector } from './tools'
 export { ThemeProvider } from './ThemeProvider'
 export { ThemeState, ThemeStateValue } from './ThemeState'
-export { default as styled } from './styled'
+export { default as styled, Styled, StyledProps } from './styled'
```
keshaldesai: You shouldn't need `Styled` or `StyledProps`.

src/client/src/shell/routes/ShellRoute.tsx
```diff
@@ -78,7 +78,7 @@ const WorkspaceWrapper = styled(Wrapper)`
 `
 
 const AnalyticsWrapper = styled(Wrapper)`
-  padding: 0.375rem 1.25rem 0 0;
+  padding: 0.5rem 0.5rem;
```
keshaldesai: `padding: 0.5rem`

src/client/src/ui/analytics/components/pnlChart/chartGradient.ts
```diff
@@ -11,9 +11,9 @@ import 'classlist-polyfill'
  * read about the SVG defs:
  * https://developer.mozilla.org/en-US/docs/Web/SVG/Element/defs
  *
- * There are four colour stops: top, bottom, and the two in the middle that have the same offset value
+ * There are four color stops: top, bottom, and the two in the middle that have the same offset value
```
keshaldesai: 😉

src/client/.prettierrc
evanrs: I agree — definitely an oversight.

src/client/package.json
evanrs: Paranoia involving VSCode — I'll remove.

src/client/src/routes/StyleguideRoute/StyleguideRoute.tsx
evanrs: There is no meaningful cost due to the use of `babel-lodash-plugin` — and no, I don't think use of a functional tool like this is excessive.

Additionally, the tooling around `lodash` is some of the best — therefore it shouldn't impose any costs on the developer. And, while the API should be obvious simply by convention it also comes with complete documentation in VSCode or Atom with an extension.

There is every reason to use it where it provides a relevant functional primitive that simplifies the expression and adds no cognitive or runtime overhead.

src/client/src/routes/StyleguideRoute/components/ButtonGrid.tsx
evanrs: 👍

src/client/src/rt-styleguide/components/Button.tsx
evanrs: I think the right framing is with consideration that `Emotion` is an  incomplete implementation of `styled-components`. While `styled-components` implements the `attrs` method to address this very common use-case, the author of `Emotion` refuses to — instead they prescribe this approach. As the officially documented approach to an existing and expected API I think we should resist introducing additional indirection or variation on a variation.

I will look to see if I can remove it simply by adding it to the wrapping components — as this component is not exported. However, where similar cases emerge I believe this framing is the correct approach to such an issue.

src/client/src/rt-theme/index.ts
evanrs: There are two points to address:
1. Some APIs entirely fail when employing the `styled.div<props>` approach. I would encourage anyone finding a solution.
2. This approach is consistent with the typing of any other non-class component
```ts
const SomeComponent: React.SFC<SomeComponentProps> = () => 
```

src/client/src/shell/routes/ShellRoute.tsx
evanrs: Fair.

src/client/src/routes/StyleguideRoute/StyleguideRoute.tsx
evanrs: 👍

src/client/src/routes/StyleguideRoute/components/ColorBlocks.tsx
evanrs: Do you mean it can be implicit return?

src/client/src/routes/StyleguideRoute/elements.tsx
evanrs: It continues to be a useful tool — at times typescript objects to that method. Additionally the typescript + styled tooling is incomplete and features are lost by that approach. Correction and formatting on CSS is lost, for example. And, as mentioned before, it is actually less consistent without how most types are described.

src/client/src/routes/StyleguideRoute/mapProps.tsx
evanrs: How so? It would be appropriate without the comment?

src/client/src/routes/StyleguideRoute/elements.tsx
evanrs: How would you achieve typing on the `is` prop — a convention from `styled–system`:
https://github.com/AdaptiveConsulting/ReactiveTraderCloud/pull/854#diff-612387fc2db9d8ac0322d7125d4ea86cR137

src/client/src/routes/StyleguideRoute/mapProps.tsx
evanrs: This specific file only serves `mapMarginPaddingProps` a useful tool for design heavy marketing content — this mimics `styled-system` convention of `px …`, `mx …` for adding white space. Unfortunately Styled System does not support Emotion — otherwise I'd readily advise adopting it here.

And, because the conventions like this rest on a great deal of existing work and experience it will necessarily not be a wholly unique implementation. However, as this code now resides within the code base it poses issue of governance or maintenance that might present a risk to Reactive Trader in the long term.

evanrs: However, `mapMarginPaddingProps` can be rewritten to isolate the implementation there.

src/client/src/routes/StyleguideRoute/styled/Paragraph.tsx
evanrs: But, is there any reason to not describe it as `interface`? For example, can a `type` be extended by another `interface`?

src/client/src/routes/StyleguideRoute/mapProps.tsx
keshaldesai: It is my opinion that we should not be tagging our work on RT. If you're comfortable with leaving it here without your name, I think that's fine. However, if you're not, I would suggest rewriting or finding a replacement. Again, this is my personal opinion and I don't mind if you find someone to overrule this.

src/client/src/routes/StyleguideRoute/styled/mapMarginPaddingProps.ts
evanrs: I will add a comment at the top to this documentation — in general anything that gets closer to `styled-system` for this kind of primitive is ideal.

https://jxnblk.com/styled-system/getting-started#margin--padding

src/client/src/routes/StyleguideRoute/elements.tsx
keshaldesai: Leave a comment on the exports of where you create the `Styled` type to explain why you're using them.

src/client/src/rt-theme/index.ts
keshaldesai: 1. That makes sense.
2. TypeScript is able to infer from the styled function. In the example you provide, TypeScript cannot infer that this function is specifically `React.SFC`, hence the explicit type.

PR #855, opened 2018-09-13
[ARTP-257] Tearoff routing

keshaldesai:


src/client/src/rt-theme/emotion.ts
```diff
@@ -13,10 +13,6 @@ export const {
...
 
-/**
- * Until Emotion 10, we need to block Emotion from using speedy mode.
- * This allows us to use a MutationObserver in our portals to watch
- * for changes in style tags.
- */
+//Enable speedy mode for improved dev performance
 
-sheet.speedy(false)
+sheet.speedy(true)
```
evanrs: 👍

src/client/src/rt-theme/globals.ts
```diff
@@ -54,10 +54,12 @@ export default injectGlobal`
...
     text-rendering: geometricPrecision;
+    height: 100%;
+    overflow: auto;
   }
-  
-  html {
-    overflow-y: auto;
+
+  #root {
+    height: 100%;
```
evanrs: When resolving the merge conflict with #854 I believe this will perform better: https://github.com/AdaptiveConsulting/ReactiveTraderCloud/blob/ARTP-250/add-styleguide/src/client/src/rt-theme/globals.ts#L68-L72

Which areas did this resolve problem? Or, could you test and verify that your problems are resolved by the version of `globals` on #854?

src/client/src/ui/spotTile/components/TileSwitch.tsx
```diff
@@ -22,7 +22,7 @@ const TileSwitch: React.SFC<Props> = ({
   currencyPair,
   spotTileData,
   executeTrade,
-  tornOff,
+  canPopout,
```
evanrs: I think `canPopout` is much better than `tornOff` — if you wrote RT fresh would you prefer `popout` over `tearOff`, `tearable` generally?

src/client/src/rt-theme/globals.ts
keshaldesai: The goal was to allow pop out components to use full width/height of their window. Please overwrite if you're able to get the same functionality with your approach.

PR #857, opened 2018-09-17
[ARTP-267] Theme changing across windows

keshaldesai:
Allows theme changes to propagate to any child or sibling windows/tabs.


src/client/src/rt-theme/ThemeState.tsx
```diff
@@ -42,50 +42,12 @@ class ThemeStateProvider extends React.Component<ThemeStateProps> {
     return <ContextConsumer children={this.renderThemeStateManager} />
   }
 
-  renderThemeStateManager = (context: ThemeStateValue) => {
```
evanrs: We lose the theme boundaries — but seeing how we weren't using them anyways, this works.

PR #858, opened 2018-09-17
Added a route wrapper which provides openfin headers in openfin

keshaldesai:
OpenFinHeader for new window routes.

### DIFF
--- a/src/client/src/rt-components/route-wrapper/RouteWrapper.tsx
+++ b/src/client/src/rt-components/route-wrapper/RouteWrapper.tsx
@@ -7,6 +7,10 @@ const RouteStyle = styled.div`
   width: 100%;
 `
 
+const closeOpenFinWindow = () => {
+  fin.desktop.Window.getCurrent().close()
+}
+
 interface Props {
   environment: EnvironmentValue
 }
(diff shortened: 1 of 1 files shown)


src/client/src/rt-components/route-wrapper/RouteWrapper.tsx
```diff
@@ -0,0 +1,29 @@
+import React from 'react'
+import { EnvironmentValue, OpenFinChrome, OpenFinHeader, withEnvironment } from 'rt-components'
+import { styled } from 'rt-theme'
+
+const RouteStyle = styled.div`
```
evanrs: This shouldn't be necessary.

evanrs: I see how you're using it as a selector, works 👍

src/client/src/rt-components/route-wrapper/RouteWrapper.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+interface Props {
+  environment: EnvironmentValue
+}
+
+const RouteWrapperBase: React.SFC<Props> = ({ environment, children }) => (
+  <RouteStyle>
+    {environment.isDesktop ? (
+      <OpenFinChrome>
+        <OpenFinHeader close={() => fin.desktop.Window.getCurrent().close()} />
```
bhavesh-desai-scratch: Should this be an inline function?

PR #859, opened 2018-09-17
Chore/update react and performance

keshaldesai:
- Update `React` and `ReactDOM` to `16.5.1`
- Convert animations to use memoized functions to prevent hundreds of function calls
- Use `PureComponent` to block extra renders


src/client/src/rt-components/adaptive-loader/AdaptiveLoader.tsx
```diff
@@ -22,14 +23,17 @@ const getBounce = (moveDistance: number) => keyframes`
   }
 `
 
+const memoizedGetBounce = _.memoize(getBounce)
```
evanrs: Nit: you can just wrap the original and stick with `getBounce` — we expect ideal behavior from our functions, if the memoization is correct than we have no need to spread that knowledge.

PR #864, opened 2018-09-19
ARTP-265 Debugger

ianchan035:
logs
verbose
logger2

- Using the logdown package https://www.npmjs.com/package/logdown,
- Supports levels, colours, markdown
- Browser Command, localStorage.debug = "app:*"

### DIFF
--- a/src/client/src/logger.ts
+++ b/src/client/src/logger.ts
@@ -1,0 +1,48 @@
+import logdown from 'logdown'
+
+enum DebugLevel {
+  error = 'error',
+  warn = 'warn',
+  debug = 'debug',
+  info = 'info',
+}
+
+export enum DebugType {
+  Error = 'Error',
+  Warning = 'Warning',
+  Debug = 'Debug',
+  Subscribing = 'Subscribing',
+  Subscribed = 'Subscribed',
+  Received = 'Received',
+  Executing = 'Executing',
+  Execute = 'Execute',
+  Info = 'Info',
+}
+
+const levelColour = {
+  error: 'Crimson',
+  warn: 'Tomato',
+  debug: 'Orange',
+  info: 'RoyalBlue',
+}
+
+export const createLogger = (debugLevel: DebugLevel) => {
+  const colour = Object.keys(levelColour)
+    .filter(key => key === debugLevel)
+    .map(level => levelColour[level])
+    .join()
+
+  return (name: string, debugType?: DebugType) =>
+    debugType !== undefined
+      ? (...args: any[]) => logdown(`app:${name}${debugType}`, { prefixColor: colour })[debugLevel](...args)
+      : (...args: any[]) => logdown(`app:${name}`, { prefixColor: colour })[debugLevel](...args)
+}
+
+const logger = {
+  info: createLogger(DebugLevel.info),
+  debug: createLogger(DebugLevel.debug),
+  warn: createLogger(DebugLevel.warn),
+  error: createLogger(DebugLevel.error),
+}
+
+export default logger
--- a/src/client/src/logger.ts
+++ b/src/client/src/logger.ts
@@ -27,10 +27,7 @@ const levelColour = {
 }
 
 export const createLogger = (debugLevel: DebugLevel) => {
-  const colour = Object.keys(levelColour)
-    .filter(key => key === debugLevel)
-    .map(level => levelColour[level])
-    .join()
+  const colour = levelColour[debugLevel]
 
   return (name: string, debugType?: DebugType) =>
     debugType !== undefined
--- a/src/client/src/routes/MainRoute.tsx
+++ b/src/client/src/routes/MainRoute.tsx
@@ -1,4 +1,4 @@
-import logdown from 'logdown'
+import logger, { DebugType } from 'logger'
 import React from 'react'
 import { Provider as ReduxProvider } from 'react-redux'
 import { timer } from 'rxjs'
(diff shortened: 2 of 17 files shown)


evanrs:
@ianchan035 to get automatic tracking in JIRA we have to use their ids. 

I've gone ahead and updated your pull request from `ARTP265-…` to `ARTP-265 …`. The same is true for branch names, so a branch name like `ARTP-265/something-something` will get picked up, but `ARTP265-…` won't. 

Hope that helps 🙂

src/client/src/routes/MainRoute.tsx
```diff
@@ -20,6 +21,7 @@ const APPLICATION_DISCONNECT = 15 * 60 * 1000
 
 const config = getEnvVars(process.env.REACT_APP_ENV!)
 const LOG_NAME = 'Application Service: '
```
evanrs: Because we no longer reference this "global" constant you can go ahead and roll this string into your logger.

src/client/src/routes/MainRoute.tsx
```diff
@@ -20,6 +21,7 @@ const APPLICATION_DISCONNECT = 15 * 60 * 1000
...
+const warnLogger = logdown(`app:${LOG_NAME}`, { prefixColor: 'Tomato' })
```
evanrs: And, since we have no other logger to speak of, we have no need for the name prefix `warn`, so:
```js
const logger = logdown(`app:services`, { prefixColor: 'Tomato' })
```

src/client/src/rt-components/tear-off/DesktopWindow.ts
```diff
@@ -1,7 +1,11 @@
+import logdown from 'logdown'
 import { WindowConfig } from './types'
 
 type DesktopWindowProps = WindowConfig
 
+const LOG_NAME = 'DesktopWindow:'
```
evanrs: Similarly this is not a useful convention here as we won't be referencing a module global in any of our methods.

src/client/src/rt-components/tear-off/DesktopWindow.ts
```diff
@@ -33,7 +37,7 @@ export const openDesktopWindow = (config: DesktopWindowProps) => {
         resolve(win)
       },
       error => {
-        console.log('Error creating window:', error)
+        errorLogger.log('Error creating window:', error)
```
evanrs: It's not significant to indicate `error` — and in general a logger represents a service before a particular state of that service. So, just `logger` is fine.

src/client/src/rt-system/ServiceStub.ts
```diff
@@ -1,11 +1,16 @@
...
 import { filter, mergeMap, switchMap } from 'rxjs/operators'
 
 import './AutoBahnTypeExtensions'
 import { ConnectionEvent, ConnectionEventType, ConnectionOpenEvent } from './connectionStream'
 
 const LOG_NAME = 'Connection:'
+const logger = logdown(`app:${LOG_NAME} Info `, { prefixColor: 'Olive' })
+const subLogger = logdown(`app:${LOG_NAME} Subscribing `, { prefixColor: 'MediumBlue' })
+const debugLogger = logdown(`app:${LOG_NAME}`, { prefixColor: 'ForestGreen' })
+const errorLogger = logdown(`app:${LOG_NAME} Error `, { prefixColor: 'DarkRed' })
```
evanrs: Does log down provide no other mechanism to achieve the color variation you're attempting to achieve?

src/client/src/logger.ts
```diff
@@ -0,0 +1,48 @@
...
+  warn: 'Tomato',
+  debug: 'Orange',
+  info: 'RoyalBlue',
+}
+
+export const createLogger = (debugLevel: DebugLevel) => {
+  const colour = Object.keys(levelColour)
+    .filter(key => key === debugLevel)
+    .map(level => levelColour[level])
+    .join()
```
keshaldesai: `const color = levelColor[debugLevel]`

src/client/src/logger.ts
```diff
@@ -0,0 +1,48 @@
...
+
+  return (name: string, debugType?: DebugType) =>
+    debugType !== undefined
+      ? (...args: any[]) => logdown(`app:${name}${debugType}`, { prefixColor: colour })[debugLevel](...args)
+      : (...args: any[]) => logdown(`app:${name}`, { prefixColor: colour })[debugLevel](...args)
```
keshaldesai: Use `arguments`.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/arguments

Example:
```() => logdown(`app:${name}${debugType}`, { prefixColor: colour })[debugLevel].apply(null, arguments)```


Try something like that(let me know if it works or not).

src/client/src/logger.ts
ianchan035: lol

ianchan035: my bad

ianchan035: Alright my bad here 👍

src/client/src/logger.ts
ianchan035: https://stackoverflow.com/questions/41731854/why-do-arrow-functions-not-have-the-arguments-array

ianchan035: this does not work really well with arrow function, but it is a very good thing to know about 👍

keshaldesai: Ah you're right, good catch!

PR #866, opened 2018-09-19
Generalize environment handling

evanrs:
This is a proposal to generalize the environment we share across applications within Reactive Trader. As Reactive Trader grows to include application environments beyond OpenFin or the browser we'll want consistency in how we think and communicate with that underlying environment.

This change consolidates the definition of the environment and window provider. It alters the interface to be invariant to changes in platform specifics.

The underlying gist of a window provider is that we adopt browser conventions where ever possible. This happily minimizes the room for debate and possible surface area as we grow the feature set for each applications requirements.

### DIFF
--- a/src/client/src/rt-components/Environment/EnvironmentContext.tsx
+++ b/src/client/src/rt-components/Environment/EnvironmentContext.tsx
@@ -1,19 +1,39 @@
 import React from 'react'
 
-export interface EnvironmentValue {
-  isDesktop: Boolean
-  openfin?: Window
+export interface EnvironmentValue<Provider extends Window = Window> {
+  provider: Provider
+  openfin?: Provider | null
+  [key: string]: Provider | null
 }
 
 export interface Window {
+  type: string | 'desktop' | 'browser'
+  platform: string | 'openfin' | 'browser'
   maximize: () => void
   minimize: () => void
   close: () => void
   open: (url: string) => void
+  [key: string]: any
 }
 
-export const Environment = React.createContext<EnvironmentValue>({
-  isDesktop: false
-})
+export function createEnvironment<Provider extends Window>(provider: any): EnvironmentValue<Provider> {
+  return {
+    provider,
+    get [provider.platform]() {
+      return provider && provider.isPresent ? provider : null
+    },
+  }
+}
+
+export const Environment = React.createContext<EnvironmentValue<Window>>(
+  createEnvironment<Window>({
+    type: 'browser',
+    platform: 'browser',
+    maximize: () => {},
+    minimize: () => {},
+    close: () => {},
+    open: () => {},
+  }),
+)
 
 export default Environment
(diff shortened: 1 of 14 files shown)


src/client/src/rt-components/Environment/EnvironmentContext.tsx
```diff
@@ -1,19 +1,39 @@
...
-export interface EnvironmentValue {
-  isDesktop: Boolean
-  openfin?: Window
+export interface EnvironmentValue<Provider extends Window = Window> {
+  provider: Provider
+  openfin?: Provider | null
+  [key: string]: Provider | null
 }
 
 export interface Window {
```
keshaldesai: I think there is already a global Window type. We should rename this to avoid a conflict.

src/client/src/rt-components/Environment/EnvironmentContext.tsx
```diff
@@ -1,19 +1,39 @@
 import React from 'react'
 
-export interface EnvironmentValue {
-  isDesktop: Boolean
-  openfin?: Window
+export interface EnvironmentValue<Provider extends Window = Window> {
+  provider: Provider
+  openfin?: Provider | null
```
keshaldesai: openFin will be of type OpenFin

src/client/src/rt-components/Environment/EnvironmentContext.tsx
```diff
@@ -1,19 +1,39 @@
 import React from 'react'
 
-export interface EnvironmentValue {
-  isDesktop: Boolean
-  openfin?: Window
+export interface EnvironmentValue<Provider extends Window = Window> {
+  provider: Provider
```
keshaldesai: Instead of passing a Provider, create a Provider type that has the properties you're using (platform, type, etc).

src/client/src/rt-components/Environment/EnvironmentContext.tsx
keshaldesai: Also pull this to the top of this file.

evanrs: When requesting a rename it's good to offer an alternative — in this case I'm open to suggestions. In this case all I can imagine is `WindowProvider`

evanrs: Regarding order: as the subject of the file is the environment the order is by hierarchy of importance to the reader. If it were `WindowContext` than `Window` would be at the top.

src/client/src/rt-components/Environment/EnvironmentContext.tsx
evanrs: That's a matter for the end user as the Environment does not currently own knowledge about the provider other than a duck type of what must be satisfied.

src/client/src/rt-components/Environment/EnvironmentContext.tsx
evanrs: Unfortunately that breaks the typing — as this is preliminary its best to leave that to the subsequent improvements you'll bring.

PR #868, opened 2018-09-20
Isolate scroll style to elements within body

evanrs:
✨*Now with even more unrelated changes!*

### DIFF
--- a/src/client/src/shell/GlobalScrollbarStyle.tsx
+++ b/src/client/src/shell/GlobalScrollbarStyle.tsx
@@ -24,11 +24,19 @@ export const css = memoize(
     border: 4.5px solid rgba(0, 0, 0, 0);
     background-clip: padding-box;
     background-color: ${color};
-    box-shadow: inset -0px -0px 0 ${color}, inset 0px 0px 0 ${color};
   }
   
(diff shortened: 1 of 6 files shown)


src/client/src/shell/components/app-header/Header.tsx
```diff
@@ -93,6 +93,9 @@ const IconButton = styled.button<{ type: string }>`
     color: ${({ theme }) => theme.button.secondary.textColor};
   }
 `
+IconButton.defaultProps = {
+  role: 'button',
+}
```
keshaldesai: According to MDN:

"Where possible, it is recommended to use native HTML buttons (`<button>`, `<input type="button" />`,  `<input type="submit" />`, `<input type="reset" />` and `<input type="image" />`) rather than the button role, as native HTML buttons are supported by all user agents and assistive technology and provide keyboard and focus requirements by default, without need for additional customization."

evanrs: You can use a markdown quote by preceding it with `>`

evanrs: Resolved IRL.

PR #875, opened 2018-09-24
StyleGuide- added hex number to other colors

ianchan035:
Added the hex color name to the other colors


evanrs:
Since the colors should only be accessed through the generated theme is there a need to show the generated value?

Without having pulled this my guess is that it would uncenter the text — is that correct?
screen shot 2018-09-24 at 1 07 56 pm

PR #878, opened 2018-09-25
Fix persistent OpenFin processes

keshaldesai:
When closing OpenFin, the OpenFin process would sometimes linger.

By setting the `nonPersist` flag to true, OpenFin should close when we close the application.

According to the OpenFin docs, `nonPersist` is:
"A flag to configure the application as non-persistent. Runtime exits when there are no persistent apps running."


evanrs:
👍

PR #879, opened 2018-09-26
ARTP-279 Add simple launcher

evanrs:
screen shot 2018-09-26 at 5 21 13 pm


src/client/src/routes/SimpleLauncher/Link.tsx
```diff
@@ -0,0 +1,125 @@
+import React from 'react'
+import { styled, Styled } from 'rt-theme'
+
+declare const global: any
+
+export interface LinkProps {
+  is?: React.ComponentType<any>
```
keshaldesai: What do you think of calling this `as`?

src/client/src/routes/SimpleLauncher/config.tsx
```diff
@@ -0,0 +1,64 @@
...
+  },
+  {
+    name: 'Reactive Analytics',
+    url: 'https://demo-reactive-analytics.adaptivecluster.com/',
+    icon: <FontAwesomeIcon icon={faChartArea} />,
+    provider: {
+      platform: 'browser',
+    },
+  },
+  // {
```
keshaldesai: Do you need the rest of this?

src/client/src/routes/SimpleLauncher/Link.tsx
```diff
@@ -0,0 +1,125 @@
...
+  to: any
+  [key: string]: any
+}
+
+const LinkRoot: Styled<{ onClick: any }> = styled.button`
```
keshaldesai: Type this

src/client/src/routes/SimpleLauncher/Link.tsx
```diff
@@ -0,0 +1,125 @@
...
+  color: currentColor;
+  position: relative;
+  z-index: 100;
+
+  &:hover {
+    color: ${({ theme }) => theme.button.accent.backgroundColor};
+  }
+`
+
+export class Link extends React.Component<any> {
```
keshaldesai: Type the props

src/client/src/routes/SimpleLauncher/Link.tsx
```diff
@@ -0,0 +1,125 @@
...
+  }
+
+  render() {
+    const { is: LinkElement, children } = this.props
+
+    return <LinkElement onClick={this.onClick}>{children}</LinkElement>
+  }
+}
+
+export async function open({ name, url, provider, provider: { options } }: any) {
```
keshaldesai: I think the type here is `Window | fin.Desktop.Window`

src/client/src/routes/SimpleLauncher/Link.tsx
```diff
@@ -0,0 +1,125 @@
...
+
+export function createApp(options: any) {
+  return new Promise((resolve, reject) => {
+    const window: any = new fin.desktop.Application(
+      {
+        name: options.name,
+        uuid: options.name,
+        nonPersistent: false,
+        mainWindowOptions: options,
+      } as any,
```
keshaldesai: Lots of any, is there a reason you can't type these?

src/client/src/routes/SimpleLauncher/Link.tsx
keshaldesai: You can also do something like `React.ReactNode` or `JSX.Element`

src/client/src/routes/SimpleLauncher/Link.tsx
evanrs: It's kind of bonkers to type a standard API of React and a DOM element which ninety-nine percent of Styled Components will resolve to — right? 

It's beyond the concern of a Styled Component — if you can point out where the original typings fail I'd love to PR on that.

src/client/src/routes/SimpleLauncher/Link.tsx
evanrs: `is` already exists as a convention for redefining the final component used in the render.

On a note of implied ontological mean `as` suggests a `role` change and to me would mean a less literal application of the supplied value.

The correct change to the typing would actually be to mark it as `string | React.ComponentType<any>` — commonly it is used for `is="a"` or `is="button"`.

PR #881, opened 2018-10-17
Remove numericCellRenderer

keshaldesai:
This function was causing a bug when receiving bad data. We are planning on removing the currency icons anyways, so I removed the whole function.


evanrs:
👍

PR #885, opened 2018-11-13
[ARTP-310] Generic Platform API

keshaldesai:


src/client/src/rt-components/platform/context.ts
```diff
@@ -0,0 +1,4 @@
+import * as React from 'react'
+import { PlatformAdapter } from './adapters'
+
+export const { Provider: PlatformProvider, Consumer: PlatformConsumer } = React.createContext<PlatformAdapter>(null)
```
evanrs: Does `useContext` require the original context object?

PR #890, opened 2018-11-19
ARTP-322-blotter-commas

snony:
**Feature**:
The Notional values in the Blotter AG table was not formatted to 0,000,000.00. 

**Solution**:
defined a cell renderer function to return the correct format as mentioned above.


src/client/src/ui/blotter/components/blotterUtils.ts
```diff
@@ -2,11 +2,14 @@ import { ColDef } from 'ag-grid'
...
 import { formatDate, UtcFormatDate } from '../../spotTile/components/notional/utils'
 import SetFilter from './filters/SetFilter'
-
+import numeral from 'numeral'
 const dateRenderer = (trade: Trade, field: string) => {
   return formatDate(trade[field], '%d-%b %H:%M:%S')
 }
 
+const NotionalRenderer = (trade: Trade, field: string) => {
+  return numeral(trade[field]).format('0,000,000[.]00')
```
keshaldesai: [http://numeraljs.com/](url)

Try `0,0[.]00` instead of `0,000,000[.]00`.

snony: Thanks. I have updated it.

PR #895, opened 2018-11-20
[ARTP-296] Reduce scrollbar width to match zeplin design

FCartier:
**Issue :**
- The scrollbar was too thick and did not match the zeplin design

**Solution :**
- Make some changes in the GlobalScrollbarStyle.tsx file so as to make it look like the zeplin design


keshaldesai:
https://github.com/AdaptiveConsulting/ReactiveTraderCloud/compare/fix-scrollbar-styling?expand=1

FCartier:
@keshaldesai Your solution gets rid of the weird transparent border, thanks!

PR #897, opened 2018-11-21
ARTP-320 Add Margin to Popped Components

snony:
**Issue**:
When popping out Components such as Blotter in a stand alone window, Components should have a margin.

**Solution**:
The issue was not a margin but a padding.  Tile Components looked fine. Analytics is not working at the moment so I dont know. Added a padding of 16px to the Blotter component when in stand alone window.


src/client/src/shell/routes/BlotterRoute.tsx
```diff
@@ -1,10 +1,20 @@
 import React from 'react'
 import { RouteWrapper } from 'rt-components'
 import { BlotterContainer } from '../../ui/blotter'
+import { styled } from 'rt-theme'
+
+const BlotterContainerStyle = styled('div')`
+  height: 100%;
+  padding-left: 10px;
+  padding-right: 10px;
+  background-color: #272d3a;
```
keshaldesai: Get the color from theme.

PR #898, opened 2018-11-21
Horizontal launcher with logo

keshaldesai:


src/client/src/rt-components/icons/LogoIcon.tsx
```diff
@@ -1,9 +1,14 @@
...
+interface Props {
+  width: number
+  height: number
+}
+
+const LogoIcon: React.SFC<Props> = ({ width, height }) => (
   <svg
-    width="12px"
-    height="14px"
+    width={width * 16}
```
FCartier: Extract the 16 as a constant to facilitate future changes ?
```
const sizeMultiplier : number = 16;
---
width={width * sizeMultiplier}
height={height * sizeMultiplier}
```

PR #902, opened 2018-11-29
Artp 318/launch limit checker

snony:
**Feature**
- User should be able to download and launch Limit Checker.

**Solution**
- Added a config file with the download link of Limit Checker.
- Downloading and Launching Limit Checker works on RT OpenFin. However, you need to download it first, and then click the download button again and it will launch it.
- While in RT browser, it only downloads the Limit Checker. There is no way to launch it due to Browser security issue.


src/client/src/routes/SimpleLauncher/tools.ts
```diff
@@ -13,10 +13,13 @@ export async function open(config: ConfigType): Promise<Window | fin.OpenFinWind
       switch (provider.as) {
         case 'window':
           return createOpenFinWindow(config)
+        case 'download':
+          //either download LimitChecker or launches it if it exists.
```
danmcdaid: is this comment neccesary?

src/client/src/routes/SimpleLauncher/tools.ts
```diff
@@ -85,3 +86,31 @@ export function createOpenFinWindow({ name, url, provider: { options } }: Config
     )
   })
 }
+
```
danmcdaid: feels to me like single responsibility violation and `tools.ts` is starting to do too much (arguably was before this addition).  I'd consider at this point splitting out these handlers?

snony: I am going to leave it as is and will address this comment in the subsequent PR.

src/client/src/routes/SimpleLauncher/tools.ts
snony: Yes, it is not obvious to a user.

PR #903, opened 2018-12-04
style: bubble chart updated as per new designs

pdeleastar:
Features as per design
- circles now overlapping slightly
- shadow added to each circle

### DIFF
--- a/src/client/src/ui/analytics/components/positions-chart/chartUtil.ts
+++ b/src/client/src/ui/analytics/components/positions-chart/chartUtil.ts
@@ -24,7 +24,7 @@ export function getPositionsDataFromSeries(series: CurrencyPairPosition[] = [], 
   return _.map(positionsPerCcyObj, (val, key) => {
     return {
       symbol: key,
-      [baseAmountPropertyName]: val
+      [baseAmountPropertyName]: val,
     }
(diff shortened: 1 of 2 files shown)


src/client/src/ui/analytics/components/positions-chart/chartUtil.ts
```diff
@@ -137,9 +138,36 @@ export function getPositionValue(id: string, positionsData: any[]) {
...
+  const feMerge = filt.append('feMerge')
+
+  feMerge.append('feMergeNode').attr('in', 'offsetBlur')
+  feMerge.append('feMergeNode').attr('in', 'SourceGraphic')
+}
+
 export function collide(alpha: number, nodes: any[], scale?: number) {
   const qt = d3.geom.quadtree(nodes)
-  const padding = 10
+  const padding = -3
```
danmcdaid: Negative padding feels a bit dirty but I don't know this API so can't off-hand suggest how else you'd do this.

bhavesh-desai-scratch: I think this is a naming issue. We relate the word padding to CSS; In this case, it seems to be an offset. This code would be better if we use long names instead of a single letter everywhere.

pdeleastar: Renamed this to offset + more descriptive names added

PR #904, opened 2018-12-05
Artp 356/update pnl chart

snony:
**Feature**
- Update the PNLChart to Recharts LineChart.
- The LineChart should be styled according to the Zeplin Design.

**Solution**
- Converted the PNLChart to Recharts LineChart.
- Feature as per design: all texts follow Zeplin Design fonts; hovering over a datapoint displays the tooltip; the title; hr bar.

Things I could not get properly:
- The label on top of the Y-axis value. Could only position it to the center inner left. So I left that out.
- Having dashed axis.


src/client/src/ui/analytics/components/Analytics.tsx
```diff
@@ -6,12 +6,19 @@ import AnalyticsBarChart from './AnalyticsBarChart'
...
+  AnalyticsStyle,
+  BubbleChart,
+  LinearChartStyle,
+  LastPosition,
+  Title,
+  LastPositionWrapper,
+  USDspan,
+} from './styled'
+import AnalyticsHeader from './AnalyticsHeader'
+import ProfitLossLineChart from './AnalyticsLineChart'
```
danmcdaid: I understand why you've used a different name for the import, but I think potentially this is confusing and not a pattern I'd adopt wholesale.

src/client/src/ui/analytics/components/AnalyticsLineChart.tsx
```diff
@@ -0,0 +1,76 @@
...
+  x: Date
+  y: string
+}
+
+interface DataPoint {
+  x: string
+  y: string
+}
+
+const tickFormatYAxis = (x: string) => numeral(x).format('0.0a')
```
danmcdaid: I think the preference is to be explicit with return types

src/client/src/ui/analytics/components/AnalyticsLineChart.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+const getGradientOffset = (data: DataPoint[]) => {
```
danmcdaid: likewise

danmcdaid: Also I'm not a fan of this function (sorry!) or the name.  It returns a value that is used as an offset but what this function actually does is return a ratio of Max(val) / min-max difference.  I feel it bloats the component a bit and could perhaps live elsewhere.

src/client/src/ui/analytics/components/AnalyticsLineChart.tsx
```diff
@@ -0,0 +1,76 @@
...
+        <LineChart data={data} margin={{ top: 15, right: 50 }}>
+          <defs>
+            <linearGradient id="colorValue" x1="0" y1="0" x2="0" y2="1">
+              <stop offset={offset} stopColor="#28c988" stopOpacity={1} strokeWidth={2} />
+              <stop offset={offset} stopColor="#f94c4c" stopOpacity={1} strokeWidth={2} />
+            </linearGradient>
+          </defs>
+          <XAxis dataKey="x" tickLine={false} interval={'preserveStartEnd'} />
+          <YAxis tickFormatter={tickFormatYAxis} tickLine={false} interval={'preserveStartEnd'} />
+          {offset < 1 && <ReferenceLine y={0} stroke="white" strokeOpacity={0.2} />}
```
danmcdaid: Definitely personal preference but consider newline for `<ReferenceLine` for readability? We probably have tslint rules for this or should.

src/client/src/ui/analytics/components/TODO.txt
```diff
@@ -0,0 +1,13 @@
+TODO
```
danmcdaid: remove this file

src/client/src/ui/analytics/components/AnalyticsLineChart.tsx
```diff
@@ -0,0 +1,118 @@
...
+            <XAxis
+              dataKey="x"
+              tickLine={false}
+              width={400}
+              ticks={[dataPoints[0].x, ...dataPoints.map(point => point.x)]}
+              interval="preserveStartEnd"
+            />
+            {offset < 1 && <ReferenceLine y={0} stroke="white" strokeOpacity={0.3} strokeWidth={1} />}
+            {dataPoints.map(point => (
+              <ReferenceLine key={point.x} x={point.x} {...lineProps} />
```
danmcdaid: nitpik: could destructure x from point

src/client/src/ui/analytics/components/AnalyticsLineChart.tsx
```diff
@@ -0,0 +1,118 @@
...
+      const offset = this.state.offset + 1 === this.intervalWidth ? 0 : this.state.offset + 1
+      this.setState({ offset })
+    }
+  }
+  render() {
+    const {
+      model: { seriesData },
+    } = this.props
+    const data = seriesData.map(point => ({ x: moment(point.x).format('hh:mm:ss A'), y: point.y }))
+    const offset = getLinearGradientOffset(data)
```
danmcdaid: nitpik: destructure point?

src/client/src/ui/analytics/components/AnalyticsLineChart.tsx
```diff
@@ -0,0 +1,118 @@
...
+              padding={{ top: 0, bottom: 0 }}
+              axisLine={false}
+              tickFormatter={tickFormatYAxis}
+            />
+
+            <XAxis
+              dataKey="x"
+              tickLine={false}
+              width={400}
+              ticks={[dataPoints[0].x, ...dataPoints.map(point => point.x)]}
```
danmcdaid: `dataPoints[0].x` is covered in the next parameter? 🤔

src/client/src/ui/analytics/components/LastPosition.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+interface FormattedLastPosition {
+  lastPosition: string | undefined
+  color: string
+}
+
+const formatLastPosition: (lastPos: number) => FormattedLastPosition = (lastPos: number) => {
+  let lastPosition = numeral(lastPos).format()
+  let color = ''
+  if (lastPos > 0) {
```
danmcdaid: would a ternary simplify this? `const color = lastPos > 0 ? 'green' : lasPos < 0  ? ' red' : ''`

src/client/src/ui/analytics/components/LastPosition.tsx
```diff
@@ -0,0 +1,37 @@
...
+interface LastPositionProps {
+  lastPos?: number | undefined
+}
+
+interface FormattedLastPosition {
+  lastPosition: string | undefined
+  color: string
+}
+
+const formatLastPosition: (lastPos: number) => FormattedLastPosition = (lastPos: number) => {
```
danmcdaid: The naming here isn't optimal!  The function takes a number and returns a type named `FormattedLastPosition` - but that type represents something to do with `color` also and not just a formatted version of the number that was provided.  I'd rethink the type and method name.

src/client/src/ui/analytics/components/LastPosition.tsx
```diff
@@ -0,0 +1,37 @@
...
+  let lastPosition = numeral(lastPos).format()
+  let color = ''
+  if (lastPos > 0) {
+    color = 'green'
```
danmcdaid: Also it would be better to use string enums to represent the colours eg.
```
enum IndicatorColors = {
  RED = 'red',
  GREEN = 'green,
}
IndicatorColors.RED etc
```

src/client/src/ui/analytics/components/AnalyticsLineChart.tsx
```diff
@@ -0,0 +1,118 @@
...
+      <ToolTipChildRight>{payload.length > 0 && numeral(payload[0].value).format('0.0a')}</ToolTipChildRight>
+    </ToolTipStyle>
+  )
+}
+
+interface LineChartState {
+  offset: number
+}
+class LineCharts extends React.PureComponent<LineChartProps, LineChartState> {
+  private intervalWidth: number = 30
```
danmcdaid: Perhaps this is a not a requirement right now, but the width of the interval probably should change depending on the window width.

danmcdaid:
@snony I don't mean to make this painful, sorry!

PR #908, opened 2018-12-20
Artp 385/styleguide clipboard

snony:
**Feature**
- When User hovers over the colour blocks at RT Styleguide, they should see the copy icon.
-  If the user clicks on the copy icon, they should be alerted that the color associated to the colour block has been copied to clipboard.

**Solution**
- Created a separate file that has the component associated with the copy icon.
- Using react-copy-to-clipboard to copy the "colour block" colour to clipboard.


src/client/src/routes/StyleguideRoute/components/ColorToClipBoard.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+interface ColorToClipBoardProps {
+  color: string
+  iconColor: string
+}
+export const CopyToClipBoardWrapper = styled.div`
+  visibility: hidden;
+`
+const onClick = () => {
+  alert('copied to clipboard')
```
danmcdaid: Is `alert` the required to be the behaviour?

src/client/src/routes/StyleguideRoute/components/ColorToClipBoard.tsx
```diff
@@ -0,0 +1,30 @@
+import React from 'react'
+import { library } from '@fortawesome/fontawesome-svg-core'
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
+import { faCopy } from '@fortawesome/free-solid-svg-icons'
+import { styled } from 'rt-theme'
+
+library.add(faCopy)
+import CopyToClipboard from 'react-copy-to-clipboard'
+
+interface ColorToClipBoardProps {
```
danmcdaid: Naming inconsistency, line 8 the original import kebab-case to PascalCase is correct.  But on line 10 and elsewhere `Clipboard` becomes `ClipBoard` which isn't consistent.  I'd either alter line 8 to match elsewhere be consistent with the original kebab case import.

src/client/src/routes/StyleguideRoute/components/ColorToClipBoard.tsx
snony: As far as I understood the requirements, I think so.

PR #909, opened 2018-12-20
Artp 373/reduce theme complexity

keshaldesai:
Updates:
* Reduce theme complexity
* Generalize colors into names (e.g. 'lightBackground')
* Switch to styled-components (we get full benefits of v4 including dynamic global styles, and direct refs)
* Remove emotion
* Remove react-app-rewired
* Remove extra babel packages

TODO:
* Better types
* Simplify theme creation (will require heavy work around style guide)
* Once theme creation is simplified, consolidate colors into a defaultTheme file


src/client/package.json
```diff
@@ -92,14 +88,12 @@
     "reselect": "3.0.1",
     "ress": "^1.2.2",
     "rxjs": "6.1.0",
+    "styled-components": "^4.1.3",
```
snony: Are we using this one (styled-components) or just react-emotion?

src/client/src/routes/OrderTicketRoute/Notification.tsx
```diff
@@ -107,7 +108,7 @@ Position.defaultProps = {
   position: 'bottom',
 }
 
-const Transistion: Styled<{ visible?: boolean; position?: 'top' | 'bottom' }> & { [k: string]: any } = styled(Block)`
+const Transistion: any = styled(Block)<{ visible?: boolean; position?: 'top' | 'bottom' }>`
```
snony: Based on @bhavesh-desai-scratch :  we should Type every variable and avoid using any type

src/client/src/routes/OrderTicketRoute/TextField.tsx
```diff
@@ -93,10 +93,10 @@ export const Input = styled.input`
   width: 100%;
   max-width: 100%;
 
-  ${curryProps(mapColorProps, { fg: 'secondary.base' })};
+  ${curryProps<any>(mapColorProps, { fg: 'secondary.base' })};
```
snony: type this one maybe?

src/client/src/routes/StyleguideRoute/styled/Color.tsx
```diff
@@ -13,8 +13,8 @@ export interface ColorProps {
 export const colorProps: MappedPropMap<ColorProps> = {
   backgroundColor: v => (colorProps.bg as MappedPropFn<ColorProps>)(v),
   textColor: v => (colorProps.fg as MappedPropFn<ColorProps>)(v),
-  bg: ({ theme, backgroundColor, bg = backgroundColor }) => bg && { backgroundColor: getColor(theme, bg) },
-  fg: ({ theme, color, textColor, fg = textColor || color }) => fg && { color: getColor(theme, fg) },
+  //bg: ({ theme, backgroundColor, bg = backgroundColor }) => bg && { backgroundColor: getColor(theme, bg) },
```
snony: Perhaps, remove this if it is no longer needed.

src/client/src/routes/OrderTicketRoute/Notification.tsx
keshaldesai: Unfortunately typescript has an issue with using the `Component.from` syntax so I had to use any here. I think this component should just be rewritten in the future to use `keyframes`.

keshaldesai:
@danmcdaid You can still have nested/child theming. Example: `/src/client/src/ui/status-bar/StatusBar.tsx`

By just using the `ThemeProvider` from `styled-components`, You can pull off the old theme using `withTheme`, or you can just supply an object.

Unfortunately, you will lose the typing of the selected or new theme. However, we did not have typing using the old method anyways. We would need to make a new `styled` with the type of the nested theme and only use that to create styled components after that point.

PR #924, opened 2019-01-10
Artp 358/padding analytics

snony:
**Feature**
- The Analytics window on popout should have a padding.

**Solution**
- Added a padding to the AnalyticsRoute component when Analytics Component is on popout.

### DIFF
--- a/src/client/src/shell/routes/AnalyticsRoute.tsx
+++ b/src/client/src/shell/routes/AnalyticsRoute.tsx
@@ -3,7 +3,7 @@ import { RouteWrapper } from 'rt-components'
 import { AnalyticsContainer } from '../../ui/analytics'
 import { styled } from 'rt-theme'
 
-const AnalyticsStyle = styled.div`
+const AnalyticsRouteStyle = styled.div`
   height: 100%;
   padding: 10px;
   background-color: ${({ theme }) => theme.core.darkBackground};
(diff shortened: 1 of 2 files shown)


src/client/src/shell/routes/AnalyticsRoute.tsx
```diff
@@ -1,10 +1,19 @@
 import React from 'react'
 import { RouteWrapper } from 'rt-components'
 import { AnalyticsContainer } from '../../ui/analytics'
+import { styled } from 'rt-theme'
 
+const AnalyticsRouteStyle = styled.div`
+  height: 100%;
+  padding: 10px;
```
keshaldesai: Use `rem` please

src/client/src/shell/routes/ShellRoute.tsx
```diff
@@ -88,7 +88,7 @@ const portalProps = {
     config: {
       name: 'analytics',
       width: 400,
```
keshaldesai: Do you also need to increase the `width` if you increased padding?

snony: Yeah, I chose that over changing the property of the Analytics Container (overflow-y) because of the convention that all other components in pop out mode do not have a overflow-y set to auto but hidden.

PR #925, opened 2019-01-14
Fix notifications

keshaldesai:


src/client/src/routes/NotificationRoute.tsx
```diff
@@ -14,28 +13,48 @@ interface State {
...
   render() {
     const { message } = this.state
-    return message == null ? null : (
+    return message ? (
       <ThemeProvider>
         <TradeNotification message={message.tradeNotification} dismissNotification={this.onDismissNotification} />
       </ThemeProvider>
+    ) : (
+      <ThemeProvider>
+        <TradeNotification message={debugTrade} dismissNotification={() => console.log('Dismiss notification')} />
```
snony: Do we really need to console log here?

src/client/src/routes/NotificationRoute.tsx
```diff
@@ -14,28 +13,48 @@ interface State {
   message: Message | null
 }
 
+const debugTrade: Trade = {
```
snony: Wouldn't it be better and easier to use elsewhere if one extracted the strings defined in this variable to an enum instead?

PR #942, opened 2019-01-22
feat: interop services added to platform object

pdeleastar:


src/client/src/ui/analytics/epics/index.ts
```diff
@@ -1,11 +1,15 @@
 import { combineEpics } from 'redux-observable'
 import { publishPositionUpdateEpic } from './analyticsServiceEpic'
 import { analyticsServiceEpic } from './epics'
+import { InteropServices, PlatformAdapter } from 'rt-components'
 
-const epics = [analyticsServiceEpic]
+export default ({ platform }: { platform: PlatformAdapter }) => {
```
snony: Perhaps move this to another file and export default the function here because of the convention we follow in index files.

src/client/src/ui/blotter/epics/index.ts
```diff
@@ -2,10 +2,15 @@ import { combineEpics } from 'redux-observable'
 import { publishBlotterEpic } from './blotterServiceEpic'
 import { blotterServiceEpic } from './epics'
 import { connectBlotterToNotifications, requestBrowserNotificationPermission } from './blotterServiceEpic'
+import { InteropServices, PlatformAdapter } from 'rt-components'
 
-const epics = [blotterServiceEpic, connectBlotterToNotifications, requestBrowserNotificationPermission]
-if (typeof fin !== 'undefined') {
-  epics.push(publishBlotterEpic)
-}
+export default ({ platform }: { platform: PlatformAdapter }) => {
```
snony: Last comment applies here as well.

src/client/src/ui/spotTile/epics/index.ts
```diff
@@ -5,11 +5,19 @@ import { pricingServiceEpic } from './pricingEpics'
 import { publishPriceUpdateEpic } from './publishPrice'
 import { spotTileEpic } from './spotTileEpics'
 import { publishTradeExecutedEpic } from './tradeExecutedEpic'
+import { InteropServices, PlatformAdapter } from 'rt-components'
 
-const epics = [spotTileEpic, pricingServiceEpic]
+export default ({ platform }: { platform: PlatformAdapter }) => {
```
snony: Same comment as the last one applies here

src/client/src/ui/analytics/epics/index.ts
danmcdaid: It seems that the type of the parmater is `ApplicationDependencies` and you should use it accordingly

src/client/src/rt-components/platform/index.ts
```diff
@@ -1,5 +1,5 @@
 export { default as Platform } from './platform'
-export { PlatformAdapter, WindowConfig, openFinNotifications, setupGlobalOpenfinNotifications } from './adapters'
+export { PlatformAdapter, WindowConfig, openFinNotifications, setupGlobalOpenfinNotifications, InteropServices } from './adapters'
```
bhavesh-desai-scratch: You should not need to export InteropServices. You sure this is necessary?

src/client/src/ui/analytics/epics/index.ts
```diff
@@ -1,11 +1,15 @@
...
+  const { interopServices }: { interopServices: InteropServices } = platform
```
bhavesh-desai-scratch: You should not need to specify the type here. Just use ApplicationDependencies as Dan suggests.

Also, you do not always need to use a deconstructor, especially for nested properties.

`const interopServices = platform.interopServices`

src/client/src/ui/analytics/epics/index.ts
pdeleastar: Included that type now

pdeleastar: Moved out of index now

PR #943, opened 2019-01-22
ARTP-411/react-scripts

snony:


src/client/.storybook/webpack.config.js
```diff
@@ -20,7 +20,7 @@ module.exports = (baseConfig, env, defaultConfig) => {
     system: path.resolve(__dirname, '../src', 'system'),
     shell: path.resolve(__dirname, '../src', 'shell')
   }
-  defaultConfig = require('../config-overrides')(defaultConfig, env)
+  // defaultConfig = require('../config-overrides')(defaultConfig, env)
```
danmcdaid: If you don't need this be brave and delete, it can always be reverted from a previous commit.  Commented lines are likely to get forgotten about and then nobody knows why it was commented and not removed.

snony: Yeah. I will delete it. There is no config-overrides anymore.

src/client/package.json
```diff
@@ -62,6 +62,7 @@
     "lodash": "^4.17.10",
     "luxon": "^1.4.4",
     "moment": "^2.22.2",
+    "nodejs": "0.0.0",
```
bhavesh-desai-scratch: You sure you need this

snony: Removing it now.

src/client/package.json
```diff
@@ -99,11 +102,12 @@
     "@types/jest": "^22.2.3",
     "@types/karma": "1.7.3",
     "@types/lodash": "4.14.100",
+    "@types/node": "^10.12.18",
```
bhavesh-desai-scratch: Do you really need this?

src/client/src/simpleTest.test.ts
```diff
@@ -0,0 +1,3 @@
+test('Simple Test to make test on CircleCI pass', () => {
+  expect(1).toBe(1)
+})
```
RxDave: Are you sure this test is necessary? Adding a unit test (or any no-op code) just to make a tool happy is a code smell :)

snony: Yeah. Because currently we do not have any test and when circleCI pulls from repo it does the react-scripts test which fails when there is no test.

PR #944, opened 2019-01-23
Browser fix

pdeleastar:
added check for Notification support in browser


src/client/src/ui/blotter/epics/blotterServiceEpic.ts
```diff
@@ -65,7 +68,9 @@ export const requestBrowserNotificationPermission: ApplicationEpic = (action$, s
   action$.pipe(
     applicationConnected,
     tap(() => {
-      Notification.requestPermission()
+      if ('Notification' in window) {
```
bhavesh-desai-scratch: why not just if(window.Notification)

PR #947, opened 2019-01-23
Artp 395/browser tools

snony:
**Feature:**
- Enable styled-component plugin macro in order to make debug easier.

**Solution**
- Switch to react-scripts and changed all import {styled} from rt-theme to import styled from ''styled-components/macro".


src/client/src/simpleTest.test.ts
```diff
@@ -0,0 +1,3 @@
+test('Simple Test to make test on CircleCI pass', () => {
+  expect(1).toBe(1)
+})
```
RxDave: Required? Maybe right a good test instead, if one is actually necessary.

snony: @RxDave  rt-theme functionality is as it used to be to create the theme. The styled macro works locally in the file which is being parsed. But because we have the ThemeProvider on the root level, this injects the styling from the theme itself the parser saw a styled component. But the styled macro takes that and attach the Component name before generating a classname for the component.

snony: @RxDave , there is a defined file called style.d.ts which import the rt-theme and redeclare the styled-components macro in order to type it.

snony:
No Longer needed.

PR #949, opened 2019-01-28
ARTP-395/browser-tools

snony:
This is an update of a previous PR which had a lot of changes due to styled-component macro. @bhavesh-desai-scratch told me try and work out things with react-app-rewired in order to add it directly to the babel plugin. The solution is achieved by using 'react-app-rewire-styled-components'.


src/client/config-overrides.js
```diff
@@ -0,0 +1,6 @@
+const rewireStyledComponents = require('react-app-rewire-styled-components');
+
+module.exports = function override(config, env) {
+    config = rewireStyledComponents(config, env);
```
bhavesh-desai-scratch: just return

PR #950, opened 2019-01-28
DO NOT MERGE

RxDave:
Diagnosing CI failures in the launcher regarding populating the event store timing out.


RxDave:
At this point, I have not discovered the root cause of the error because I was unable to trigger a CI build with this pull request. Since this pull request was made, I've moved onto unit testing instead, so I'll close it for now.

PR #955, opened 2019-02-06
Add tests to ServiceStub Request-Reply messaging

bhavesh-desai-scratch:
Writing a test for this class revealed a bug. We needed a `take(1)` to clean up the subscription, also changed the `mergeMap` to `switchMap`


src/client/src/rt-system/__tests__/ServiceStub.test.ts
```diff
@@ -0,0 +1,73 @@
...
+
+describe('ServiceStub', () => {
+  const payload = 'payload'
+  const replyTo = 'responseTopic'
+  const Username = 'USER'
+  const procedure = 'procedure'
+  const result = 'result'
+
+  describe('Request Response', () => {
+    it('invokes the a remote procedure with the correct payload', () => {
```
snony: Minor issue but you have "the a"

src/client/src/rt-system/ServiceStub.ts
```diff
@@ -40,58 +37,57 @@ export class ServiceStub {
    */
   subscribeToTopic<T>(topic: string, acknowledgementObs?: NextObserver<string>): Observable<T> {
```
snony: There is no test associated with subscribeToTopic? I could not find any.

src/client/src/rt-system/__tests__/ServiceStub.test.ts
```diff
@@ -0,0 +1,73 @@
...
+      const session = new MockSession()
+      const stub = new ServiceStub(Username, of(createMockConnection(session)))
+      const rpc = stub.requestResponse<string, string>(procedure, payload, replyTo).subscribe()
+      expect(session.call).lastCalledWith(procedure, [{ payload, replyTo, Username }])
```
snony: Correct me if I am wrong, but I think it is redundant to test whether it returns the correct payload as the responsibility for a correct payload lies with the back-end side. As such, I will assume that it shouldn't be tested.

src/client/src/rt-system/__tests__/ServiceStub.test.ts
```diff
@@ -0,0 +1,73 @@
...
+    })
+
+    it('returns the result and completes after a connection is established', () => {
+      new MockScheduler().run(({ cold, expectObservable }) => {
+        const variables = { c: createMockConnection(new MockSession()), r: result }
+
+        //emit connection event after two frames
+        const connection$ = cold<ConnectionEvent>('--c', variables)
+        const stub = new ServiceStub(Username, connection$)
+        const expected = '--(r|)'
```
snony: The marble diagram here tells me that it emits the result and complete at the same time. I dont know if that's possible. I would assume that the marble diagram should be '--r|' .

bhavesh-desai-scratch: No it is possible as it happens in the same frame (otherwise the test would fail)

src/client/src/rt-system/__tests__/ServiceStub.test.ts
bhavesh-desai-scratch: We are testing that we call the session object with the correct payload. We are also testing that the correct value is returned when from the session object. The session object is mocked.

src/client/src/rt-system/ServiceStub.ts
bhavesh-desai-scratch: That is next

src/client/src/rt-system/__tests__/ServiceStub.test.ts
bhavesh-desai-scratch: good find

PR #956, opened 2019-02-06
style: blotter width + height changed to 100%

pdeleastar:


src/client/src/shell/routes/BlotterRoute.tsx
```diff
@@ -5,9 +5,9 @@ import { styled } from 'rt-theme'
 
 const BlotterContainerStyle = styled('div')`
   min-height: 28.125rem;
```
snony: Since height is defined to 100 %, min-height is no longer needed?

pdeleastar: removed min height + width

PR #957, opened 2019-02-07
Artp 431/finsemble interop

pdeleastar:
Finsemble platform object additions

app.open - now uses finsemble spawn api 
https://documentation.chartiq.com/finsemble/LauncherClient.html#spawn

interop - uses Finsemble RouterClient api
https://documentation.chartiq.com/finsemble/tutorial-TheRouter.html

changed currency chart URL - directed at new setup which has interop message handlers for both OF + FSBL


src/client/src/rt-components/platform/adapters/platformAdapter.ts
```diff
@@ -19,8 +19,7 @@ export interface PlatformAdapter {
   }
 
   interop?: {
-    subscribe: (sender: string, topic: string, listener: () => void) => void
-    unsubscribe: (sender: string, topic: string, listener: () => void) => void
+    subscribe$: (topic: string) => any
```
bhavesh-desai-scratch: I thought this returns an observable

PR #969, opened 2019-02-20
OrderTicket refactor: take 1

spcfran:
My first PR!

- Simplified Timer component and moved under rt-components (as it's not OrderTicket specific)
- Strongly typed Timer component to enforce one and only one of timeout/interval props
- Strongly typing of OrderTicket button handlers


src/client/src/rt-util/index.ts
```diff
@@ -1,5 +1,6 @@
 import { ActionCreatorsMapObject } from 'redux'
 export { withDefaultProps } from './reactTypes'
 import { action, ActionUnion } from './ActionHelper'
+export * from './utilityTypes'
```
snony: export * breaks the convention.  The convention is `import {someTypes} from someModule` then `export type someTypes = someTypes`.

src/client/src/rt-components/timer/Timer.tsx
```diff
@@ -0,0 +1,36 @@
...
+export class Timer extends PureComponent<TimerProps> {
+  id: number
+
+  componentDidMount() {
+    const { interval, timeout, duration, immediate } = this.props
+
+    this.id = interval ? window.setInterval(interval, duration) : window.setTimeout(timeout, duration)
+
+    if (immediate) {
+      ;(interval || timeout)()
```
snony: Did you misplace the semi colon?

snony:
Also, we are using circleCI for sanity check to ensure that all the tests pass.

src/client/src/rt-components/timer/Timer.tsx
spcfran: not sure how that got in there 😳

spcfran: Apparently the `;` was added by prettier during a precommit task, and [this](https://github.com/prettier/prettier/issues/1244) seems to be the reason. 

There was a similar construct in the previous version of the code, I wondered the same when I first saw it.

spcfran:
Closed in favor of #970

PR #973, opened 2019-02-25
Use hooks for platform context

bhavesh-desai-scratch:
Remove withPlatform
Remove defaultProps type
Remove tsconfig files no longer on use
Fix types for interop


src/client/src/rt-components/tear-off/ExternalWindow.tsx
```diff
@@ -12,51 +11,46 @@ const defaultConfig: WindowConfig = {
...
-    this.externalWindow = await platform.window.open(config, this.release)
+  let externalWindow: Window
 
-    if (this.externalWindow) {
-      window.addEventListener('beforeunload', this.release)
+  const release = () => {
+    if (externalWindow) {
+      window.removeEventListener('beforeunload', release)
     }
+    onUnload.call(null)
```
spcfran: Default value for onUnload is null but we are calling it here without null checking.

src/client/src/rt-components/tear-off/ExternalWindow.tsx
```diff
@@ -12,51 +11,46 @@ const defaultConfig: WindowConfig = {
...
+  config = defaultConfig,
+}) => {
+  const platform = usePlatform()
 
-    this.externalWindow = await platform.window.open(config, this.release)
+  let externalWindow: Window
 
-    if (this.externalWindow) {
-      window.addEventListener('beforeunload', this.release)
+  const release = () => {
```
spcfran: declaring `release` like this on a functional component will overwrite the reference to the function on every render, so by the time we do
```
window.removeEventListener('beforeunload', release)
```
below, `release` could no longer be the same function we added the event listener to. See [this](https://stackoverflow.com/questions/53845595/wrong-react-hooks-behaviour-with-event-listener) similar issue on stackoverflow.

When using hooks we may want to use something like [useCallback](https://reactjs.org/docs/hooks-reference.html#usecallback).

src/client/src/rt-components/tear-off/ExternalWindow.tsx
```diff
@@ -12,51 +11,46 @@ const defaultConfig: WindowConfig = {
...
+const ExternalWindow: FC<ExternalWindowProps> = ({
+  title = '',
+  onBlock = null as () => void,
+  onUnload = null as () => void,
+  config = defaultConfig,
+}) => {
+  const platform = usePlatform()
 
-    this.externalWindow = await platform.window.open(config, this.release)
+  let externalWindow: Window
```
spcfran: We may want to use the `useRef` hook here so that the variable is preserved across renders. As per [this article](https://reactjs.org/docs/hooks-faq.html#is-there-something-like-instance-variables) it's supposed to be the equivalent of instance fields in a class component

src/client/src/rt-components/tear-off/ExternalWindow.tsx
bhavesh-desai-scratch: The TS definitions require a null here. The default its `{}`

src/client/src/rt-components/tear-off/ExternalWindow.tsx
bhavesh-desai-scratch: Good catch.

src/client/src/rt-components/tear-off/ExternalWindow.tsx
spcfran: I mean the `onUnload` prop can be null, and is in fact the default:
```
const ExternalWindow: FC<ExternalWindowProps> = ({
  onUnload = null as () => void,
```
Note that this problem was already there before this PR, just that the only place in which the component is used is providing a valid function so it hasn't surfaced before. I'd suggest something along the lines of `onUnload && onUnload.call(null)` or `typeof onUnload === 'function'` or similar

PR #975, opened 2019-02-27
Refactor orderticket 3: static typing for styles, general tidy up

spcfran:


src/client/src/routes/OrderTicketRoute/Notification.tsx
```diff
@@ -29,37 +30,26 @@ const intents = {
...
+      duration: 250,
+      onLeave: onEnd,
+    },
+  )
+  const visible = stage === 0
 
   return (
     <Position position={position} p={1}>
-      <Transition key={visible ? 'show' : 'hide'} visible={visible} position={position}>
+      <Transition key={visible ? 'show' : 'hide'} visible={visible} position={position} duration={200}>
```
snony: duration props here has been provided a value of 200, while there is a duration provided as a parameter as well. Were they meant to be the same thing?

spcfran: They are different things - The `duration` as a prop for the `<Notification>` component is the time the notification should be visible on screen, whereas the inner `duration` passed to the `<Transition>` component is the time for the css transition (i.e. the duration of the slide-in / slide-out animations)

PR #981, opened 2019-03-05
Artp 325/add price history

snony:


src/client/src/ui/spotTile/spotTileReducer.ts
```diff
@@ -34,7 +34,7 @@ const spotTileReducer = (
     case TILE_ACTION_TYPES.SPOT_TILE_SUBSCRIBE:
       return state
     case TILE_ACTION_TYPES.SPOT_PRICES_UPDATE:
-      return { ...state, price: action.payload }
+      return { ...state, price: action.payload, historicPrices: [...state.historicPrices.slice(0), action.payload] }
```
spcfran: what is the intention with `[...state.historicPrices.slice(0), action.payload]` ?
Note that `state.historicPrices.slice(0)` is an array with the same elements as `state.historicPrices`, which is the same effect as spreading the array. Unless I'm missing something:
- if you intend to remove the first element of the array, should use `.slice(1)`
- if you intend to copy all elements and append the payload, just `[...state.historicPrices, action.payload]` will do

src/client/src/ui/spotTile/components/analyticsTile/AnalyticsTileChart.tsx
```diff
@@ -1,40 +1,44 @@
...
-            <ReferenceLine y={0} {...lineProps} />
-          </LineChart>
-        </ResponsiveContainer>
-      </AnalyticsTileChartWrapper>
-    )
-  }
+interface Props {
+  history: SpotPriceTick[]
+}
+const AnalyticsTileChart: FC<Props> = ({ history }) => {
```
spcfran: In some places throughout the project we are doing:
```typescript
import React from 'react'
const SomeComponent: React.FC<Props> ...
```
whereas in others
```typescript
import React, { FC } from 'react'
const SomeComponent: FC<Props> ...
```
The same applies to class components (`extends React.Component` VS `extends Component`)

It's not a huge deal but I think we should consistently follow one - I'd vote for the first one as it's currently the approach used in most places

src/client/src/ui/spotTile/spotTileReducer.ts
snony: Oups, thanks for spotting that out, I meant to use slice(1). As new price data is coming in, I am removing the first one. So, I am essentially moving historicPrices by one value to the left.

src/client/src/ui/spotTile/components/analyticsTile/AnalyticsTileChart.tsx
bhavesh-desai-scratch: I like the latter

PR #986, opened 2019-03-06
Artp 461/refactor theme

spcfran:


spcfran:
Not really expecting anyone goes through all the crap in this PR but I need an approval :) Happy to go through it if you want @bhavesh-desai-scratch @snony

PR #996, opened 2019-03-13
Handling commas and exceeding values in `NotionalInput`

gablabelle:
:warning: We still need to validate a few things.

We now show commas in field when inputing number. Resets to DEFAULT_NOTIONAL_VALUE when input value exceeds MAX_NOTIONAL_VALUE.

When finished:
Screen Shot 2019-03-13 at 6 11 56 PM

When inputting (Before):
Screen Shot 2019-03-13 at 6 12 12 PM

When inputting (After):
Screen Shot 2019-03-13 at 6 10 47 PM

When the value exceeds `MAX_NOTIONAL_VALUE`, it is set back to `DEFAULT_NOTIONAL_VALUE` and a red line appear to indicate that the entry was invalid. It disappears after 2 seconds.

Screen Shot 2019-03-13 at 6 11 22 PM (2)

We need to validate these points: 

- [x] `DEFAULT_NOTIONAL_VALUE` is currently hardcoded to `1000000`, maybe it should be `0` ?
- [x] the UI for the input error
- [x] we should probably also show a message or tooltip or the end user will not know what was the problem probably.
- [ ] we should probably also imprement increase/decrease of the number using up/down arrows? It should behave like a number type input field.

⚠️ I did not leave the entered value in the field when exceeding `MAX_NOTIONAL_VALUE` because you don't want the end user to Buy or Sell with an invalid value entered.

### DIFF
--- a/src/client/src/ui/spotTile/components/Tile.tsx
+++ b/src/client/src/ui/spotTile/components/Tile.tsx
@@ -6,6 +6,7 @@ import numeral from 'numeral'
 import { AnalyticsTile } from './analyticsTile'
 import { TileViews } from '../../workspace/workspaceHeader'
 import { DEFAULT_NOTIONAL_VALUE } from './notional/NotionalInput'
+
 interface Props {
   currencyPair: CurrencyPair
   spotTileData: SpotTileData
@@ -16,17 +17,23 @@ interface Props {
 
 interface State {
   notional: string
+  inError: boolean
 }
 
 class Tile extends React.PureComponent<Props, State> {
   state = {
     notional: DEFAULT_NOTIONAL_VALUE,
+    inError: false, // should probably be a list of errors at some point
   }
+
   tileComponents = {
     [TileViews.Normal]: SpotTile,
     [TileViews.Analytics]: AnalyticsTile,
   }
+
   updateNotional = (notional: string) => this.setState({ notional })
+
+  setInErrorStatus = (inError: boolean) => this.setState({ inError })
 
   executeTrade = (direction: Direction, rawSpotRate: number) => {
     const { currencyPair, executeTrade } = this.props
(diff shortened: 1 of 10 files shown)


src/client/src/ui/spotTile/components/Tile.tsx
```diff
@@ -15,18 +17,24 @@ interface Props {
 
 interface State {
   notional: string
+  disabled: boolean
 }
 
 class Tile extends React.PureComponent<Props, State> {
   state = {
-    notional: '1000000',
+    notional: DEFAULT_NOTIONAL_VALUE,
```
spcfran: Any reason why we are storing the `notional` state at Tile level as a string instead of numeric? I know it was already the case before this PR, but it's a good opportunity to bring up the discussion.

I think it should be the input itself who deals with the commas, the shorthand notation etc, and expose a clean, valid number to the Tile component (unless I'm missing something!)

gablabelle: @spcfran I've asked myself the same question and then I realized that the component was using an input field of `text` type and that the user can enter shortcuts, ie: `1k` for `1,000` and `1m` for `1,000,000` and that we use `numeral(notional).format()` to return a string to render a correctly formatted number ie: `1,234,567.89`.

spcfran: Yep makes sense to treat it as a string for that reason, but dealing with the commas, the `1k`/`1m` notation etc. is the responsibility of the `NotionalInput` component right? That component could handle the state as a string (either as a member on its own state, or directly referring to the `inputRef.value` as it is doing right now and is fine), but then call `this.props.updateNotional` with a parsed number. In fact, the only time `updateNotional` is called is right after doing `numeral(inputValue)`, so we could pass whatever number is (if valid) or `null` if not (`numeral("string").value()` returns `null` apparently) to the Tile component. 

No big deal though as it's working perfectly fine right now, so it's probably better to focus in more important things :)

PR #999, opened 2019-03-18
Artp 492/limit checker

snony:


src/client/src/shell/openFin/openFin.ts
```diff
@@ -2,13 +2,16 @@ import { Observable } from 'rxjs'
...
-
+const LIMIT_CHECKER_UUID = 'LIMIT-CHECKER'
+const LIMIT_CHECKER_STATUS_TOPIC = 'request-limit-check-status'
 export class OpenFinLimitChecker {
   private limitCheckSubscriber: string | null = null
   private limitCheckId: number = 1
 
   constructor() {
     if (typeof fin !== 'undefined') {
+      this.setLimitCheckSubscriber = this.setLimitCheckSubscriber.bind(this)
```
spcfran: Why don't we just use an arrow function instead of this-binding here? i.e.
```typescript
// Was: setLimitCheckSubscriber(uuid: string, topic: string) { ... }
setLimitCheckSubscriber = (uuid: string, topic: string) => { ... }
```
There are different opinions online about the 2 approaches but in RT codebase there seems to be a massive preference for the arrow function syntax (and it is less verbose). No showstopper but just for the sake of consistency of the code :)

src/client/src/shell/openFin/openFin.ts
```diff
@@ -56,6 +59,12 @@ export class OpenFinLimitChecker {
     fin.desktop.main(() => {
       fin.desktop.InterApplicationBus.addSubscribeListener(this.setLimitCheckSubscriber)
       fin.desktop.InterApplicationBus.addUnsubscribeListener(this.removeLimitCheckSubscriber)
+      fin.desktop.InterApplicationBus.subscribe(LIMIT_CHECKER_UUID, null, LIMIT_CHECKER_STATUS_TOPIC, (m, u, n) => {
```
spcfran: please can you rename the parameters `(m, u, n)` with meaningful names?

Also if you don't need `u` or `n` it is conventional to replace with `_` to indicate you ignore them ([see this](https://stackoverflow.com/questions/41085189/using-underscore-variable-with-arrow-functions-in-es6-typescript)) or in this case, since they are last in the list of parameters, just omit them directly on the handler:
```typescript
fin.desktop.InterApplicationBus.subscribe(LIMIT_CHECKER_UUID, null, LIMIT_CHECKER_STATUS_TOPIC, message => { ... }
```

src/client/src/shell/openFin/openFin.ts
snony: I tried that, and it did not work.

PR #1013, opened 2019-03-26
added possibility to launch limitChecker from openfin without installer

snony:


spcfran:
Please also don't forget to:
- Add the `appAssets` entry for `launcher-dev.json` and `launcher-demo.json`, pointing to their respective URLs (web-dev / web-demo)
- Delete the unnecessary fields from the `Limit Checker` entry in `applicationConfigurations.tsx` (just the url I guess?) to avoid confusion 
- Something to discuss: Should the Limit Checker only open from the launcher, or should we add the option to launch it from within the Reactive Trader interface (like we have with the excel button)? If so, also add the appAsset to the other .json manifest files

snony:
For the second point: the url is needed when we visit the launcher from a webbrowser and the user clicks on the limit checker icon, it is meant to download the zip file.

spcfran:
> @spcfran for the second point: the url is needed when we visit the launcher from a webbrowser and the user clicks on the limit checker icon, it is meant to download the zip file.

OK in that case we probably need to amend the link, as it is currently to a very old version of the limitchecker ([see this](https://github.com/AdaptiveConsulting/ReactiveTraderCloud/tree/gh-pages/install/LimitChecker)). 
I think we should probably 
1- put the unzipped version of the limitchecker on the `public/plugin` folder, and 
2- Reference the local/dev/demo environment accordingly from `applicationConfigurations.tsx` (probably just by using `${location.host}` as in the rest of the appConfigs on that file)

snony:
> I think we should probably
> 1- put the unzipped version of the limitchecker on the `public/plugin` folder, and
----------------------------------------------------------------------------------------------------------------
I think the zipped version is enough cause if someone opens it, they will see the .exe file and can install from there. Plus, it there won't be more files in the plugin folder.
For point 2, I have already done that.

PR #1017, opened 2019-03-28
ARTP-524 / Ability to submit RFQ

gablabelle:
- [x] RFQ feature
- [x] NotionalInput improvements

Screen Shot 2019-04-03 at 10 25 42 AM

PR #1023, opened 2019-04-01
ARTP-532 / Default amount NZDUSD to 10,000,000

gablabelle:
This PR is based on the following PR: https://github.com/AdaptiveConsulting/ReactiveTraderCloud/pull/1017


snony:
@gablabelle , it would be great to know the differences between this PR and the one it is derived from PR (#1017 ) .

gablabelle:
> @gablabelle , it would be great to know the differences between this PR and the one it is derived from PR (#1017 ) .

@snony this PR is based on the RFQ PR #1017 as described in description ... When we will merge #1017 ... this one will become much more smaller. We just have to merge the other first ;-)

PR #1024, opened 2019-04-02
ARTP - 547 : Initial prices on load are 0

inoabrian:
Added check in component's template to conditionally display the button disabled and with a dash '-' instead of the default value 0

### DIFF
--- a/src/client/src/ui/spotTile/components/PriceButton.tsx
+++ b/src/client/src/ui/spotTile/components/PriceButton.tsx
@@ -92,6 +92,9 @@ const getBigFigureDisplay = (bigFigure: number, rawRate: number) =>
 const renderBigFigureDisplay = (bigFigureDisplay: string) =>
   bigFigureDisplay.toString().length === 3 ? `${bigFigureDisplay}0` : bigFigureDisplay
 
+const displayDashAsDefaultPrice = (big: number, pip: number, tenth: number, raw: number) =>
+  big === pip && pip === tenth && tenth === raw && raw === 0
+
 const PriceButtonComp: React.FC<PriceButtonProps> = ({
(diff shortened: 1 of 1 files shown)


src/client/src/ui/spotTile/components/PriceButton.tsx
```diff
@@ -102,18 +105,28 @@ const PriceButtonComp: React.FC<PriceButtonProps> = ({
...
-          <Big>{renderBigFigureDisplay(bigFigure)}</Big>
-        </BigWrapper>
-        <Pip>{renderPips(pip)}</Pip>
-        <Tenth>{tenth}</Tenth>
-      </Price>
-    </TradeButton>
-  )
+  if (!displayDashAsDefaultPrice(big, pip, tenth, rawRate)) {
+    return (
+      <TradeButton direction={direction} onClick={() => handleClick()} disabled={disabled}>
```
bhavesh-desai-scratch: This needs to be kept DRY. This seems like a very tactical fix. I think you need to look at a more holistic solution.

bhavesh-desai-scratch:
```  {rawRate ? (
        <Price>
          <BigWrapper>
            <DirectionLabel>{direction.toUpperCase()}</DirectionLabel>
            <Big>{renderBigFigureDisplay(bigFigure)}</Big>
          </BigWrapper>
          <Pip>{renderPips(pip)}</Pip>
          <Tenth>{tenth}</Tenth>
        </Price>
      ) : (
        <Price>
          <Pip>-</Pip>
        </Price>
      )}```

taylor-adaptive:
Closing. This is addressed in a newer PR.

PR #1028, opened 2019-04-04
Excel legacy plugin

spcfran:
- Added a new excel Adapter for the old .NET plugin
  - Fixed some epics along the way to make it work
- Created a feature switch to choose either adapter (at development time)
- Installation instructions page for legacy plugin. Excel launcher button displays that page.
- Included a local copy of `ReactiveTraderExcel.zip`
- `README`s and comments throughout


src/client/public/static/excel/instructions.html
```diff
@@ -0,0 +1,64 @@
...
+          >ReactiveTraderExcel</a
+        >
+        you need to follow these steps.
+      </p>
+      <p>
+        Download the the zip file
+        <a href="../../plugin/ReactiveTraderExcel.zip">from here</a>.
+      </p>
+      <p>Don't extract the zip file yet.</p>
+      <p>Right click, select properties, check "Unblock" and press "ok".</p>
```
snony: Could you please specify which property of the file we have to check "Unblock"?

spcfran: To be honest this html file is an exact replica of [`docs/setup/reactive-trader-excel.md`](https://github.com/AdaptiveConsulting/ReactiveTraderCloud/blob/develop/docs/setup/reactive-trader-excel.md) so haven't paid much attention to the actual contents! I managed to follow it to make it work though.

The 'unblock' it refers is this one on the properties dialog:

image

I think it may not be always visible, maybe only when downloading from the internet. Ralph couldn't see the checkbox after downloading the .zip file when running on localhost (as the .zip is downloaded from localhost as well)

PR #1038, opened 2019-04-09
The takeUntil on rfqRequestEpic also needs to be filtered.

gablabelle:
The takeUntil on rfqRequestEpic also needs to be filtered. Or else it cancels all, and other Tiles in `requested` state will never received the prices.

This PR fixes that.

### DIFF
--- a/src/client/src/ui/spotTile/epics/rfqTileEpics.ts
+++ b/src/client/src/ui/spotTile/epics/rfqTileEpics.ts
@@ -54,14 +54,25 @@ const fetchRfqQuote = (payload: RfqRequest) => ({
 export const rfqRequestEpic: ApplicationEpic = (action$, state$) =>
   action$.pipe(
     ofType<Action, RfqRequestActionType>(TILE_ACTION_TYPES.RFQ_REQUEST),
-    mergeMap(action =>
+    mergeMap(action => {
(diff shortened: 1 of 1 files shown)


src/client/src/ui/spotTile/epics/rfqTileEpics.ts
```diff
@@ -54,14 +54,25 @@ const fetchRfqQuote = (payload: RfqRequest) => ({
...
-    mergeMap(action =>
+    mergeMap(action => {
+      const cancel$ = action$.pipe(
+        ofType<Action, RfqCancelActionType>(TILE_ACTION_TYPES.RFQ_CANCEL),
+        filter(
+          cancelAction =>
+            cancelAction.payload.currencyPair.symbol === action.payload.currencyPair.symbol,
+        ),
+      )
       // TODO Subscribe to Pricing service instead of passing the current price
```
snony: If the TODO has been resolved, it is best to remove it.

gablabelle: Not done yet, at one point you will need to do that because @spcfran and I didn't understand that the RFQ prices must continue to tick which it doesn't right now.

So in the future you will need to subscribe to pricing service to do so and refactor that epic a bit.

PR #1039, opened 2019-04-10
Artp 598/feat added react hooks

mattydono:
- feat: Updated BlotterHeader to use react hooks.

- fix: removed `type={themeName || 'primary'}` from IconButton.
   - type for button should be submit || button || reset, so it wouldn't compile

### DIFF
--- a/src/client/src/shell/components/app-header/Header.tsx
+++ b/src/client/src/shell/components/app-header/Header.tsx
@@ -30,7 +30,7 @@ class Header extends React.Component {
 const ThemeControl = () => {
   const { themeName, toggleTheme } = useTheme()
   return (
-    <IconButton onClick={toggleTheme} type={themeName || 'primary'}>
+    <IconButton onClick={toggleTheme}>
       <i className={`fa${themeName === ThemeName.Light ? 'r' : 's'} fa-lightbulb`} />
     </IconButton>
   )
@@ -82,7 +82,8 @@ const IconButton = styled.button`
 
   cursor: pointer;
 
-  transition: background-color ${({ theme }) => theme.motion.duration}ms ${({ theme }) => theme.motion.easing};
+  transition: background-color ${({ theme }) => theme.motion.duration}ms
+    ${({ theme }) => theme.motion.easing};
 
   &:hover {
     background-color: ${({ theme }) => theme.button.secondary.active.backgroundColor};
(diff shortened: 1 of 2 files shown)


src/client/src/ui/blotter/components/BlotterHeader.tsx
```diff
@@ -48,66 +44,57 @@ const Fill = styled.div`
...
+  const [quickFilterText, setQuickFilterText] = useState('')
 
-  private quickFilterChangeHandler = (event: React.FormEvent<HTMLInputElement>) => {
-    const { gridApi } = this.props
-    const { value } = event.currentTarget
-    this.setState({ quickFilterText: value })
-    return gridApi && gridApi.setQuickFilter(value)
-  }
+  const quickFilterChangeHandler = useCallback(
+    event => {
```
Sharakai: minor: Would be nice to keep this typed. Either like `useCallback<React.FormEventHandler<HTMLInputElement>>` or `(event: React.FormEvent<HTMLInputElement>) => ...`

src/client/src/ui/blotter/components/BlotterHeader.tsx
```diff
@@ -48,66 +44,57 @@ const Fill = styled.div`
...
+      setQuickFilterText(event.currentTarget.value)
+      return gridApi && gridApi.setQuickFilter(event.currentTarget.value)
+    },
+    [setQuickFilterText, gridApi],
```
Sharakai: I still wonder if `gridApi && gridApi.setQuickFilter` should be included in the deps, but that would only occur if `gridApi` was the same object, but the `setQuickFilter` changed, which would be a horrible API to work with.

Leave it as-is - we can investigate later if this was to ever cause an issue.

src/client/src/shell/components/app-header/Header.tsx
```diff
@@ -30,7 +30,7 @@ class Header extends React.Component {
 const ThemeControl = () => {
   const { themeName, toggleTheme } = useTheme()
   return (
-    <IconButton onClick={toggleTheme} type={themeName || 'primary'}>
+    <IconButton onClick={toggleTheme}>
```
Sharakai: Good fix. Thank you!

src/client/src/ui/blotter/components/BlotterHeader.tsx
taylor-adaptive: React guarantees `setState`  function's identity and it won't change on re-renders, so it can be omitted from the dependency list. 
However `gridApi` should stay .

PR #1040, opened 2019-04-10
ARTP-599 - null checks order flow

james-gray-adaptive:
Contains null check fixes in preparation for turning on --strictNullChecks setting

### DIFF
--- a/src/client/src/routes/NotificationRoute.tsx
+++ b/src/client/src/routes/NotificationRoute.tsx
@@ -36,11 +36,8 @@ export class NotificationRoute extends React.Component<{}, State> {
     fin.desktop.InterApplicationBus.publish(InteropTopics.HighlightBlotter, this.state.message)
 
   render() {
-    let { message } = this.state
-    if (!message) {
-      message = { tradeNotification: undefined }
-    }
-    return (
+    const { message } = this.state
+    return !!message ? (
       <ThemeProvider>
         <TradeNotification
           trade={message.tradeNotification}
@@ -48,7 +45,7 @@ export class NotificationRoute extends React.Component<{}, State> {
           highlightTradeInBlotter={this.highlightTradeInBlotter}
         />
       </ThemeProvider>
-    )
+    ) : null
   }
 }
 
--- a/src/client/src/routes/OrderTicketRoute/MediaPlayer.tsx
+++ b/src/client/src/routes/OrderTicketRoute/MediaPlayer.tsx
@@ -2,10 +2,10 @@ import * as React from 'react'
 
 export interface Props {
   src?: Blob | string
-  at?: number
+  at: number
+  rate: number
   play?: boolean
   loop?: boolean
-  rate?: number
   context: AudioContext
   output?: AudioNode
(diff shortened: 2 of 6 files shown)


src/client/src/routes/NotificationRoute.tsx
```diff
@@ -36,19 +36,16 @@ export class NotificationRoute extends React.Component<{}, State> {
...
+    return !!message ? (
       <ThemeProvider>
         <TradeNotification
           trade={message.tradeNotification}
           dismissNotification={this.onDismissNotification}
           highlightTradeInBlotter={this.highlightTradeInBlotter}
         />
       </ThemeProvider>
-    )
+    ) : null
```
Sharakai: Wouldn't this previously always render `<TradeNotification />` but `message.tradeNotification` would be undefined? Or was that actually an erroneous scenario?

src/client/src/routes/OrderTicketRoute/MediaPlayer.tsx
```diff
@@ -18,7 +18,7 @@ export interface State {
     position: number
     pausedAt: number
   }
-  source?: AudioBufferSourceNode
+  source?: AudioBufferSourceNode | null
```
Sharakai: `AudioBufferSourceNode | null | undefined`? 😥 Any way to unify the `null | undefined`, or are they used for different scenarios?

src/client/src/routes/OrderTicketRoute/VoiceInput.tsx
```diff
@@ -80,7 +80,10 @@ export class VoiceInput extends React.PureComponent<Props, State> {
     sessionInstance: null,
   }
 
-  static getDerivedStateFromProps({ context, source, requestSession }: Props, state: State): Partial<State> {
+  static getDerivedStateFromProps(
+    { context, source, requestSession }: Props,
+    state: State,
+  ): Partial<State> {
     let next: any = null
     const { analyser, blob, destination }: any = state
```
Sharakai: Not your PR, but why does this destructuring have an `any` type 🤔

src/client/src/routes/NotificationRoute.tsx
Sharakai: Ah I see now. The `TradeNotification` without a `trade` (or with the `defaultProps.trade` is.... Odd.

james-gray-adaptive: Yeah, doesn't make sense to have a notification when all the details are not present.

src/client/src/routes/OrderTicketRoute/MediaPlayer.tsx
james-gray-adaptive: Good shout. Removing the undefined. Thanks :)

src/client/src/routes/OrderTicketRoute/VoiceInput.tsx
james-gray-adaptive: Added the correct type

PR #1045, opened 2019-04-22
change finsemble platform check

pdeleastar:


src/client/src/rt-components/platform/platform.ts
```diff
@@ -1,6 +1,12 @@
 import { Browser, Finsemble, OpenFin, PlatformAdapter } from './adapters'
 
-const isFinsemble = typeof window.FSBL !== 'undefined'
+/*
```
bhavesh-desai-scratch: Please elaborate here

PR #1048, opened 2019-05-14
ARTP-619 - Firefox scrollbars

taylor-adaptive:
Added scrollbar styles for firefox

**Before**
ARTP-619-1

**After**
ARTP-619-2

### DIFF
--- a/src/client/src/shell/GlobalScrollbarStyle.tsx
+++ b/src/client/src/shell/GlobalScrollbarStyle.tsx
@@ -10,6 +10,9 @@ export const css = memoize(
     body ::-webkit-scrollbar-thumb {
       background-color: ${color};
     }
+    body {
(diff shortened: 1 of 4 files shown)


src/client/src/shell/GlobalScrollbarStyle.tsx
```diff
@@ -10,6 +10,9 @@ export const css = memoize(
     body ::-webkit-scrollbar-thumb {
       background-color: ${color};
     }
+    body {
+      scrollbar-color: ${color} transparent;
```
taylor-adaptive: Adding `scrollbar-width` to `body` has no affect, so I had to add these specifically to our scrollable panels. I left colour here to avoid recreating the `rgba` value in each file.

PR #1064, opened 2019-05-23
ARTP-635 - Clicking the OpenFin logo, bottom right, should go straight to the OpenFin website

taylor-adaptive:
- Added `OpenFinBrowserLink` component to open `<a>` in a new browser window from OpenFin
- Link the OpenFin logo to OpenFin's website

### DIFF
--- a/src/client/src/rt-components/open-fin/OpenFinBrowserLink.tsx
+++ b/src/client/src/rt-components/open-fin/OpenFinBrowserLink.tsx
@@ -1,0 +1,19 @@
+import React, { FC, MouseEvent, ReactNode } from 'react'
+
+interface Props {
+  href: string
+  children: ReactNode
+  className?: string
+}
+
+const OpenFinBrowserLink: FC<Props> = props => (
+  <a
+    {...props}
+    onClick={(e: MouseEvent<HTMLAnchorElement>) => {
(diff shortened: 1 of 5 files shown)


src/client/src/rt-components/open-fin/OpenFinBrowserLink.tsx
```diff
@@ -0,0 +1,19 @@
...
+  children: ReactNode
+  className?: string
+}
+
+const OpenFinBrowserLink: FC<Props> = props => (
+  <a
+    {...props}
+    onClick={(e: MouseEvent<HTMLAnchorElement>) => {
+      e.preventDefault()
+      fin.System.openUrlWithBrowser(e.currentTarget.href)
```
taylor-adaptive: Adding `href` attribute to the anchor isn't necessarily needed, but I chose to use it to get the native anchor tag behaviours for free.

PawelAP: the method handling onClick should be placed outside of the function i.e.
onClick={handleClick}

taylor-adaptive: Good spot, addressed in a580156

PR #1069, opened 2019-05-28
Openfin/notifications

marion35700:
Use openFin notifications service instead of openFin native notifications. 
OpenFin Notifications service comes with a 'notifications center' that displays the notifications. It is available for the standalone Openfin app version and OpenFin Cloud version.
The rendering of the notification itself cannot be overridden (by design according to OpenFin). I used the 'button' feature available in the notification to keep the 'highlight in blotter' feature.
Currently the template for the toast &  the notification in the center is slightly different (no buttons in the toast) but it will be the same template for both in the next release of the service. 
There is currently an issue with the sorting of the notifications in the panel, OpenFin support team told me it should be fixed in the next release (expected by the end of the week)


src/client/src/rt-components/platform/adapters/openfin/openFin.ts
```diff
@@ -4,25 +4,35 @@ import { openDesktopWindow } from './window'
...
   readonly name = 'openfin'
   readonly type = 'desktop'
 
+  private createNotification: (options: NotificationOptions) => Promise<Notification>
+
+  constructor() {
+    super()
+    this.createNotification = require('openfin-notifications').create
+
+    require('openfin-notifications').addEventListener(
```
taylor-adaptive: Is there a reason for the inline `require` rather than importing at the top of the file?

src/client/src/rt-components/platform/adapters/openfin/openFin.ts
```diff
@@ -108,11 +118,35 @@ export default class OpenFin extends BasePlatformAdapter {
...
   notification = {
-    notify: (message: object) =>
-      new fin.desktop.Notification({
-        url: '/notification',
-        message,
-        timeout: 8000,
-      } as any),
+    notify: (message: object) => {
+      this.createNotification(
+        Object.assign({
```
taylor-adaptive: Is there any need for `Object.assign` here?

src/client/src/rt-components/platform/adapters/openfin/openFin.ts
```diff
@@ -108,11 +118,35 @@ export default class OpenFin extends BasePlatformAdapter {
...
+        })
+        .catch(err => {
+          console.error('Notification error', err)
+        })
+    },
+  }
+
+  getNotificationTitle({ tradeNotification }: NotificationMessage) {
+    const status = tradeNotification.status === 'done' ? 'Accepted' : 'Rejected'
+    return `Trade ${status}: ${tradeNotification.direction} ${tradeNotification.dealtCurrency} ${
```
taylor-adaptive: I'm not sure this is the best place for the Notification content to be generated, what if we want to support different types of notifications? Is there a way we can move this out of the Platform adapter?

src/client/src/rt-components/platform/adapters/openfin/openFin.ts
marion35700: Yes, we need the inline require, otherwise the browser version is not working. Fyi, I have this error
`ReferenceError: fin is not defined
(anonymous function)
C:/git/ReactiveTraderCloud/src/client/node_modules/openfin-notifications/dist/client/main-bundle.js:535
  532 | 
  533 | exports.eventEmitter = new events_1.EventEmitter();
  534 | 
> 535 | if (fin.Window.me.uuid !== internal_1.SERVICE_IDENTITY.uuid || fin.Window.me.name !== internal_1.SERVICE_IDENTITY.name) {
      | ^  536 |   exports.channelPromise = typeof fin === 'undefined' ? Promise.reject(new Error('fin is not defined. The openfin-notifications module is only intended for use in an OpenFin application.')) : fin.InterApplicationBus.Channel.connect(internal_1.SERVICE_CHANNEL, {
  537 |     payload: {
  538 |       version: '0.10.1'
View compiled
__webpack_require__
C:/git/ReactiveTraderCloud/src/client/node_modules/openfin-notifications/dist/client/main-bundle.js:62`

However, I improved a bit the code to at least not repeat the require

src/client/src/rt-components/platform/adapters/openfin/openFin.ts
marion35700: no, I removed it

src/client/src/rt-components/platform/adapters/openfin/openFin.ts
marion35700: We already have the 'NotificationMessage' which is the notification across platform. We need this code here because it is specific to OpenFin. Each platform has their own constraints/format... When we have other types of notification to handle, then we can maybe refactor the code...

PR #1074, opened 2019-06-07
add mac installers

bhavesh-desai-scratch:


garyadaptive:
Delete .DS_Store and add to .gitignore

PR #1087, opened 2019-06-25
ARTP-667: Blotter Highlight Logic Revisit

pmilla1606:
After feedback from @stewart-adaptive and @bhavesh-desai-scratch this is a refactor of #1084.

We are no longer firing off actions with empty arrays in them, instead we filter out all entries that do not have trades; that is, an old `tradeId` that comes from the Notifications pane and no longer matches the `tradeId`'s in the blotter. 

Doing this avoids code duplication and allows us to simplify the `removeHighlightTradeEpic` - it will never be called with a missing trade and therefore doesn't need any logic.

I'm curious to hear your feedback on this @stewart-adaptive and @bhavesh-desai-scratch. Let me know if you'd like to see any changes.


pmilla1606:
@stewart-adaptive thanks for the commit - looks good!

PR #1089, opened 2019-07-02
feat[fdc3]: broadcast currency pair from blotter

limerentfox:
This PR is the first step to using fdc3 context. Here we broadcast the currency pair when the user clicks on a row in the blotter. Separately in Adaptive Insights, there will be a PR for that app to receive the RT context.

### DIFF
--- a/src/client/src/ui/blotter/components/Blotter.tsx
+++ b/src/client/src/ui/blotter/components/Blotter.tsx
@@ -1,5 +1,5 @@
-import { GridApi } from 'ag-grid'
 import { AgGridReact } from 'ag-grid-react'
+import { GridApi } from 'ag-grid'
 // tslint:disable-next-line:no-submodule-imports
 import 'ag-grid/dist/styles/ag-grid.css'
 import React, { useState, useCallback } from 'react'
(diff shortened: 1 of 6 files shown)


src/client/src/ui/blotter/components/Blotter.tsx
```diff
@@ -8,6 +8,24 @@ import { Trade, TradeStatus } from 'rt-types'
...
+  }
+  const currencyPairContext: Context = {
+    type: 'fdc.instrument',
+    name: symbol,
+    market: 'currency',
+    id: {
+      id: symbol,
+    },
+  }
+  const fdc3 = require('openfin-fdc3')
```
limerentfox: Here I use require because `openfin-fdc3` is not compatible with web. Importing fdc3 results in errors where the site says it doesn't know what `fin` is. Imports are hoisted and at the time of process led to the error.

src/client/src/ui/blotter/components/Blotter.tsx
```diff
@@ -83,11 +67,19 @@ const Blotter: React.FC<BlotterProps> = props => {
...
   const onModelUpdated = useCallback(
     () => gridApi && setDisplayedRows(gridApi.getDisplayedRowCount()),
     [gridApi],
   )
 
+  const broadcastContext = (context: Context) => {
+    if (platform.name !== 'openfin') {
+      return
+    }
+    platform.fdc3.broadcast(context)
```
stewart-adaptive: can you change to 
const broadcastContext = (context: Context) => platform.broadcast(context)

then have the non openfin adapters do nothing on broadcast  so your not checking platform name in the component

PR #1090, opened 2019-07-03
Only allow drag tear off on tiles

stewart-adaptive:
new boolean to set dragable functionality


stewart-adaptive:
> This does not change the draggable area of the component, correct? Approving but maybe we should create a ticket to restrict the draggable area as per Fergus' comment.

I discussed with Fergus, changing the draggable area is a bigger change, will do later under a different ticket

PR #1111, opened 2019-08-06
Add NLP to Reactive Trader 🦄

bhavesh-desai-scratch:


src/server/Adaptive.ReactiveTrader.Server.Bot/src/nlp-services.ts
```diff
@@ -0,0 +1,54 @@
+import DialogueFlow from 'dialogflow';
+import { from, Observable } from 'rxjs';
+import { map, mergeMap } from 'rxjs/operators';
+import { SymphonyClient } from 'symphony';
+import { Message } from 'symphony-api-client-node';
+import uuid from 'uuid';
+
+interface DFResultWithOriginalMessage {
```
rikoe: Shall we just call this "DialogFlowResult"? People may not understand what DFResult means...

PR #1117, opened 2019-08-13
Bugfix/artp 676

my-code-is-lit:


src/client/src/apps/MainRoute/widgets/spotTile/components/Tile/TileBusinessLogic.ts
```diff
@@ -209,6 +210,7 @@ export const resetNotional = ({
   currencyPair,
 }: DerivedStateFromNotionalReset): TileState => {
   const notional = getDefaultNotionalValue(currencyPair)
+  if (window.DEFAULT_NOTIONAL !== undefined) window.DEFAULT_NOTIONAL = notional;
```
fdore: Why do we store it on the window object? Could we not save it in the redux store?

my-code-is-lit: @fdore because when the tile is "torn off" it's a separate window instance and so the separate window redux-store will be a different scope. It's the cleanest way parent/child windows can communicate

fdore: Child windows actually run in the same process as their parent window. The prices in the opened spot tile are coming from that same redux store, so I'm not sure why the notional couldn't be... 🤔

my-code-is-lit: @fdore let me see if I can check console and find a better solution.

my-code-is-lit: > Child windows actually run in the same process as their parent window. The prices in the opened spot tile are coming from that same redux store, so I'm not sure why the notional couldn't be... 🤔

@fdore Just realized, the value of the input component doesn't go in the store
So I threw it in window as a band-aid. Let me know if this works for you. If need be, it can be refactored to be added to redux.

PR #1125, opened 2019-09-02
Added two sample market adapters that set the startup values of the c…

Ian887:
…urrency pairs to current market rates.

Changed the timer functionality to re-sync the currency pairs on startup and then once every 24 hours.
Moved the Observables from the anonymous method in the PriceGenerator constructor down into the new BaseWalkPriceGenerator class


src/server/Adaptive.ReactiveTrader.Server.Pricing/Adaptive.ReactiveTrader.Server.Pricing.csproj
```diff
@@ -18,6 +18,7 @@
   </ItemGroup>
 
   <ItemGroup>
+    <PackageReference Include="HtmlAgilityPack" Version="1.11.12" />
```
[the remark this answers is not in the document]
jonclare: Agree with this. If we have price sources that are through an API, why do we need this at all?

src/server/Adaptive.ReactiveTrader.Server.IntegrationTests/TestAddress.cs
```diff
@@ -2,6 +2,6 @@ namespace Adaptive.ReactiveTrader.Server.IntegrationTests
 {
     public static class TestAddress
     {
-        public const string Broker = "ws://broker:8000/ws";
+        public const string Broker = "ws://127.0.0.1:8000/ws";
```
jonclare: Did you mean to check this in? Will this work on a deployed environment?

src/server/Adaptive.ReactiveTrader.Server.Pricing/PriceSourceAdapters/FixerAdapter.cs
```diff
@@ -0,0 +1,92 @@
...
+{
+  /// <summary>
+  /// Scrapes currency pair data from a website called data.fixer.io and returns the currency pairs
+  /// Note https://fixer.io is only free if there are less than a restricted number of API calls per month.
+  /// </summary>
+  public class FixerAdapter : FixerBaseAdapter, IMarketDataAdapter
+  {
+    private const string requestUriString = "https://data.fixer.io/api/latest";
+    private const string sourceName = "Fixer";
+    private const string API_Key = "APIKEY";
```
jonclare: One thing to be aware of here, when we introduce a proper API key is not to check it in.
We should be populating that through config. We don't want people who download the code and run it to use up our request allocation.
And we should provide a suitable fallback (and log), whether through the hardcoded values or whatever if someone tries to run things without a valid API key

src/server/Adaptive.ReactiveTrader.Server.Pricing/PriceSourceAdapters/FixerAdapter.cs
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    /// <summary>
+    /// Not currently used because we do not have an API key yet, and the European Central Bank is free!
+    /// </summary>
+    /// <returns></returns>
+    public override async Task<IEnumerable<MarketData>> GetMarketData()
+    {
+      /*
+       The data is returned in this format for each base currency:
```
jonclare: Maybe this could be an XML comment on the method?

src/server/Adaptive.ReactiveTrader.Server.Pricing/PriceSourceAdapters/EuropeanCentralBankAdapter.cs
```diff
@@ -0,0 +1,81 @@
...
+      return result.Values;
+    }
+
+    private async Task<JArray> GetJson()
+    {
+      var results = new List<JObject>();
+      foreach (var baseCcy in _currencies)
+      {
+        //We do multiple requests, one for each base ccy
+        var symbols = string.Join(',', _currencies.Where(x => x != baseCcy));
```
jonclare: It looks like we're doing requests for all combinations of base/quote ccy here.
This should be based on wherever we're populating the master list of pairs from, is that in the hardcoded PriceSource class?
Maybe we should introduce a proper CurrencyPair domain object here rather than treating things as strings

src/server/Adaptive.ReactiveTrader.Server.Pricing/PriceSourceAdapters/FixerAdapter.cs
```diff
@@ -0,0 +1,92 @@
...
+      }
+      return result.Values;
+    }
+
+    private async Task<JArray> GetJson()
+    {
+      var results = new List<JObject>();
+      foreach (var baseCcy in _currencies)
+      {
+        //We do multiple requests, one for each base ccy
```
jonclare: And for this one, we'll need to be mindful of API rate limits.
The free Fixer API says you get 1000 calls per month.

If we're refreshing once per hour, that's up to 744 refreshes per month. If each refresh calls the API multiple times, one for each currency, we're going to blow our limit here

src/server/Adaptive.ReactiveTrader.Server.Pricing/PriceSourceAdapters/AdapterBase.cs
```diff
@@ -0,0 +1,120 @@
...
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Net.Http;
+using System.Threading.Tasks;
+
+namespace Adaptive.ReactiveTrader.Server.Pricing.PriceSourceAdapters
+{
+  public abstract class AdapterBase
```
jonclare: This feels like it would be better done via composition rather than inheritance.
And I think we can split things up a bit more as well, especially where we have things for parsing HTML documents, which are clearly never going to be used by API.
This looks like it's a bucket of things that are useful across various adapters rather than being a true base class.

adapters.

src/server/Adaptive.ReactiveTrader.Server.Pricing/PriceSource.cs
```diff
@@ -84,21 +79,48 @@ private static IPriceGenerator CreatePriceGenerator(string symbol, decimal initi
       return new MeanReversionRandomWalkPriceGenerator(symbol, initial, precision);
     }
 
-    private IObservable<Unit> RegisterPriceTrigger(string symbol)
+    private void RefreshMarketRates(DateTime refreshDateTime)
+    {
+      foreach (var adapter in _marketAdapters)
+      {
+        var marketData = new List<MarketData>(adapter.GetMarketData().Result);
```
jonclare: We shouldn't be using the Result of a task directly

jonclare: Also, we need to handle the case if a market data source is down/the adapter fails for whatever reason.
We should probably fall back to the hard coded values in that case

src/server/Adaptive.ReactiveTrader.Server.Pricing/Adaptive.ReactiveTrader.Server.Pricing.csproj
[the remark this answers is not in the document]
Ian887: This is only the fourth fall-back adapter..
First: EuropeanCentralBankAdapter (json API)
Second: FinancialModelingPrepAdapter (json API)
Third: Fixer.IO (json API not enabled yet because we do not have an API key yet)
Fourth: Yahoo.com (html scrape)

src/server/Adaptive.ReactiveTrader.Server.Pricing/BaseWalkPriceGenerator.cs
```diff
@@ -0,0 +1,58 @@
...
+    protected decimal _previousMid;
+
+    public BaseWalkPriceGenerator(string symbol, decimal initial, int precision)
+    {
+      Symbol = symbol;
+      _initial = _previousMid = initial;
+      _precision = precision;
+      EffectiveDate = new DateTime(2019, 1, 1);
+      SourceName = PriceSource.HardCodedSourceName;
+      //PriceChanges = _priceChanges.Replay(1).RefCount();
```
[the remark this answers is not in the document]
Ian887: done

src/server/Adaptive.ReactiveTrader.Server.IntegrationTests/TestAddress.cs
Ian887: Yes. All the server  Integration tests both locally (and on the build server) are broken.
This was the original value by Qiming and I changed it to this to get the tests running locally.
It was changed to "broker" on 17 Jan by Keshai which broke the tests.

I want to see if this fixes the build server tests. If it does then running on deployed environments obviously needs more work.

src/server/Adaptive.ReactiveTrader.Server.Pricing/PriceSourceAdapters/FixerAdapter.cs
Ian887: Agreed.

src/server/Adaptive.ReactiveTrader.Server.Pricing/PriceSourceAdapters/FixerAdapter.cs
Ian887: Agreed. The original specification from Loic to update the rates once a day. When I demoed RT yesterday to Mike and Amanda they asked if I could change the update frequency to every hour so I did. This is now running on all 3 of the free EuropeanCentral, FinancialModelingPrep, and the Yahoo adapters.

The Fixer.IO adapter is currently disabled and waiting for an API key. If we decide to go with Fixer.IO later (maybe for real time push) then we will pay the money and it will not be polled.

src/server/Adaptive.ReactiveTrader.Server.Pricing/PriceSource.cs
Ian887: 1. Done.
2. Agreed. The PriceSource class initializes itself with all the required currency pairs, with the same hard-coded defaults as before. Then every hour, the 3 adapters are called (primary one first). This means if the EuropeanCentralBank, FinancialModellingPrep, and Yahoo are all down then the defaults are used.

src/server/Adaptive.ReactiveTrader.Server.Pricing/PriceSourceAdapters/EuropeanCentralBankAdapter.cs
Ian887: Good idea. Done.

src/server/Adaptive.ReactiveTrader.Server.IntegrationTests/TestAddress.cs
Ian887: I have changed it back to "broker". "localhost" allows the tests to run locally in VS2019 if I have already started the Service.
Someone who knows the build server will have to fix this asap.

src/server/Adaptive.ReactiveTrader.Server.Pricing/PriceSourceAdapters/FixerAdapter.cs
Ian887: done

src/server/Adaptive.ReactiveTrader.Server.Pricing/PriceSourceAdapters/AdapterBase.cs
Ian887: done

src/server/Adaptive.ReactiveTrader.Server.Pricing/PriceSourceAdapters/FixerAdapter.cs
Ian887: just spoke to Loic, we have decided to remove Fixer.io for now and just rely on the other 3 adapters.

src/server/Adaptive.ReactiveTrader.Server.Pricing/Adaptive.ReactiveTrader.Server.Pricing.csproj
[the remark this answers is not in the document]
Ian887: Need to discuss. Is RT a real time trading platform to demonstrate to potential clients (Amanda), or is RT a trading platform demo showcasing reactive programming principles applied across the full application stack (the root readme.md) . IMHO web scrapes are nasty but they are a necessary evil in the real world.

src/server/Adaptive.ReactiveTrader.Server.Pricing/PriceSource.cs
```diff
@@ -1,92 +1,120 @@
...
+      return new MeanReversionRandomWalkPriceGenerator(new CurrencyPair(baseCcy, quoteCcy), initial, precision);
     }
 
-    private void RefreshMarketRates(DateTime refreshDateTime)
+    private async Task RefreshMarketRates(DateTime refreshDateTime)
     {
       foreach (var adapter in _marketAdapters)
       {
-        var marketData = new List<MarketData>(adapter.GetMarketData().Result);
+        var marketData = new List<MarketData>(await adapter.GetMarketData());
```
jonclare: Is this going to do what we want if this GetMarketData method throws?
I'd have thought in that case we'd want to fall back to one of the other providers?

Ian887: Ok, I was a bit lazy and assumed that all adapters, now and in the future would trap their own exceptions. I have added a lovely try{} catch{} here to protect future adapters.

PR #1132, opened 2019-09-05
Refactor/dynamic import

my-code-is-lit:


fdore:
We can incorporate the async imports once #1153 is merged in

bhavesh-desai-scratch:
#1153 is now merged. This takes care of Async platform initialization but not the lazy loading of the modules themselves. @fdore @BenGladman will sort this out.

PR #1144, opened 2019-09-17
Fix trade date render

my-code-is-lit:


fdore:
Should be handled server side. UI should't have to do these sorts of date manipulation.

rikoe:
@fdore @my-code-is-lit I agree - shall we close this PR without merging then

PR #1147, opened 2019-09-23
Refactor platform abstraction

fdore:
This PR is a first stab at simplifying the platform abstraction, and making it less bloated.

First, here is below a summary of the current platform interface, and api availability across platforms.

|                 | API              | Browser | Openfin | Finsemble | Symphony | Glue42 |
|-----------------------|------------------|---------|---------|-----------|----------|--------|
| window                | open             | X       | X       | X         | X        | X      |
|                | close            | X       | X       | X         | X        | X      |
|                | maximize         | \.      | X       | \.        | \.       | \.     |
|                | minimize         | \.      | X       | \.        | \.       | \.     |
|                | publishToChannel | \.      | \.      | \.        | \.       | X      |
| notification          | notify           | X       | X       | X         | ?        | X      |
| app                   | open             | \.      | X       | X         | \.       | \.     |
|                   | exit             | \.      | X       | \.        | \.       | \.     |
| interop               | subscribe$       | \.      | X       | X         | \.       | \.     |
|               | publish          | \.      | X       | X         | \.       | \.     |
| fdc3                  | broadcast             | \.      | X       | \.        | \.       | \.     |
| epics                 | epics            | \.      | X       | \.        | \.       | \.     |
| share                 | share            | \.      | \.      | \.        | X        | \.     |
| notificationHighlight | init             | \.      | X       | \.        | \.       | \.     |
| limitChecker          | limitChecker     | \.      | X       | \.        | \.       | \.     |
| excel                 | adapterName      | \.      | X       | \.        | \.       | \.     |
|                 | open             | \.      | X       | \.        | \.       | X      |
|                 | isOpen           | \.      | X       | \.        | \.       | X      |
|                 | publishPositions | \.      | X       | \.        | \.       | X      |
|                 | publishBlotter   | \.      | X       | \.        | \.       | X      |
|               | open             | \.      | X       | X         | \.       | \.     |

Here are a few notes regarding the current api:
1) the notificationHighlight api is really just a proxy for the underlying interop api, so we could make it redundant
2) limitChecker could be moved out of this interface. In practice, limitChecker is an api to a 3rd party app, and can be injected as a normal dependency. The actual implementation will depend on the platform.
3) The name "excel" is kind of misleading, as this is not a generic api, but similar to the limitChecker. This is an api to a 3rd party app, which happens to be based on Excel. Again, we can move it out, and inject it as a normal dependency (and provide different implementations based on the platform)
4) ChartIQ is redundant, as this is a generic api to call an external app


src/client/src/apps/MainRoute/widgets/blotter/epics/blotterHighlightEpic.ts
```diff
@@ -6,6 +6,7 @@ import { combineEpics, ofType } from 'redux-observable'
 import { Action } from 'redux'
 import { Trade } from 'rt-types'
 import { EMPTY } from 'rxjs'
+import { InteropTopics } from '../../../../../rt-platforms'
```
bhavesh-desai-scratch: Auto import gone bad.

PR #1153, opened 2019-09-27
Resolve platform asynchronously

fdore:


rikoe:
@fdore I am busy checking if the RTC Symphony extension will continue to work if this is merged - we need it for the demo on Thursday. Will approve if happy, then we can merge. Looking to fix a bug with the loading of the Symphony platform where it doesn't always capture the Symphony window property correctly.

src/client/src/rt-platforms/platform.ts
```diff
@@ -1,15 +1,15 @@
-import { Browser, Finsemble, OpenFin, PlatformAdapter, Symphony } from './'
+import { Browser, Finsemble, OpenFin, Symphony } from './'
+import { waitForObject } from '../rt-util'
 
 const urlParams = new URLSearchParams(window.location.search)
 
 const isFinsemble = 'FSBL' in window
 const isOpenFin = 'fin' in window
 const isSymphony = urlParams.has('waitFor') && urlParams.get('waitFor') === 'SYMPHONY'
```
rikoe: We need to fix this line, it is a temp workaround @bhavesh-desai-scratch and I did...

src/client/src/rt-platforms/platform.ts
```diff
@@ -1,15 +1,15 @@
...
 
-const getPlatform: () => PlatformAdapter = () => {
-
+export const getPlatformAsync = async () => {
   if (isSymphony) {
-    console.info('Using Symphony')
```
rikoe: 👍

src/client/src/rt-platforms/platform.ts
rikoe: @fdore will these constants actually work? Don't they get resolved before the platform is resolved?
I think that is one of the reasons we modified the isSymphony constant, it sometimes didn't get set correctly.

Would it work if they go inside the `getPlatformAsync` function?

fdore: The only thing getPlatformAsync does is wait for symphony to be registered on the window, but don't you know from the start (.i.e from the launch url) that you are running on symphony? (disclaimer, I haven't tested the symphony use case)

PR #1156, opened 2019-10-01
Glue integration

drodriguez-Tick42:
Glue integration


src/client/src/rt-platforms/glue/glue.ts
```diff
@@ -0,0 +1,96 @@
...
+import { openGlueWindow, registerWindowMethods } from './window'
+import { sendNotification, NotifyPermission } from './utils/sendNotification'
+import { getAddinStatus, openSheet, updateSheet, BlotterData } from './utils/excel'
+import DefaultRoute from 'rt-platforms/defaultRoute'
+import { ApplicationEpic } from 'StoreTypes'
+import { GlueHeader, GlueLogoLink } from './'
+
+// TODO remove onGlueLoaded and its usages when the whole app is rendered after glue has loaded
+export const onGlueLoaded = (callback: () => void) => {
+  const interval = setInterval(() => {
```
fdore: There is a `waitForObject` util you can use instead (which does the same thing)

src/client/src/rt-platforms/glue/glue.ts
```diff
@@ -0,0 +1,96 @@
...
+  constructor() {
+    super()
+    Glue({ channels: true })
+      .then((glue: GlueInterface.Glue) => {
+        window.glue = glue
+        registerWindowMethods()
+        return Glue4Office()
+      })
+      .then((glue4office: Glue42OfficeInterface.Glue4Office) => {
+        window.glue4office = glue4office
```
fdore: this could be done outside of the promise

src/client/src/rt-platforms/glue/utils/excel.ts
```diff
@@ -0,0 +1,99 @@
...
+  return sheet
+}
+
+export const updateSheet = async (blotterData: BlotterData[]) => {
+  if (dataLength === blotterData.length) {
+    // don't update if the number of rows is the same
+    return
+  }
+  if (!sheet) {
+    await openSheet(blotterData)
```
fdore: Should that be `await openSheet(convertBlotterData(blotterData))` instead?

src/client/src/rt-platforms/glue/utils/excel.ts
```diff
@@ -0,0 +1,99 @@
...
+}
+
+/**
+ * Converts the blotter data in human readable format where necessary.
+ * New record is displayed on top.
+ */
+const convertBlotterData = (blotterData: BlotterData[]) =>
+  blotterData
+    .map(data => {
+      data.status = `${data.status.charAt(0).toUpperCase()}${data.status.slice(1)}`
```
fdore: not a fan of mutating that array. Can we map to a new instance instead?

src/client/src/rt-platforms/glue/utils/excel.ts
```diff
@@ -0,0 +1,99 @@
...
+    'Jul',
+    'Aug',
+    'Sep',
+    'Oct',
+    'Nov',
+    'Dec',
+  ]
+  const hours = date.getHours() < 10 ? `0${date.getHours()}` : `${date.getHours()}`
+  const minutes = date.getMinutes() < 10 ? `0${date.getMinutes()}` : `${date.getMinutes()}`
+  const seconds = date.getSeconds() < 10 ? `0${date.getSeconds()}` : `${date.getSeconds()}`
```
fdore: I would create an helper method for that

src/client/src/rt-platforms/glue/window.ts
```diff
@@ -0,0 +1,182 @@
...
+
+  window.glue.interop.register('tabAllWindows', () => {
+    listOfOpenedWindows.forEach(async (wnd, index) => {
+      if (index > 0) {
+        await listOfOpenedWindows[0].attachTab(wnd, index)
+      }
+    })
+  })
+
+  window.window.glue.interop.register('openWorkspace', (args: { symbol: string }) => {
```
fdore: should that be `window.glue.interop` instead?

src/client/src/rt-platforms/glue/glue.ts
drodriguez-Tick42: If my understanding is correct, loading Glue4Office happens after Glue has loaded and this is where we should asynchronously render the application, once the refactoring is done

src/client/src/rt-platforms/glue/glue.ts
drodriguez-Tick42: It seems to have a slightly different signature, which makes it a bit more complex to use in our case inside the useEffect() hook, but this check would be removed when the app is loaded asynchronously inside the Glue().then() anyway. However, we have slightly refactored this and temporarily created a custom hook called useOnGlueLoaded

src/client/src/rt-platforms/glue/utils/excel.ts
drodriguez-Tick42: I believe the conversion was done inside the openSheet function itself, but yes, it can probably be removed from there and this can be changed, given its current usages

drodriguez-Tick42:
Some shapshots are failing, shall we update them?

PR #1159, opened 2019-10-02
Allow PRs from forks to build, suppress docker push for PRs

sergei-ivanov:


.circleci/config.yml
```diff
@@ -5,6 +5,14 @@ jobs:
...
     steps:
       - checkout
+      - run:
+          name: 'Configure the environment'
+          command: |
+            # Inject a dummy user for builds triggered from forks, because secure variables are not available there.
+            # This allows the job to proceed with building and testing docker images.
+            if [[ -n "${CIRCLE_PR_REPONAME}" ]] ; then
+              echo 'export DOCKER_USER=adaptive' >> $BASH_ENV
+            fi
```
rikoe: 👍

.circleci/config.yml
```diff
@@ -77,17 +85,24 @@ jobs:
       - run:
           name: Run client tests
           command: 'docker run -e CI=true ${DOCKER_USER}/client-test${CIRCLE_BUILD_NUM} npm run test -- --coverage'
+      - run:
+          name: 'Skip the next steps for PRs'
+          command: |
+            # Check the environment and gracefully stop the job for PRs
+            if [[ "$CIRCLE_BRANCH" != "master" ]] && [[ "$CIRCLE_BRANCH" != "develop" ]] ; then
+              circleci-agent step halt
+            fi
```
rikoe: Is there no way to make steps conditional, instead of force stopping the build early?

sergei-ivanov: Conditional steps configuration is very limited, and we cannot split the job in two because we need to reuse the local docker cache. I cannot see an easy alternative way of doing it.

PR #1160, opened 2019-10-02
Test:(ARTP 726&727) Protractor ui tests for reactive trader cloud application

PendliKranthi:
* Created test framework with protractor and  typescript and added all the necessary files
https://weareadaptive.atlassian.net/browse/ARTP-726
https://weareadaptive.atlassian.net/browse/ARTP-727


src/client/__tests__/e2e/pages/blotter.component.ts
```diff
@@ -0,0 +1,45 @@
...
+  label: string
+}
+
+export class BlotterComponent {
+  tradesTable: Record<string, Record<string, ElementFinder>>
+
+  constructor(private browser: ProtractorBrowser, public root: ElementFinder) {
+    this.tradesTable = {
+      executedTrades: {
+        tradeID: root.element(by.xpath('//*[@id="root"]/div/div/div[2]/div/div[1]/div[2]/div/div[2]/div/div/div[2]/div/div/div[1]/div/div[3]/div[2]/div/div/div[1]/div[2]')),
```
BenGladman: These locators are quite flakey as they heavily rely on the DOM structure.

It might be worth adding a `data-testid="xxx"` attribute to the elements you want to target in tests and then using `by.css("[id="root"] [data-testid=xxx]")` in the locators

(Also it's worth noting that `by.xpath` is not efficient, `by.css` is much faster because it can use the browser's `querySelectorAll`)

BenGladman:
A couple of general questions if I may :)
- What was the reasoning behind choosing Protractor instead of Selenium or WebdriverIO? I thought Protractor was aimed at Angular apps?
- Did you have any issue with clashing global types between Jest and Jasmine?
- Did we consider https://www.cypress.io/ ? It's Chrome only but it looks much more fun to use than webdriver based e2e testing

src/client/__tests__/e2e/pages/blotter.component.ts
PendliKranthi: We can't implement data-qa id's for blotter components but i am gonna come with somthing better than xpath

PendliKranthi:
> A couple of general questions if I may :)
> 
> * What was the reasoning behind choosing Protractor instead of Selenium or WebdriverIO? I thought Protractor was aimed at Angular apps?
> * Did you have any issue with clashing global types between Jest and Jasmine?
> * Did we consider https://www.cypress.io/ ? It's Chrome only but it looks much more fun to use than webdriver based e2e testing

Protractor is used in most of the projects company wide so i guess that's the standard for them and it can be used for non angular projects as well.
Cypress is much better of course will bring up a discussion with qa folks and see we can implement them in any of the projects 
Jasmine is mostly used with protractor instead of jest

src/client/__tests__/e2e/pages/blotter.component.ts
PendliKranthi: @BenGladman i removed the validations using these locators for the blotter section as there are unstable with css selectors until we find the better solutions for them.

src/client/.gitignore
```diff
@@ -5,3 +5,4 @@ node_modules/
 .DS_Store
 # Excel temp files
 ~$*.xls*
+**/tmp/*
```
BenGladman: I believe `tmp/` will ignore everything in any `tmp` directory anywhere in the directory tree

PendliKranthi: this is my root folder for tmp files `src\client\__tests__\e2e\tmp` do you how exactly i should add this in .gitignore file.?i tried like this it didn't work

PendliKranthi: `**/__tests__/e2e/tmp/*` this is working is it okay?

BenGladman: It depends how specific you want to be. If you want to ignore all `tmp` dirs no matter where they are then `tmp/` will do. If you want to only ignore the `tmp` dir in that one place then `/src/client/__tests__/e2e/tmp/` should work. I'd go for `/tmp` myself :)

PendliKranthi: fixed it ignored specific to the tests only

src/client/__tests__/e2e/browser-manager.ts
```diff
@@ -0,0 +1,26 @@
...
+}
+
+async function prepareBrowser(): Promise<ProtractorBrowser> {
+  // We are testing a React project. We need to disable Angular sync.
+  await browser.waitForAngularEnabled(false)
+
+  await browser.get(browser.params.baseUrl, 10000)
+
+  // Wait for first rendering, for good measure.
+  await wait(5000)
```
BenGladman: would we be able to wait for something in the UI to happen instead of waiting for 5 seconds?

E.g. wait for `status-button__toggle-button` to show `Connected`?

PendliKranthi: I guess that button is more related to analytics panel anyway i tried it's unstable it doesn't connect all the time even though the page is loaded  and status will be changing in between and it's hard to assert

src/client/__tests__/e2e/pages/blotter.component.ts
BenGladman: Could you use something like:
`by.css('[data-qa="shell-route__blotter-wrapper"] .ag-body .ag-row:nth-child(1) [col-id="tradeId"]`

that looks reasonably resilient to me

PendliKranthi: i will try i am not really sure

PendliKranthi: it worked thanks i will update them all

BenGladman:
I wonder if it might be tidy to split the e2e tests out into a separate directory `{repoRoot}/src/e2e` instead of adding them into `{repoRoot}/src/client`? It would mean the tsconfigs and package deps would be nicely separated

Just a thought..

PR #1161, opened 2019-10-02
refactor the way we store notional in redux store

fdore:


fdore:
Consolidated PR here https://github.com/AdaptiveConsulting/ReactiveTraderCloud/pull/1177

PR #1166, opened 2019-10-03
add position offset to torn windows in browser

gomezsantiago:
Give a position offset to the torn windows so that they don't completely overlap one another.

## Before bug fix
browser-popout-bug

## After bug fix
artp-769

### DIFF
--- a/src/client/src/rt-platforms/browser/window.ts
+++ b/src/client/src/rt-platforms/browser/window.ts
@@ -4,9 +4,18 @@ import { WindowConfig } from '../types'
 
 type BrowserWindowProps = WindowConfig
 
-export const openBrowserWindow = (config: BrowserWindowProps, onClose?: () => void) => {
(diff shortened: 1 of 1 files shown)


src/client/src/rt-platforms/browser/window.ts
```diff
@@ -23,12 +32,22 @@ export const openBrowserWindow = (config: BrowserWindowProps, onClose?: () => vo
...
   return Promise.resolve(win)
-}
+}.bind({ prevWindow: null })
 
-function calculatePosition(center: string = 'parent', width: number, height: number) {
+function calculatePosition(
+  center: string = 'parent',
+  width: number,
+  height: number,
+  reference?: { top: number; left: number },
```
fdore: you could default to `{top: 0, left: 0}` to avoid the conditional check line 71

PR #1170, opened 2019-10-03
ARTP-747: Decimals in tile amount shortcut

rafgarciaa:
See [jira ticket](https://weareadaptive.atlassian.net/browse/ARTP-747?atlOrigin=eyJpIjoiNzMyNDBkNTk1MjFkNDc3NmE5YTM1N2FmMjgzZWE2MzYiLCJwIjoiaiJ9) for description.

**Bug:**
Entering `7.7m` in the notional input does not format it to `7,700,000`
pressing `Enter` in the input field doesn't do anything

**Fix:**
Entering `7.7m` in the notional input formats it to `7,700,000` 
by pressing `Enter`, input value is formatted
- example: entering `1.1` in the notional input field and pressing `Enter` formats to `1.10`


BenGladman:
I tested this locally, and it changes '1m' to '1,000,000' as soon as you press the 'm' key (i.e. no wait for enter, this is unchanged behaviour from before this change); but if you type '1.1m' you need to press enter before it changes to '1,100,000' (this is what the JIRA says should happen)

Is this what we want? it's a bit inconsistent

src/client/src/apps/MainRoute/widgets/spotTile/components/Tile/TileBusinessLogic.ts
```diff
@@ -120,7 +120,12 @@ export const getDerivedStateFromUserInput = ({
   const { symbol } = currencyPair
   const { rfqState } = spotTileData
 
-  const notional = !isEditMode(value) ? getFormattedValue(value) : value
+  let notional: string
+  if (type === 'keypress' && isEditMode(value)) {
+    notional = getFormattedValue(value)
+  } else {
+    notional = !isEditMode(value) ? getFormattedValue(value) : value
+  }
```
fdore: Looks like you could achieve the same result changing line 147
from `if (type === 'blur' && isEditMode(notional))`
to `if ((type === 'blur' || type === 'keypress') && isEditMode(notional))`

fdore:
> I tested this locally, and it changes '1m' to '1,000,000' as soon as you press the 'm' key (i.e. no wait for enter, this is unchanged behaviour from before this change); but if you type '1.1m' you need to press enter before it changes to '1,100,000' (this is what the JIRA says should happen)
> 
> Is this what we want? it's a bit inconsistent

Looks like it could be to do with the `editModeRegex`...

src/client/src/apps/MainRoute/widgets/spotTile/components/Tile/TileBusinessLogic.ts
rafgarciaa: @fdore you're right. Refactored the logic to this, thanks.

gomezsantiago:
> > I tested this locally, and it changes '1m' to '1,000,000' as soon as you press the 'm' key (i.e. no wait for enter, this is unchanged behaviour from before this change); but if you type '1.1m' you need to press enter before it changes to '1,100,000' (this is what the JIRA says should happen)
> > Is this what we want? it's a bit inconsistent
> 
> Looks like the `editModeRegex` needs updating... Not sure I fully understand the logic tbh (looking at the tests, 4.5 should not be formatted on the fly, but 12.87 should 🤷‍♂ )

I updated the regex, but that behavior is correct.  if `4.5` is formatted, then you'll immediately get `4.50` and you won't be able to enter something like `4.51` here are some examples of the regex: https://regex101.com/r/MrSCRE/7

PR #1171, opened 2019-10-04
Reset spot tile after expired RFQ stays idle for a while

fdore:
JIRA ARTP-712

### DIFF
--- a/src/client/src/apps/MainRoute/widgets/spotTile/epics/rfqEpics.ts
+++ b/src/client/src/apps/MainRoute/widgets/spotTile/epics/rfqEpics.ts
@@ -3,7 +3,7 @@ import { ofType } from 'redux-observable'
 import { delay, filter, map, mergeMap, takeUntil } from 'rxjs/operators'
 import { ApplicationEpic } from 'StoreTypes'
 import { SpotTileActions, TILE_ACTION_TYPES } from '../actions'
-import { from, merge, Observable, of, timer } from 'rxjs'
+import { concat, from, Observable, of, timer } from 'rxjs'
 import { RfqReceived, RfqRequest } from '../model/rfqRequest'
 import { SpotTileState } from '../spotTileReducer'
 import { CurrencyPairState } from '../../../data/referenceData'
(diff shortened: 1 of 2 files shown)


src/client/src/apps/MainRoute/widgets/spotTile/epics/rfqEpics.ts
```diff
@@ -78,20 +90,38 @@ export const rfqReceivedEpic: ApplicationEpic = action$ =>
...
+        mergeMap(() =>
+          merge(
+            of(rfqExpired({ currencyPair })),
+            timer(IDLE_TIME_MS).pipe(
+              mergeMap(() =>
+                from([
+                  setNotional({
+                    currencyPair: currencyPair.symbol,
+                    notional: {
+                      value: getDefaultNotionalValue(),
```
BenGladman: Related to [Santiago's comment on your other PR](https://github.com/AdaptiveConsulting/ReactiveTraderCloud/pull/1180#discussion_r332540177) should this be `getDefaultInitialNotionalValue` so that NZDUSD resets to 10m?

src/client/src/apps/MainRoute/widgets/spotTile/epics/rfqEpics.ts
```diff
@@ -78,20 +90,38 @@ export const rfqReceivedEpic: ApplicationEpic = action$ =>
...
-        filter(
-          cancelAction =>
-            cancelAction.payload.currencyPair.symbol === action.payload.currencyPair.symbol,
-        ),
+        filter(cancelAction => cancelAction.payload.currencyPair.symbol === currencyPair.symbol),
       )
 
       return timer(action.payload.timeout + 1000).pipe(
-        map(() => rfqExpired({ currencyPair: action.payload.currencyPair })),
+        mergeMap(() =>
```
BenGladman: Potentially you could use `concat` to make this slightly clearer (I think) because the timers would then be at a similar level...

```
return concat(
  timer(action.payload.timeout + 1000).pipe(map(() => /* expired */)),
  timer(IDLE_TIME_MS).pipe(mergeMap(() => /* reset */))
).pipe(takeUntil(cancel$))
```

src/client/src/apps/MainRoute/widgets/spotTile/epics/rfqEpics.ts
fdore: We want the prices to stream again after a reset, so if we reset to 10m, then, that won't be the case...

PR #1173, opened 2019-10-04
make redux store available to child windows

fdore:
**DO NOT MERGE**

This code seems to cause the following error `Uncaught Invariant Violation: Maximum update depth exceeded` when closing a torn spot tile...


BenGladman:
I've been looking this afternoon at rendering the external windows in a Portal from the main app instead of creating a completely new instance... this could be more performant I think, and the context stuff would automatically be there. Needs a little work, but do you think it's a good idea worth continuing with?

fdore:
I have a feeling we were using Portals for external windows at some point. Not sure why this is not the case anymore. Is that right @bhavesh-desai-scratch ?

BenGladman:
> I'm pretty sure we were using Portals for external windows at some point (#745). Not sure why this is not the case anymore

You're correct, the original Portal was removed in #855 . I'll put up a PR anyway as it looks like it works pretty well for me and would prevent the need for this workaround to share the redux store.

PR #1177, opened 2019-10-08
Move notional to redux store

fdore:


src/client/src/apps/MainRoute/widgets/spotTile/components/Tile/Tile.tsx
```diff
@@ -57,7 +60,17 @@ class Tile extends React.PureComponent<TileProps, TileState> {
 
   // State management derived from props
   static getDerivedStateFromProps(nextProps: TileProps, prevState: TileState) {
-    return getDerivedStateFromProps(nextProps, prevState)
+    const { setTradingMode, spotTileData, currencyPair } = nextProps
+    return {
+      ...getDerivedStateFromProps(nextProps, prevState),
+      ...getDerivedStateFromUserInput({
```
fdore: This is not great, and should be the object of another PR

PR #1180, opened 2019-10-08
fix merge issue

fdore:


src/client/src/apps/MainRoute/widgets/spotTile/components/SpotTile.tsx
```diff
@@ -55,7 +55,7 @@ export default class SpotTile extends PureComponent<Props> {
     } = getConstsFromRfqState(rfqState)
     const showResetButton =
       !isTradeExecutionInFlight &&
-      getDefaultNotionalValue(currencyPair) !== notional &&
+      getDefaultNotionalValue() !== notional &&
```
gomezsantiago: I don't think this fix will work.  the resetButton needs to be shown when the notional is different from it's initial default amount.  for currencyPair NZD/USD that default notional is `10,000,000`, which is different than all the other default notionals of `1,000,000`.... For NZD/USD the reset button shouldn't be shown when the notional is `10,000,000`

fdore: Ok - I guess the reason I used getDefaultNotionalValue is to be consistent with what happens when we reset an Rfq... but I don't have a strong view on that one.

PR #1183, opened 2019-10-08
Glue42 fully functional with excel

my-code-is-lit:


src/client/src/rt-platforms/glue/utils/excel.ts
```diff
@@ -0,0 +1,102 @@
...
+  blotterData
+    .map(data => ({
+      ...data,
+      status: `${data.status.charAt(0).toUpperCase()}${data.status.slice(1)}`,
+      tradeDate: convertTradeDate(data.tradeDate),
+    }))
+    .reverse()
+
+const convertTradeDate = (tradeDate: string) => {
+  const date = new Date(tradeDate)
```
fdore: Can we not use moment to format the date?

src/client/src/rt-platforms/glue/window.ts
```diff
@@ -0,0 +1,182 @@
...
+  if (isTabWindow) {
+    listOfOpenedWindows.push(win)
+    window.glue.interop.invoke('toggleHeaderButtons', {
+      numberOfOpenedWindows: listOfOpenedWindows.length,
+    })
+  }
+
+  if (win) {
+    if (onClose) {
+      win.onClose(onClose)
```
fdore: Should that not be `win.onClose = onClose` instead?

src/client/src/rt-platforms/glue/window.ts
```diff
@@ -0,0 +1,182 @@
...
+    modifiedWidth,
+    modifiedHeight,
+    relativeTo,
+    relativeDirection,
+  } = calculatePosition(myWindow, width, height, url)
+  const fullUrl = `${location.href.slice(0, -1)}${url}`
+  const isTabWindow = isSpot(url)
+
+  const win = await window.glue.windows.open(name, fullUrl, {
+    title: capitalizeFirstLetter(name.split(' ')[0]),
```
fdore: can use capitalize from lodash instead

src/client/src/rt-platforms/glue/utils/excel.ts
```diff
@@ -0,0 +1,102 @@
...
+
+/**
+ * Converts the blotter data in human readable format where necessary.
+ * New record is displayed on top.
+ */
+const convertBlotterData = (blotterData: BlotterData[]) =>
+  blotterData
+    .map(data => ({
+      ...data,
+      status: `${data.status.charAt(0).toUpperCase()}${data.status.slice(1)}`,
```
fdore: can use capitalize from lodash

src/client/src/rt-platforms/glue/components/GlueHeader.tsx
```diff
@@ -0,0 +1,82 @@
...
+    waitForGlue()
+  }, [])
+
+  // TODO expose onClick on logo?
+
+  const stackAll = () => window.glue.interop.invoke('stackAllWindows')
+
+  const tabAll = () => window.glue.interop.invoke('tabAllWindows')
+
+  const toggleCollapse = () => window.glue.interop.invoke('toggleCollapse')
```
fdore: are these equivalent to minimize, maximize etc.? If so, we should follow the same approach as with openfin, where these are passed as props, and set in <RouteWrapper />

src/client/src/rt-platforms/glue/window.ts
my-code-is-lit: @fdore I tried setting `win.onClose = onClose`, it won't reset the tiles back to parent on close. This isn't a browser, it's Glue42

src/client/src/rt-platforms/glue/components/GlueHeader.tsx
my-code-is-lit: @fdore these are dependent on how many child windows (tiles) are open, so they are not equivalent to minimize/maximize/close. Unique situation to Glue42, best if implemented directly in the rt-platform module

src/client/src/rt-platforms/glue/window.ts
fdore: 👍  should have checked the api

src/client/src/rt-platforms/glue/utils/excel.ts
```diff
@@ -71,32 +73,6 @@ const convertBlotterData = (blotterData: BlotterData[]) =>
   blotterData
     .map(data => ({
       ...data,
-      status: `${data.status.charAt(0).toUpperCase()}${data.status.slice(1)}`,
-      tradeDate: convertTradeDate(data.tradeDate),
+      status: _.startCase(_.toLower(data.status)),
```
fdore: why don't you use _.capitalize?

src/client/src/rt-platforms/glue/window.ts
```diff
@@ -82,7 +83,7 @@ export const openGlueWindow = async (config: BrowserWindowProps, onClose?: () =>
   const isTabWindow = isSpot(url)
 
   const win = await window.glue.windows.open(name, fullUrl, {
-    title: capitalizeFirstLetter(name.split(' ')[0]),
+    title: _.startCase(_.toLower(name)),
```
fdore: same here

src/client/src/rt-platforms/glue/utils/excel.ts
my-code-is-lit: @fdore honestly, I don't want to split hairs here, but the intent was to use title case `startCase` and `_.toLower` converts anything from second character onwards in the string to lower-case in case something is out of place

PR #1186, opened 2019-10-09
create shared server side folder

fdore:
Example of use 
```
// node-server/service/tsconfig.json
{
  "extends": "../../node-server/tsconfig-base",
  "compilerOptions": {
    "outDir": "lib"
  },
  "references": [
    {"path": "../shared"}
  ],
  "include": ["./src/**/*"]
}
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,4 @@
-src/client/node_modules
+node_modules
 src/client/coverage
 .idea/
 .vs/
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,4 @@
-node_modules
+node_modules/
 src/client/coverage
 .idea/
 .vs/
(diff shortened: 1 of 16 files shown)


src/node-server/tsconfig-base.json
```diff
@@ -0,0 +1,11 @@
+{
+  "compilerOptions": {
+    "target": "es6",
+    "module": "commonjs",
+    "lib": ["es6","dom"],
```
BenGladman: probably not `dom`?

.gitignore
```diff
@@ -1,4 +1,4 @@
-src/client/node_modules
+node_modules
```
BenGladman: Usually specify `node_modules/` but doesn't matter much

BenGladman:
Cool -- interested to try out composite projects as not really looked at it before

PR #1187, opened 2019-10-09
Refactored ARTP 728

PendliKranthi:


src/client/__tests__/e2e/pages/tile.component.ts
```diff
@@ -66,6 +68,7 @@ export class TileComponent {
     }
     await waitForElementToBeClickable(this.browser, buttonElement)
     await buttonElement.click()
+    await wait(2000)
```
fdore: Would be nice not to have to hard code these...

PendliKranthi: I had to use this wait as the application is not up to the speed of tests it's just an extra precaution to stabilize the tests.

PR #1189, opened 2019-10-09
ARTP-782: Refactor notional amount input

BenGladman:
- Extract the numeric format-as-you-type behaviour into a new component
- Remove the 'notional' value from local state in the Tile component
- Change the 'notional' value in redux store from string to number
- Move some date utils from notional dir (where they weren't used to blotter (where they are)

Also fixes the problem described [in the second comment on this JIRA](https://weareadaptive.atlassian.net/browse/ARTP-772)


src/client/src/apps/MainRoute/widgets/spotTile/components/Tile/TileBusinessLogic.test.ts
```diff
@@ -16,108 +11,11 @@ import { PriceMovementTypes } from '../../model/priceMovementTypes'
...
-
-test('isEditMode should be false', () => {
-  shouldFormatOnTheFly.forEach(v => {
-    console.log('shouldFormatOnTheFly', v)
-    expect(isEditMode(v)).toBe(false)
-  })
-})
-
 // invalid trading values, should disable trading (Buy/Sell buttons)
 // https://regex101.com/r/OWDRCO/2
```
fdore: Do we still need that comment?

src/client/src/apps/MainRoute/widgets/spotTile/components/TileSwitch.tsx
```diff
@@ -48,10 +48,18 @@ const TileSwitch: React.FC<Props> = ({
     isRfqStateNone,
   } = getConstsFromRfqState(spotTileData.rfqState)
 
+  const spotTileDataWithNotional =
```
bhavesh-desai-scratch: You should not need the `=== null`

src/client/src/apps/MainRoute/widgets/spotTile/components/TileSwitch.tsx
```diff
@@ -69,9 +77,7 @@ const TileSwitch: React.FC<Props> = ({
           <TileBooking
             show={!spotTileData.isTradeExecutionInFlight && isRfqStateCanRequest}
             color="blue"
-            onBookingPillClick={() =>
-              rfq.request({ notional: getNumericNotional(notional), currencyPair })
-            }
+            onBookingPillClick={() => rfq.request({ notional, currencyPair })}
```
bhavesh-desai-scratch: is it worth using a useCallback here?

src/client/src/apps/MainRoute/widgets/blotter/components/blotterUtils.ts
```diff
@@ -1,10 +1,18 @@
...
-import { formatDate, UtcFormatDate } from '../../spotTile/components/notional/utils'
 import SetFilter from './filters/SetFilter'
 import numeral from 'numeral'
 import { capitalize } from 'lodash'
 
+function UtcFormatDate(date: Date, format: string = '%b %e, %H:%M:%S') {
+  return utcFormat(format)(date)
+}
+
+function formatDate(date: Date, format: string = '%b %e, %H:%M:%S') {
```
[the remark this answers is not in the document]
BenGladman: I prefer `function` myself too where it makes sense. I don't think there'll be a lint fix for it because arrows are semantically different from functions so it would be dangerous to auto change

src/client/src/apps/MainRoute/widgets/spotTile/components/TileSwitch.tsx
BenGladman: I do because `notional` could be `0`

src/client/src/apps/MainRoute/widgets/spotTile/components/TileSwitch.tsx
BenGladman: Only if we needed referential identity (e.g. if `TileBooking` was `memo`d), which we don't

src/client/src/apps/MainRoute/widgets/spotTile/components/Tile/TileBusinessLogic.test.ts
BenGladman: yep, it refers to the new code `const invalidTradingValues = [0, -0, Infinity, -Infinity, NaN]` on line 16

PR #1191, opened 2019-10-09
Bugfix/artp 781

gomezsantiago:
ensure only 1 `priceButton` status (`Awaiting Price`, `Price Unavailable`...) is visible at a time when executing a trade

### Before bugfix
actual-executing-state

### After bugfix
Screen Shot 2019-10-09 at 4 00 05 PM


src/client/src/apps/MainRoute/widgets/spotTile/components/PriceControls/PriceControls.tsx
```diff
@@ -49,69 +49,67 @@ const PriceControls: React.FC<Props> = ({
 
   const isDisabled = !isRfqStateReceived && disabled
   const hasPrice = Boolean(priceData.bid && priceData.ask)
+  let priceButtonDisabledStatus: JSX.Element = null
+  let priceButtonBuy: JSX.Element = null
+  let priceButtonSell: JSX.Element = null
+
+  if (isRfqStateNone || isRfqStateReceived || isRfqStateExpired || isTradeExecutionInFlight) {
```
bhavesh-desai-scratch: Might be worth putting these 4 conditions  in a small function

BenGladman:
Alternatively you could have refactored this as a ternary - prettier is pretty good at formatting ternaries containing jsx because it's a pretty common pattern...

```
<PriceControlsStyle>
  {isRfqStateNone || isRfqStateReceived || isRfqStateExpired || isTradeExecutionInFlight ? (
    <x />
  ) : isRfqStateCanRequest ? (
    <y />
  ) : isRfqStateRequested  ? (
    <z />
  ) : null}
</PriceControlsStyle>
```

src/client/src/apps/MainRoute/widgets/spotTile/components/PriceControls/PriceControls.tsx
gomezsantiago: I updated this file along with the TileBusinessLogic to clean things up a little

src/client/src/apps/MainRoute/widgets/spotTile/components/PriceControls/PriceControls.tsx
```diff
@@ -47,71 +47,57 @@ const PriceControls: React.FC<Props> = ({
...
+        priceAnnounced={isRfqStateReceived}
+        disabled={disabled}
+        expired={isRfqStateExpired}
+        currencyPairSymbol={currencyPair.symbol}
+      />
+    ) : null
+  }
+
+  if (!showPrices) {
+    var priceButtonDisabledStatus = isRfqStateCanRequest ? (
```
BenGladman: Ouch, I've not seen a `var` in the wild for a long time!

Could we get rid of it by removing the `if` block and moving the `showPrices` logic into the ternary?

gomezsantiago: sure, I'll move the `!showPrices` logic into both of the ternaries, but I personally don't see the issue with using `var` in this particular case. Is it a problem with consistency?

BenGladman: Personally I avoid ever using it because of its "surprising" hoisting and scoping behaviour. You've used that behaviour here but it's not obvious

I'm a bit surprised we don't lint against it...

PR #1192, opened 2019-10-10
add NLP service skeleton

fdore:


src/node-server/nlp/package.json
```diff
@@ -0,0 +1,49 @@
...
+    "@types/dotenv": "^6.1.0",
+    "@types/node": "^11.9.3",
+    "@types/uuid": "^3.4.4",
+    "cross-env": "^5.2.0",
+    "husky": "^1.3.1",
+    "lint-staged": "^8.1.4",
+    "nodemon": "^1.18.10",
+    "prettier": "1.16.4",
+    "ts-node": "^8.0.2",
+    "tslint": "^5.13.0",
```
BenGladman: Should we start using typesccript-eslint now because tslint is going to be deprecated pretty soon

src/node-server/nlp/src/index.ts
```diff
@@ -0,0 +1,84 @@
...
+import {
+  filter,
+  map,
+  mapTo,
+  shareReplay,
+  switchMap,
+  takeUntil
+} from "rxjs/operators";
+import uuid from "uuid/v1";
+import logger from "../../shared/src/logger";
```
BenGladman: maybe would be nice to set up `shared` in tsconfig `paths`  so could `import { logger } from 'shared'`?

fdore: Definitely

PR #1194, opened 2019-10-10
integrate with DialogFlow

fdore:


src/node-server/nlp/src/dialogFlowClient.ts
```diff
@@ -0,0 +1,23 @@
+import uuid from "uuid/v1";
+import DialogueFlow from "dialogflow";
```
BenGladman: you've spelled the import differently from the package name

PR #1197, opened 2019-10-11
build nlp service

fdore:


src/node-server/shared/src/transport/AutobahnSessionProxy.ts
```diff
@@ -3,7 +3,7 @@ import { Session, Subscription } from 'autobahn'
 /**
  * AutobahnSessionProxy: makes the autobahn session api more explicit, aids testing
  */
-export default class AutobahnSessionProxy {
+export class AutobahnSessionProxy {
```
BenGladman: like

BenGladman:
I do feel we should be able to do this without webpack.. maybe I can have a play after this is merged

PR #1199, opened 2019-10-11
keep popout windows ontop when a new window is opened

gomezsantiago:
focus all external windows when a new one is opened

## Before bugfix
poput-windows

## After bugfix
popout-windows-focus


src/client/src/rt-platforms/browser/window.ts
```diff
@@ -44,9 +48,10 @@ export const openBrowserWindow = function(
   }
 
   this.prevWindow = win
+  this.externalWindows = this.externalWindows.filter(window => !window.closed).concat(win)
```
BenGladman: Not your change ...but using `this` to hold the state is a bit confusing to me. Would be simpler IMO to define a couple of `let` vars outside of the function and use the closure.

src/client/src/rt-platforms/browser/window.ts
```diff
@@ -5,7 +5,7 @@ import { WindowConfig } from '../types'
 type BrowserWindowProps = WindowConfig
 
 export const openBrowserWindow = function(
-  this: { prevWindow: Window },
+  this: { prevWindow: Window; externalWindows: Window[] },
```
BenGladman: if you brought the `this.externalWindows = this.externalWindows.filter(window => !window.closed)` you could remove the `prevWindow` altogether I think, and use `this.externalWindows[this.externalWindows.length -1]` instead

src/client/src/rt-platforms/browser/window.ts
gomezsantiago: this was my change from a previous PR actually.  I removed the `this` binding to make the function more intuitive

src/client/src/rt-platforms/browser/window.ts
gomezsantiago: I used the `prevWindow` for the `left` and `top` properties in the `windowReferencePosition` object so I think it's still a good idea to keep it as a separate variable.  However, I did update the function to use the last element in the array for the `prevWindow`

PR #1201, opened 2019-10-11
Add Nlp service to CI

fdore:


src/docker-compose.yml
```diff
@@ -83,3 +83,15 @@ services:
     environment:
      - BROKER_HOST=broker
      - BROKER_PORT=8000
+  nlp:
+    build:
+      dockerfile: Nlp.Dockerfile
```
bhavesh-desai-scratch: Is the casing right here?

bhavesh-desai-scratch:
You need to create some k8 files as well to get it into enviroments

src/docker-compose.yml
fdore: yes

fdore:
> You need to create some k8 files as well to get it into enviroments

Forgot to run it from the circle ci config

PR #1203, opened 2019-10-11
Fix trade and value date

my-code-is-lit:


src/client/src/apps/MainRoute/widgets/blotter/components/blotterUtils.ts
```diff
@@ -1,22 +1,18 @@
...
+import { utcFormat } from 'd3-time-format'
 import { Trade, TradeStatus } from 'rt-types'
 import SetFilter from './filters/SetFilter'
 import numeral from 'numeral'
 import { capitalize } from 'lodash'
 
-const dateRenderer = (value: Date) => {
-  return DateTime.fromJSDate(value).toFormat('dd-MMM hh:mm:ss')
+function UtcFormatDate(date: Date, format: string = '%b %e, %H:%M:%S') {
+  return utcFormat(format)(date)
```
BenGladman: I only just changed this to use luxon in #1200, is there a reason to use d3-time-fornat instead?

If there is then you should add it to the project deps because I removed it

my-code-is-lit: @BenGladman I put back the luxon date. Thanks!

src/client/src/apps/MainRoute/widgets/blotter/components/blotterUtils.ts
```diff
@@ -5,18 +5,14 @@ import SetFilter from './filters/SetFilter'
 import numeral from 'numeral'
 import { capitalize } from 'lodash'
 
-const dateRenderer = (value: Date) => {
-  return DateTime.fromJSDate(value).toFormat('dd-MMM hh:mm:ss')
+function UtcFormatDate(date: Date, format: string = '%b %e, %H:%M:%S') {
```
fdore: format is not used

PR #1206, opened 2019-10-14
basic integration with NLP service

fdore:


src/client/src/apps/SpotlightRoute/Spotlight.tsx
```diff
@@ -39,8 +43,15 @@ const Contacting = styled.span`
 
 const INPUT_ID = 'spotlight'
 
+const autobahn = new AutobahnConnectionProxy(
+  process.env.REACT_APP_BROKER_HOST || location.hostname,
+  'com.weareadaptive.reactivetrader',
+  +(process.env.REACT_APP_BROKER_PORT || location.port),
```
sizov: very minor and edge case: looks like port is a number, which means that if `process.env.REACT_APP_BROKER_PORT` is set is 0, it will be disregarded in favor of `location.port`

src/client/src/apps/SpotlightRoute/Spotlight.tsx
```diff
@@ -51,10 +62,17 @@ export const Spotlight: FC = () => {
...
-        setTimeout(
-          () => dispatch({ type: 'RECEIVE_RESPONSE', response: `I heard "${value}".` }),
-          500,
-        )
+        serviceStub.createRequestResponseOperation('nlp', 'getNlpIntent', value)
+          .pipe(
+            timeout(5000),
+            take(1)
+          )
+          .subscribe((response: any) => {
```
sizov: should we leave a comment to fix `any` in the future (to introduce actual type)?

src/client/src/apps/SpotlightRoute/Spotlight.tsx
fdore: I'm not sure 0 would be valid anyway..

src/client/src/apps/SpotlightRoute/Spotlight.tsx
fdore: let me add the types for dialogflow

src/client/src/apps/SpotlightRoute/hooks.ts
```diff
@@ -0,0 +1,48 @@
...
+      multicast(() => {
+        return new ReplaySubject<ServiceCollectionMap>(1)
+      }),
+      refCount(),
+    )
+
+    const loadBalancedServiceStub = new ServiceStubWithLoadBalancer(serviceStub, serviceStatus$)
+
+    setServiceStub(loadBalancedServiceStub);
+  }, [])
```
sizov: looks like `autobahn` should be added as a dependency. Do we ave linting for the react hooks dependencies setup?

src/client/src/apps/SpotlightRoute/Spotlight.tsx
```diff
@@ -51,10 +62,17 @@ export const Spotlight: FC = () => {
       case 'Enter':
         const value = e.currentTarget.value
         dispatch({ type: 'SEND_REQUEST', request: value })
-        setTimeout(
-          () => dispatch({ type: 'RECEIVE_RESPONSE', response: `I heard "${value}".` }),
-          500,
-        )
+        serviceStub.createRequestResponseOperation('nlp', 'getNlpIntent', value)
```
sizov: do you have any patterns in the project to separate concerns of component VS server-side communication (like redux middleware or something similar)?

fdore: this is v much poc at this stage

src/client/src/apps/SpotlightRoute/hooks.ts
fdore: autobahn is imported from AutobahnConnection

src/client/src/apps/SpotlightRoute/Spotlight.tsx
fdore: we'll pbly want to use redux-observable at some point anyway, to handle concurrent requests.

src/client/src/apps/SpotlightRoute/hooks.ts
sizov: i mean it has to be a dependency of the hook

PR #1207, opened 2019-10-14
Bugfix/artp 783: Tile RFQ timeout (reset to ESP) not working on quote reject

rafgarciaa:
[Jira issue](https://weareadaptive.atlassian.net/browse/ARTP-783)

Enable reseting of tile to default state after 60 second idle time when user rejects an RFQ


src/client/src/apps/MainRoute/widgets/spotTile/epics/rfqEpics.test.ts
```diff
@@ -211,44 +211,57 @@ describe('rfqEpics', () => {
       })
     })
 
-    const expiryCases = [['rejected', 'b', 'e'], ['reset', 'd', 'g']]
+    describe('when rfq is reject', () => {
```
fdore: => `rejected`

PR #1209, opened 2019-10-14
ARTP-804: Changelog Generator Script

pmilla1606:
This PR adds a script to generate a `MARKDOWN.md` file comprised of merges to master.

For now it includes only the most basic information but it can be easily extended to include anything available from a `git log` output.

It's also trivial to call this from a pre-commit hook or from the CI/CD pipeline. I have not done that as I want to get some feedback from the group about what we think the best approach would be.


scripts/make_changelog.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+
+COMMITS=$(git log --merges --pretty=format:"%H" origin/master)
+
+GIT_USER_NAME=$(git config user.name)
+GIT_USER_EMAIL=$(git config user.email)
+DATE=$(date '+%Y-%m-%d')
+
+MARKDOWN="## Changelog \n ### Generated by [$GIT_USER_NAME]($GIT_USER_EMAIL) on ${DATE}"
```
fdore: Not sure we need to show `Generated by [$GIT_USER_NAME]($GIT_USER_EMAIL)`

fdore:
Looks good. However, we probably need to tag when we merge to master, and then show the list of merges for that release...

scripts/make_changelog.sh
pmilla1606: Sure, I will remove.

pmilla1606:
@fdore yes that's another approach which would work. Do you want me to reconfigure this so it looks for tags and prints out commits in between tags?

fdore:
> @fdore yes that's another approach which would work. Do you want me to reconfigure this so it looks for tags and prints out commits in between tags?

I think that would be a good idea, as we want to see what goes out in each release...

pmilla1606:
@fdore Updated the script to allow either (tag range or everything). Let me know what you think.

fdore:
> @fdore Updated the script to allow either (tag range or everything). Let me know what you think.

I was more thinking about outputting something like that:

**1.x.1 (2019-10-15)**
- [PR]: PR description
- [PR]: PR description
- [PR]: PR description

**1.x.0 (2019-10-03)**
- [PR]: PR description
- [PR]: PR description
- [PR]: PR description

rikoe:
I would personally like us to follow https://keepachangelog.com/en/1.0.0/, i.e. have a CHANGELOG.md file that shows changes between releases.

But that would require us to move to a release and versioning model in the first place! Which we should do...

pmilla1606:
@rikoe that's essentially what I'm working on right now, you are correct that we need to start tagging releases for this to work though. I'll push my changes soon and will update here.

pmilla1606:
@rikoe @fdore Changes made. See attached for sample output generated from https://github.com/Kopernicus/Kopernicus

Screen Shot 2019-10-18 at 9 36 24 AM

ilikejames:
Shame we are not using the [conventional commits standard](https://www.conventionalcommits.org/en/v1.0.0-beta.2/). 
It makes the git history way more saner for humans and great for machines. On a previous project I generated changelogs for QA/PM consumption using only the `^feat|^fix`.

sizov:
please can we hold of with merging this? I am adding conventional commits support to the repo and it will help us with generating changelog.

sizov:
also, what is `backport` in the generated log? I assume we need an actual version of the project, but there's no top-level versioning for now at all as I understand. @ilikejames is working on this.

So am I missing something, or version in the changelog does not have any meaning yet?

fdore:
> also, what is `backport` in the generated log? I assume we need an actual version of the project, but there's no top-level versioning for now at all as I understand. @ilikejames is working on this.

The screenshot is an example of the script against the `https://github.com/Kopernicus/Kopernicus` repo

sizov:
> The screenshot is an example of the script against the https://github.com/Kopernicus/Kopernicus repo

thank you @fdore , i didn't know that

PR #1216, opened 2019-10-15
Address react-hooks/exhaustive-deps warnings

rafgarciaa:
[Enabling `eslint` to warn missing deps caused some components to throw warnings during compile time.](https://github.com/AdaptiveConsulting/ReactiveTraderCloud/pull/1212)

Screen Shot 2019-10-15 at 9 47 15 AM

** I purposely put `// eslint-disable-next-line react-hooks/exhaustive-deps` on some `useEffect` hooks since I assumed they are being used as a `componentDidMount()` (empty deps array).

### DIFF
--- a/src/client/src/apps/MainRoute/widgets/analytics/components/analyticsLineChart/AnalyticsLineChart.tsx
+++ b/src/client/src/apps/MainRoute/widgets/analytics/components/analyticsLineChart/AnalyticsLineChart.tsx
@@ -80,7 +80,8 @@ const LineCharts: React.FC<LineChartProps> = React.memo(props => {
   useEffect(() => {
     const newOffset = offsetState + 1 === intervalWidth ? 0 : offsetState + 1
     setOffsetState(newOffset)
-  }, [props, offsetState])
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [props])
 
   const getDataPoint = useCallback(
     (dataPoints: DataPoint[]) =>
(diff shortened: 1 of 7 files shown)


src/client/src/apps/MainRoute/widgets/analytics/components/analyticsLineChart/AnalyticsLineChart.tsx
```diff
@@ -80,7 +80,7 @@ const LineCharts: React.FC<LineChartProps> = React.memo(props => {
   useEffect(() => {
     const newOffset = offsetState + 1 === intervalWidth ? 0 : offsetState + 1
     setOffsetState(newOffset)
-  }, [props])
+  }, [props, offsetState])
```
rafgarciaa: @fdore do we need `props` here as a dependency? It doesn't look right to me.

fdore: `offsetState` should be enough

fdore: actually, I don't think it makes sense to have a dep on`offsetState` here, given the only place were we change it is inside the effect

rafgarciaa: should I leave it as an empty array then?

rafgarciaa: Ah, got it. It's dependent on the changes in `model.seriesData`.

PR #1219, opened 2019-10-15
Artp 586

gomezsantiago:
show RFQ prices in Analytics Tile view when executing a trade

## Before bugfix:
rfq-executing-price-bug

## After bugfix:
rfq-executing-price-bug-fix


src/client/src/apps/MainRoute/widgets/spotTile/components/PriceControls/PriceControls.tsx
```diff
@@ -20,6 +20,7 @@ interface Props {
   disabled: boolean
   rfqState: RfqState
   isTradeExecutionInFlight: boolean
+  analyticsView?: boolean
```
fdore: I know we are not quite consistent there, but I find calling boolean props `is...` or `has...` (in this case `isAnalyticsView`) a bit more obvious when reading the code.

src/client/src/apps/MainRoute/widgets/spotTile/components/PriceControls/PriceControls.tsx
```diff
@@ -29,6 +30,7 @@ const PriceControls: React.FC<Props> = ({
   rfqState,
   disabled,
   isTradeExecutionInFlight,
+  isAnalyticsView: isAnalyticsView,
```
fdore: => `isAnalyticsView`

PR #1220, opened 2019-10-16
refactor Spotlight bootstrapping

fdore:
- move serviceStub to context
- launch app on click, instead of on enter, so we can potentially support multiple results


src/client/src/apps/SpotlightRoute/handleIntent.ts
```diff
@@ -29,10 +29,10 @@ function showCurrencyPair(currencyPair: string, { window }: PlatformAdapter) {
 }
 
 export const handleIntent = (
-  response: DetectIntentResponse[],
+  response: DetectIntentResponse,
   platformAdapter: PlatformAdapter,
 ) => {
-  const queryResult = response[0].queryResult
+  const queryResult = response.queryResult
```
sizov: why this has changed?

sizov: is it not an array anymore?

sizov: i see now, ignore please

src/client/src/apps/SpotlightRoute/Spotlight.tsx
```diff
@@ -109,11 +109,13 @@ export const Spotlight: FC = () => {
       <Response>
         {contacting ? (
           <>
-            <AdaptiveLoader size={14} speed={0.8} seperation={1.5} type="secondary" />
+            <AdaptiveLoader size={14} speed={0.8} seperation={1.5} type="secondary"/>
             <Contacting>Contacting…</Contacting>
           </>
         ) : (
-          response
+          <Suggestion onClick={() => handleIntent(response, platform)}>
```
sizov: as i understand you want user to explicitly select an intent, is this correct? wouldn't it be more "responcive" if we handle intent straightway? (at least when we know there's a single intent)

src/client/src/apps/SpotlightRoute/handleIntent.ts
fdore: I guess I can leave it as it was. I'm just not sure we ever get more than one result

PR #1223, opened 2019-10-16
Intial nlp blotter handler

sizov:
just opening blotter for now, no filtering

### DIFF
--- a/src/client/src/apps/SpotlightRoute/intent-handlers/defaultWindowConfig.ts
+++ b/src/client/src/apps/SpotlightRoute/intent-handlers/defaultWindowConfig.ts
@@ -1,0 +1,14 @@
+import { WindowConfig } from 'rt-platforms'
+
+export const defaultConfig: WindowConfig = {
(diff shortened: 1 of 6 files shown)


src/client/src/apps/SpotlightRoute/intent-handlers/showCurrencyPair.ts
```diff
@@ -0,0 +1,10 @@
+import { PlatformAdapter } from 'rt-platforms'
```
sizov: just moved from `handleIntent.ts `

src/client/src/apps/SpotlightRoute/intent-handlers/defaultWindowConfig.ts
```diff
@@ -0,0 +1,14 @@
+import { WindowConfig } from 'rt-platforms'
```
sizov: just moved from `handleIntent.ts ` to be resued elsewhere

PR #1224, opened 2019-10-16
add support for inline trade results

fdore:


src/client/src/apps/SpotlightRoute/Spotlight.tsx
```diff
@@ -124,6 +125,14 @@ export const Spotlight: FC = () => {
         )
         : null
       }
+      {isTradeIntent(response)
+        ? (
+          <Suggestion onClick={() => handleIntent(response, platform)}>
+            <InlineBlotter currency={response.queryResult.parameters.fields.Currency.stringValue}/>
```
sizov: consider reusing `getCurrency`

PR #1228, opened 2019-10-17
move nlp request to effect

fdore:


src/client/src/apps/SpotlightRoute/Spotlight.tsx
```diff
@@ -62,6 +69,37 @@ export const Spotlight: FC = () => {
...
+        (err: any) => {
+          console.error(err)
+          dispatch({type: 'RECEIVE_RESPONSE', response: null})
+        },
+      )
+
+    return () => {
+      subscription.unsubscribe()
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
sizov: why did you need this comment? was it because you didn't want to add `request` as dep?

fdore: because request keeps changing as you type. And I only want to trigger the request when contacting goes to `true`

fdore: because we have this 
```
const handleChange: ChangeEventHandler<HTMLInputElement> = e => {
    dispatch({type: 'SET_REQUEST', request: e.target.value})
  }
```

sizov: that's interesting, I first thought that then `request` should probablt go into ref (not the state), but for now maybe it would be worth just adding a comment on line 99 :

>> because request keeps changing as you type. And I only want to trigger the call when contacting goes to true

PR #1233, opened 2019-10-17
Added test for high notional and reset button

MartinLavigneAdaptive:
Added test for notional higher than allowed for streaming and the reset button


src/client/__tests__/e2e/tests/trade.spec.ts
```diff
@@ -50,6 +51,19 @@ describe('UI Tests for Reactive Trader Cloud Web Application', async () => {
     })
   })
 
+  it('Should validate a notional value larger than the maximum allowed', async () => {
+    await mainPage.tile.setNotional('EURToUSD', '99999999')
+    const notional = await mainPage.tile.tradeType.EURToUSD.notional
+    var returnedNotional = notional.getAttribute('value')
```
fdore: => `const`

PR #1241, opened 2019-10-21
Set up smoke tests

ilikejames:
I've moved the existing e2e tests into its own package. Felt cleaner and will mean we can start up the docker container / tear it down during build & deploy. 

Happy if we have a better name than `client-tests`. 

The overall thought was we could run the container against the deployed client and use the existing e2e tests to verify the deploy/build. 

docker-compose

TODO (either in this PR or in another): 
- [ ] Update circleci definition
- [ ] Post to slack the results


src/client-tests/src/protractor.config.js
```diff
@@ -31,51 +32,56 @@ function getBrowserCapabilities() {
...
     includeStackTrace: true,
     defaultTimeoutInterval: 300000,
-    print: () => {},
+    print: () => { },
   },
-  onPrepare: function() {
-    setUpCustomLocators()
+  onPrepare: function () {
+    setUpCustomLocators();
+    require('ts-node').register()
```
ilikejames: I added this to make both debugging easier, and the process simpler to just debug the ts-node files directly. 
This works really well in vs-code (and intellij) with the correct config. 
debugging

src/client-tests/README.md
```diff
@@ -0,0 +1,27 @@
+# Client Tests
```
ilikejames: I know its usual in this repo to have docs all together, but this seems more straight forward to have information about debugging with the package itself.

PR #1242, opened 2019-10-21
Spotlight - basic fdc3 integration

fdore:
Spotlight now shows results from the fdc3 app directory, along with existing results
(we might not want to keep that long term, but it does get us started with fdc3)

This PR lets us open the blotter using the fdc3 api.

The app directory is hosted by the launcher, using the following configuration:
```
// launcher.local.json
{
  "name": "fdc3",
  "manifestUrl": "http://localhost:3000/config/openfin/local-fdc3.json"
},
```
```
// local-fdc3.json
"serviceConfiguration": {
  "applicationDirectory": "http://localhost:3000/config/openfin/local-app-directory.json"
}

```

### DIFF
--- a/src/client/src/apps/SpotlightRoute/Spotlight.tsx
+++ b/src/client/src/apps/SpotlightRoute/Spotlight.tsx
@@ -1,4 +1,12 @@
-import React, { ChangeEventHandler, FC, FocusEventHandler, KeyboardEventHandler, useEffect, useReducer } from 'react'
+import React, {
+  ChangeEventHandler,
+  FC,
+  FocusEventHandler,
+  KeyboardEventHandler,
+  useEffect,
+  useReducer,
+  useState,
+} from 'react'
 import { styled } from 'rt-theme'
 import { AdaptiveLoader } from 'rt-components'
 import { initialState, reducer } from './reducer'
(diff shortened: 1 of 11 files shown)


src/client/src/apps/SpotlightRoute/Spotlight.tsx
```diff
@@ -173,7 +207,17 @@ export const Spotlight: FC = () => {
         onChange={handleChange}
         onKeyDown={handleOnKeyDown}
       />
-      <Response>{contacting ? loader : suggestions}</Response>
+      <Response>
+        {contacting ? (
+          loader
+        ) : (
+          <>
+            {directoryAppSuggestions}
```
sizov: am I right that user here selects which of the registered apps is handling the intent? if so, I somehow assumed that it should not require user's action. If app is registered to handle the intent, then it just handles it silently without user action

fdore: Also, currently, there are two ways to open the blotter (using fdc3, and directly). The fdc3 way doesn't support filters yet.

sizov: ok, got it, I will then assume that it is for illustration only

fdore: That's right

PR #1244, opened 2019-10-21
show error message, if we fail to get a response from the server

fdore:


src/client/src/apps/SpotlightRoute/Spotlight.tsx
```diff
@@ -210,6 +210,8 @@ export const Spotlight: FC = () => {
       <Response>
         {contacting ? (
           loader
+        ) : response === null ? (
```
sizov: any chance you can break it down to not have two ternary operator next to each other?

sizov: just hard to read (for me personally)

PR #1247, opened 2019-10-22
Refactor price history service to use shared lib

fdore:
- Moved PriceHistory service to `node-server` directory
- Remove duplicated connection logic, and reference `shared` lib instead


src/node-server/PriceHistory.Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM node:alpine
```
ilikejames: Shouln't this `DockerFile` be within the `./price-history` package directory?

rikoe:
@ilikejames @fdore what should we do with this PR? I think you said there is a different way it should be done, @ilikejames?

rikoe:
Closing for now.

PR #1248, opened 2019-10-22
adding interop currency pair switch for currency pair route

sizov:


src/client/src/apps/MainRoute/widgets/spotTile/SpotTileContainer.tsx
```diff
@@ -27,7 +27,8 @@ export interface SpotTileContainerOwnProps {
 }
 
 const mapDispatchToProps = (dispatch: Dispatch, ownProps: SpotTileContainerOwnProps) => ({
-  onMount: () => dispatch(SpotTileActions.subscribeToSpotTile(ownProps.id)),
+  onCurrencyPairChanged: (currencyPair: string) =>
```
sizov: removed `onMount` in favour of `onCurrencyPairChanged` which is more explicit in saying what does it do

sizov: and also it covers both scenarios - component being mounted and changes in props

PR #1251, opened 2019-10-23
Move nlp dockerfile to nlp project

fdore:


ilikejames:
I know it was my original suggestion to have the `Dockerfile` within the project its self, but that was before I was unaware of the usage of the `shared` folder and how `docker build` would have to be run from within that parent context. 
Thinking about this further, I think perhaps it would be better to have all these nodejs services built within a single container, and be then reuse this base image with a different `RUN` command for each service. 
Going to close this in the meantime. I'll also revert the docker changes in the [Refactor Price History PR](https://github.com/AdaptiveConsulting/ReactiveTraderCloud/pull/1247).

PR #1256, opened 2019-10-24
Bugfix/artp 802: reset NZDUSD notional to 10m instead of 1m

rafgarciaa:
[Jira ticket](https://weareadaptive.atlassian.net/browse/ARTP-802)

Two functions existed in the `TileBusinessLogic.ts` file, one `getDefaultNotionalValue` that **_did not_** account for `NZDUSD`'s initial value and another one `getDefaultInitialNotionalValue ` that did account for `NZDUSD`'s initial value (10m).

I replaced `getDefaultNotionalValue` with `getDefaultInitialNotionalValue` to the `rfqReceivedEpic` so that when timer runs out, it can `setNotional` to the correct default values - accounting for `NZDUSD` currency pair.

### DIFF
--- a/src/client/src/apps/MainRoute/widgets/spotTile/epics/rfqEpics.ts
+++ b/src/client/src/apps/MainRoute/widgets/spotTile/epics/rfqEpics.ts
@@ -7,7 +7,7 @@ import { concat, from, Observable, of, timer } from 'rxjs'
 import { RfqReceived, RfqRequest } from '../model/rfqRequest'
 import { SpotTileState } from '../spotTileReducer'
 import { CurrencyPairState } from '../../../data/referenceData'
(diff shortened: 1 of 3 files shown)


src/client/src/apps/MainRoute/widgets/spotTile/epics/rfqEpics.ts
```diff
@@ -115,7 +115,7 @@ export const rfqReceivedEpic: ApplicationEpic = action$ =>
             from([
               setNotional({
                 currencyPair: currencyPair.symbol,
-                notional: getDefaultNotionalValue(),
+                notional: getDefaultInitialNotionalValue(currencyPair),
```
fdore: not sure we should dispatch `setTradingMode({mode: 'esp'})` for NZDUSD tho

PR #1265, opened 2019-11-06
feat: build and deploy using semver versioning

ilikejames:
Reworking of the build process to move towards _continous delivery_.

- All merges to master will create a new _semver_ minor  version (1.1.0, 1.2.0, 1.3.0...). 
- Master will be deployed to `dev` environment
- We will use the [circleapi](https://circleci.com/docs/2.0/env-vars/#injecting-environment-variables-with-the-api) (at this initial stage) to run the deployment job passing the environment variables `BUILD_TARGET=demo` `BUILD_VERSION=1.23.0`. 

There is a  significant reworking of how we were doing the build/run/test stages within ciricleci. This moves all the setup config to be done purely via docker-compose. Now, to run the server integration tests:
```
docker-compose -f docker-compose.e2e.yml run integration
```
This simplifies differences between local reproduction and from within CI. 
I've added the semver build version to the console log and to within the Connection Status footer (will check with UX).

Overall this will be a significant change in the release / patching process. This diagram will explain a little more the process. 

Group 1

Screenshot 2019-11-06 at 12 18 37

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,168 +1,223 @@
-version: 2
-jobs:
-  build_and_test:
-    machine:
-      enabled: true
+version: 2.1
+
+references:
+  push_images_list: &push_images_list
+    client pricehistory bot nlp servers broker eventstore nginx
+
+commands:
+  install_tools:
     steps:
-      - checkout
       - run:
-          name: 'Configure the environment'
+          name: Install tools
           command: |
-            # Inject a dummy user for builds triggered from forks, because secure variables are not available there.
-            # This allows the job to proceed with building and testing docker images.
-            if [[ -n "${CIRCLE_PR_REPONAME}" ]] ; then
-              echo 'export DOCKER_USER=adaptive' >> $BASH_ENV
-            fi
+            sudo npm i -g semver@6.3.0
       - run:
-          name: Create network
-          command: 'docker network create -d bridge circleci'
-      - run:
-          name: Build client
-          command: 'cd ./src/client  && docker build -t ${DOCKER_USER}/client:${CIRCLE_BUILD_NUM} .'
-      - run:
-          name: Build servers
-          command: 'cd ./src/server && docker build -t ${DOCKER_USER}/servers:${CIRCLE_BUILD_NUM} .'
-      - run:
-          name: Build nginx
-          command: 'cd ./src/services/nginx  && docker build -t ${DOCKER_USER}/nginx:${CIRCLE_BUILD_NUM} .'
-      - run:
-          name: Build broker
-          command: 'cd ./src/services/broker  && docker build -t ${DOCKER_USER}/broker:${CIRCLE_BUILD_NUM} .'
-      - run:
-          name: Build eventstore
-          command: 'cd ./src/services/eventstore  && docker build -t ${DOCKER_USER}/baseeventstore .'
-      - run:
-          name: Build pricehistory
-          command: 'cd ./src/server/Adaptive.ReactiveTrader.Server.PriceHistory/ && docker build -t ${DOCKER_USER}/pricehistory:${CIRCLE_BUILD_NUM} .'
-      - run:
-          name: Build bot
-          command: 'cd ./src/server/Adaptive.ReactiveTrader.Server.Bot/ && docker build -t ${DOCKER_USER}/bot:${CIRCLE_BUILD_NUM} .'
-      - run:
-          name: Build nlp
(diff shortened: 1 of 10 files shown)


src/client/Dockerfile
```diff
@@ -1,8 +1,17 @@
...
+ARG REACT_APP_BUILD_VERSION
+
 COPY ./package*.json ./
 RUN npm install
 COPY . .
+
+ENV CI=$CI
+ENV REACT_APP_BUILD_VERSION=$REACT_APP_BUILD_VERSION
+
+RUN npm run test -- --coverage
```
ilikejames: Client unit tests run within the container build stage. Its one of those _better practices_

rikoe:
Is this right: "master will be deployed to the dev environment"? Can you elaborate?

ilikejames:
> Is this right: "master will be deployed to the dev environment"? Can you elaborate?

The idea is ditch `develop`. All PRs will then target `master` or `release/1.23.0` in the case of a patch of an existing deployed version. After a master merge, a successful build will then trigger an automatic deployment to `dev` environment. With `dev` now acting as a proper continuous delivery point.
When it comes to a round of regression tests (we now have e2e tests to be added from this [PR](https://github.com/AdaptiveConsulting/ReactiveTraderCloud/pull/1241)) "someone" will use the circleci api to run the deploy job to move e.g. `1.23.0` -> `env-uat` (new environment).
I'm updating the documentation right now!

rikoe:
What does this mean for the current setup where master is more stable/updated less frequently.

The reason I ask is that OpenFin uses master/demo for their client demos and that’s a good thing because PRs don’t affect the version they use until we do a planned promotion.

We are meeting with them today for another demo walkthrough and we cannot afford right now to take all the latest develop changes into the “demo” deployment (which matches up to master).

This is a long-winded way of asking what the impact of merging this PR is on the current demo environment.

src/client/src/apps/MainRoute/widgets/footer-version/FooterVersion.tsx
```diff
@@ -0,0 +1,12 @@
+import React, { FC } from 'react'
+import { styled } from 'rt-theme'
+
+export const Wrapper = styled.div`
+  color: ${props => props.theme.textColor};
+  opacity: 0.59;
+  font-size: 0.75rem;
+`
+
+const FooterVersion: FC = () => <Wrapper>v{process.env.REACT_APP_BUILD_VERSION}</Wrapper>
```
sizov: 👍

src/docker-compose.deploy.yml
```diff
@@ -0,0 +1,10 @@
+version: '3'
+services:
+  servers:
+    image: ${DOCKER_USER}/servers:${BUILD_VERSION}
+    build:
+      dockerfile: Dockerfile
+      context: ./server
+    depends_on:
+      - eventstore
+    command: 'echo "Override based docker-compose to not populate event store"'
```
sizov: based?

src/client/Dockerfile
keithbloom: Very nice. If the build fails does it stop the process and put the build report somewhere?

.circleci/config.yml
```diff
@@ -1,168 +1,223 @@
...
             echo $GCLOUD_SERVICE_KEY | gcloud auth activate-service-account --key-file=-
             gcloud config set project ${GOOGLE_PROJECT_ID}
             gcloud config set compute/zone ${GOOGLE_COMPUTE_ZONE}
             gcloud container clusters get-credentials cluster
       - run:
-          name: Set namespace
-          command: declare -A envs=( ["develop"]="dev" ["master"]="demo"); echo "export NAMESPACE='${envs[$CIRCLE_BRANCH]}'" >> $BASH_ENV;
+          name: Updating deployment definition to new images
+          command: |
+            # TODO: looking into looking for new deployment config from git history
```
keithbloom: MINOR: typo `looking into looking`

ilikejames:
@rikoe quite right about openfin. I think that was the original motivator for change. The fact that openfin should be extra stable.
This change is about being able to have a constant delivery of versions. The versions will be deployed to `dev` for an initial check. And versions can be deployed to other environments manually. we are going to create an initial new environment `uat` for testing, then only on high confidence will something be moved to `demo`. I expect shortly we'd be creating another new environment (I have a backlog items to be able to create an environment on demand) which will be dedicated for openfin, as I understand they are using it for demos of FDC3 (?).

src/client/Dockerfile
ilikejames: If the build fails, it will be within the `Build images` stage log. It will be the last item in the logs showing the failure. 
As the steps are all within the `DockerFile`, it should be trivial to reproduce by running the `docker-compose build client` step. I like things being in reproducible docker steps.

PR #1271, opened 2019-11-08
feat: smoke tests

ilikejames:
I've moved the existing e2e tests into its own package. Felt cleaner and will mean we can start up the docker container / tear it down during build & deploy.

Lots of changes due to:
- added eslint config which meant a lot of type issues
- some flaky tests needed resolving

Features:
- Runs in circleci
- Improved setup for debugging. 

image


src/e2e/src/browser-manager.ts
```diff
@@ -0,0 +1,32 @@
+import { MainPage } from './pages/main.page';
```
ilikejames: Most of these files are identical to the current. I don't know what these are not listed as moved and instead deleted :/

src/e2e/src/browser-manager.ts
```diff
@@ -0,0 +1,32 @@
...
+
+async function prepareBrowser(): Promise<ProtractorBrowser> {
+  // We are testing a React project. We need to disable Angular sync.
+  await browser.waitForAngularEnabled(false)
+
+  console.info('Getting page: ', browser.baseUrl);
+  await browser.get(browser.baseUrl, 10000)
+
+  // Wait for application to be up and connected to services
+  await waitForConnected(browser)
```
ilikejames: Here is a change. Starting the test session now waits for the application to be connected to the broker.

src/e2e/src/protractor.config.js
```diff
@@ -1,13 +1,21 @@
...
 
 const CHROME_CAPABILITIES = {
   browserName: 'chrome',
   chromeOptions: {
-    args: ['--start-maximized', '--disable-infobars', '--disable-notifications', '--headless', '--no-sandbox']
+    args: ['--start-maximized', '--disable-infobars', '--disable-notifications', '--no-sandbox']
+  }
+}
+
+const CHROME_CAPABILITIES_HEADLESS = {
```
ilikejames: Added a headless and a standard *visible* chrome option to help with debugging.

src/e2e/src/utils/assert.utils.ts
```diff
@@ -1,64 +1,57 @@
...
 
 let mainPage: MainPage
 let browser: ProtractorBrowser
 
 export async function confirmationMessageAsserts(currencies: string, transaction: string, expectedResult: string, notional: string, timeout: boolean) {
   browser = await getBrowser()
   mainPage = new MainPage(browser)
-  let tradeWord: string
-  let parsedTradeText: string
+  const tradeWord: string = transaction === 'buy' ? 'bought' : 'sold';
```
ilikejames: This method was a tad *racey*. Reworked it a little but kept the overall assertions. This could be better thought out.

PR #1282, opened 2019-11-11
Improved ExecuteTrade integration tests

PiRXlv:
Also added *.cs indent of 4 spaces in .editorconfig as that's default spacing in C# world

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -28,3 +28,6 @@ trim_trailing_whitespace = true
 # Denotes whether file should end with a newline
 # Possible values - true, false
 insert_final_newline = true
+
+[*.cs]
+indent_size = 4
--- a/src/server/Adaptive.ReactiveTrader.Server.IntegrationTests/ExecuteTradeIntegrationTests.cs
+++ b/src/server/Adaptive.ReactiveTrader.Server.IntegrationTests/ExecuteTradeIntegrationTests.cs
@@ -1,7 +1,6 @@
-﻿using System;
+using System;
 using System.Reactive.Linq;
 using System.Reactive.Threading.Tasks;
-using System.Threading;
 using System.Threading.Tasks;
 using Adaptive.ReactiveTrader.Common;
 using Adaptive.ReactiveTrader.Contract;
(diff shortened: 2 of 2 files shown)


.editorconfig
```diff
@@ -28,3 +28,6 @@ trim_trailing_whitespace = true
 # Denotes whether file should end with a newline
 # Possible values - true, false
 insert_final_newline = true
+
+[*.cs]
+indent_size = 4
```
rikoe: 👍

src/server/Adaptive.ReactiveTrader.Server.IntegrationTests/ExecuteTradeIntegrationTests.cs
```diff
@@ -33,30 +29,18 @@ public ExecuteTradeIntegrationTests()
...
-            _executionServiceInstance = _heartbeatStream
-                .Where(hb => hb.Type == ServiceTypes.Execution)
-                .Select(hb => hb.Instance)
-                .Take(1)
-                .Timeout(ResponseTimeout)
-                .ToTask()
-                .Result;
-
-            _timeoutCancellationTokenSource = new CancellationTokenSource();
+            _executionServiceInstance = GetInstanceName(ServiceTypes.Execution).Result;
```
jonclare: I know it used Task.Result before, but it would be nice if we could make this async.
I wonder if we can do it outside of the constructor maybe?

src/server/Adaptive.ReactiveTrader.Server.IntegrationTests/ExecuteTradeIntegrationTests.cs
```diff
@@ -66,44 +50,35 @@ public async void ShouldReceiveExecutedTradeInBlotter()
...
-            _analyticsServiceInstance = await _heartbeatStream
-                .Where(hb => hb.Type == ServiceTypes.Analytics)
-                .Select(hb => hb.Instance)
-                .Take(1);
+            var asyncAssertion = new TaskCompletionSource<bool>();
+            var analyticsServiceInstance = await GetInstanceName(ServiceTypes.Analytics);
 
             // this is the callback when the analytics svc receives the executed trade notification
-            Action<dynamic> analyticsCallbackAssertion = d =>
+            void AnalyticsCallbackAssertion(dynamic d)
```
jonclare: Anonymous methods FTW!

src/server/Adaptive.ReactiveTrader.Server.IntegrationTests/ExecuteTradeIntegrationTests.cs
PiRXlv: I think that this test could benefit from bigger refactoring, but didn't want to get carried away

PR #1283, opened 2019-11-11
build: create/update kubernetes deployment configuration

ilikejames:
Currently, adding a new deployment is a manual step.
This changes the process slightly. This build step:
- finds all matching `*-deployment.yaml` 
- does a variable substitution. The BUILD_NUMBER is the environment variable that will be replaced in the deploys.
- the `kubectl apply` command will create the deployment if it doesn't exist yet.


rikoe:
@ilikejames - if approved, is this ready to merge?

PR #1284, opened 2019-11-11
Fixed for notional display

sizov:
- fixed the regression related to the default value of the notional
- reworked they way how the default values of `SpotTileData` are propagated through the data hierarchy (this is now controlled by selectors, not views)

see more comments inside the PR

### DIFF
--- a/src/client/src/apps/MainRoute/widgets/spotTile/SpotTileContainer.tsx
+++ b/src/client/src/apps/MainRoute/widgets/spotTile/SpotTileContainer.tsx
@@ -50,8 +50,10 @@ const mapDispatchToProps = (dispatch: Dispatch, ownProps: SpotTileContainerOwnPr
 const makeMapStateToProps = () => (state: GlobalState, ownProps: SpotTileContainerOwnProps) => ({
   pricingStatus: selectPricingStatus(state),
   executionStatus: selectExecutionStatus(state),
-  currencyPair: selectCurrencyPair(state, ownProps),
-  spotTileData: selectSpotTileData(state, ownProps),
+
+  // here 'ownProps.id' is an ID of the tile, but it's ID of the currency pair too (same thing for now)
+  currencyPair: selectCurrencyPair(state, ownProps.id),
+  spotTileData: selectSpotTileData(state, ownProps.id),
 })
 
 type SpotTileContainerDispatchProps = ReturnType<typeof mapDispatchToProps>
--- a/src/client/src/apps/MainRoute/widgets/spotTile/components/SpotTile.tsx
+++ b/src/client/src/apps/MainRoute/widgets/spotTile/components/SpotTile.tsx
@@ -3,7 +3,7 @@ import { spotDateFormatter } from '../model/dateUtils'
 import NotionalInput from './notional'
 import PriceControls from './PriceControls'
 import TileHeader from './TileHeader'
-import { getDefaultInitialNotionalValue } from './Tile/TileBusinessLogic'
+import { getDefaultNotionalValue } from './Tile/TileBusinessLogic'
 import {
   NotionalInputWrapper,
   SpotTileWrapper,
@@ -49,7 +49,7 @@ export default class SpotTile extends PureComponent<SpotTileProps> {
     } = getConstsFromRfqState(rfqState)
     const showResetButton =
       !isTradeExecutionInFlight &&
-      getDefaultInitialNotionalValue(currencyPair) !== notional &&
+      getDefaultNotionalValue(currencyPair) !== notional &&
       (isRfqStateNone || isRfqStateCanRequest || isRfqStateExpired)
     const showTimer = isRfqStateReceived && rfqTimeout
     const priceData = (isRfqStateReceived || isRfqStateExpired) && rfqPrice ? rfqPrice : price
--- a/src/client/src/apps/MainRoute/widgets/spotTile/components/Tile/Tile.tsx
+++ b/src/client/src/apps/MainRoute/widgets/spotTile/components/Tile/Tile.tsx
@@ -1,13 +1,13 @@
 import React from 'react'
 import { CurrencyPair, Direction, ServiceConnectionStatus } from 'rt-types'
-import { createTradeRequest, ExecuteTradeRequest, SpotTileData, TradeRequest } from '../../model'
+import { createTradeRequest, ExecuteTradeRequest, SpotTileDataWithNotional, TradeRequest } from '../../model'
 import SpotTile from '../SpotTile'
 import { AnalyticsTile } from '../analyticsTile/index'
 import { TileViews } from '../../../workspace/workspaceHeader'
 import { RfqActions, TileSwitchChildrenProps, TradingMode } from '../types'
 import { ValidationMessage } from '../notional'
 import {
-  getDefaultInitialNotionalValue,
+  getDefaultNotionalValue,
   getDerivedStateFromProps,
   getDerivedStateFromUserInput,
   isValueInRfqRange,
@@ -17,7 +17,7 @@ import { CurrencyPairNotional } from '../../model/spotTileData'
 
 export interface TileProps {
   currencyPair: CurrencyPair
-  spotTileData: SpotTileData
+  spotTileData: SpotTileDataWithNotional
   executionStatus: ServiceConnectionStatus
   executeTrade: (tradeRequestObj: ExecuteTradeRequest) => void
   setTradingMode: (tradingMode: TradingMode) => void
@@ -110,9 +110,11 @@ class Tile extends React.PureComponent<TileProps, TileState> {
     })
   }
 
+  // TODO: it looks like this should be moved to a reducer or epic. There should be an
+  // action called ResetNotional and all logic below should be there
   resetNotional = () => {
     const { setTradingMode, currencyPair, updateNotional } = this.props
-    const defaultNotional = getDefaultInitialNotionalValue(currencyPair)
+    const defaultNotional = getDefaultNotionalValue(currencyPair)
     const isInRfqRange = isValueInRfqRange(defaultNotional)
     setTradingMode({ symbol: currencyPair.symbol, mode: isInRfqRange ? 'rfq' : 'esp' })
     updateNotional({
--- a/src/client/src/apps/MainRoute/widgets/spotTile/components/TileSwitch.tsx
+++ b/src/client/src/apps/MainRoute/widgets/spotTile/components/TileSwitch.tsx
@@ -1,7 +1,7 @@
 import React from 'react'
 import { CurrencyPair, ServiceConnectionStatus } from 'rt-types'
 import { ExecuteTradeRequest, SpotTileDataWithNotional } from '../model'
-import { PriceMovementTypes } from '../model/priceMovementTypes'
+// import { PriceMovementTypes } from '../model/priceMovementTypes'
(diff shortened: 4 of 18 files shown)


src/client/src/apps/MainRoute/widgets/spotTile/SpotTileContainer.tsx
```diff
@@ -50,8 +50,10 @@ const mapDispatchToProps = (dispatch: Dispatch, ownProps: SpotTileContainerOwnPr
 const makeMapStateToProps = () => (state: GlobalState, ownProps: SpotTileContainerOwnProps) => ({
   pricingStatus: selectPricingStatus(state),
   executionStatus: selectExecutionStatus(state),
-  currencyPair: selectCurrencyPair(state, ownProps),
-  spotTileData: selectSpotTileData(state, ownProps),
+
+  // here 'ownProps.id' is an ID of the tile, but it's ID of the currency pair too (same thing for now)
+  currencyPair: selectCurrencyPair(state, ownProps.id),
```
sizov: notice the change in the signature of `selectSpotTileData` selectors - it now receives simple ID instead of props of component - this makes it more reusable in the future

src/client/src/apps/MainRoute/widgets/spotTile/components/SpotTile.tsx
```diff
@@ -3,7 +3,7 @@ import { spotDateFormatter } from '../model/dateUtils'
 import NotionalInput from './notional'
 import PriceControls from './PriceControls'
 import TileHeader from './TileHeader'
-import { getDefaultInitialNotionalValue } from './Tile/TileBusinessLogic'
+import { getDefaultNotionalValue } from './Tile/TileBusinessLogic'
```
sizov: a simple name change to remove redundant term (initial == default in this context?)

src/client/src/apps/MainRoute/widgets/spotTile/components/Tile/Tile.tsx
```diff
@@ -1,13 +1,13 @@
 import React from 'react'
 import { CurrencyPair, Direction, ServiceConnectionStatus } from 'rt-types'
-import { createTradeRequest, ExecuteTradeRequest, SpotTileData, TradeRequest } from '../../model'
+import { createTradeRequest, ExecuteTradeRequest, SpotTileDataWithNotional, TradeRequest } from '../../model'
```
sizov: name change since we now have two types - `SpotTileData` and `SpotTileDataWithNotional` (maybe this is a horrible name). Bottom line - `SpotTileData` is in store with optional notional, but there's a selector that always adds `notional` into existing `SpotTileData`. the result is `SpotTileDataWithNotional`. This has been done because we don't know notional when we create `SpotTileData` - default `notional` fill change from currency to currency. Thus we are adding a default `notional` in the selector (we add it based on currency).

I can't say I like this approach - I would prefer to do this in a reducer. But it would have been much bigger change - resetting reference data would need to create new tiles in reducer, then new tiles will need to reset all `tileData` slice. I'd like to have that and left some TODOs there

src/client/src/apps/MainRoute/widgets/spotTile/components/TileSwitch.tsx
```diff
@@ -47,18 +45,10 @@ const TileSwitch: React.FC<Props> = ({
     isRfqStateNone,
   } = getConstsFromRfqState(spotTileData.rfqState)
 
-  const spotTileDataWithNotional =
```
sizov: this logic has been moved to a selector (see `widgets/spotTile/selectors.ts`). The problem of having this data modification in the view is that we would need to duplicate this code if any other view needs this data slice too

src/client/src/apps/MainRoute/widgets/spotTile/components/TileSwitch.tsx
```diff
@@ -111,35 +101,4 @@ const TileSwitch: React.FC<Props> = ({
   )
 }
 
-TileSwitch.defaultProps = {
```
sizov: moved to `widgets/spotTile/selectors.ts`

src/client/src/apps/MainRoute/widgets/spotTile/epics/closePositionEpic.ts
```diff
@@ -43,19 +48,18 @@ export const closePositionEpic: ApplicationEpic = (action$, state$, { platform }
     applicationConnected,
     switchMapTo(interopSubscribe$),
     withLatestFrom(state$),
+    filter(([message, state]) => !!getSpotTilesDataFromMessage(message, state)),
     map(([message, state]) => {
-      const [msg, uuid] = message as Message
+      const [msg, uuid] = message
       const trade = createTrade(
         msg,
-        state.spotTilesData[msg.symbol],
```
sizov: types were updated to represent the case where we might not get spotTilesData for certain currencies. So now we need to handle unhappy path with filter

src/client/src/apps/MainRoute/widgets/spotTile/epics/rfqEpics.ts
```diff
@@ -40,36 +41,44 @@ export const IDLE_TIME_MS = 60000
 const rfqService = (
   request: RfqRequest,
   currencyPairs: CurrencyPairState,
-  spotTilesData: SpotTileState,
+  spotTilesData: SpotTileData,
```
sizov: type has changed since we validate spotTileData upstream now

src/client/src/apps/MainRoute/widgets/spotTile/epics/rfqEpics.ts
```diff
@@ -87,7 +96,7 @@ export const rfqRequestEpic: ApplicationEpic<{}> = (action$, state$) =>
       return rfqService(
         action.payload,
         state$.value.currencyPairs,
-        state$.value.spotTilesData,
```
sizov: types were updated to represent the case where we might not get spotTilesData for certain currencies. So now we need to handle unhappy path with filter

src/client/src/apps/MainRoute/widgets/spotTile/model/spotTileData.ts
```diff
@@ -20,9 +20,12 @@ export interface SpotTileData {
   rfqReceivedTime: number | null
   rfqTimeout: number | null
   rfqPrice: SpotPriceTick | null
-  notional: number
+  notional?: number
 }
 
+type SpotTileDataNotional = Pick<SpotTileData, 'notional'>
+export type SpotTileDataWithNotional = SpotTileData & Required<SpotTileDataNotional>
```
sizov: see [this](https://github.com/AdaptiveConsulting/ReactiveTraderCloud/pull/1284/files#r344925453) explanation

src/client/src/apps/MainRoute/widgets/spotTile/selectors.test.ts
```diff
@@ -0,0 +1,107 @@
+import { GlobalState } from 'StoreTypes'
```
sizov: i hope this test is self-explanatory

src/client/src/apps/MainRoute/widgets/spotTile/selectors.ts
```diff
@@ -1,19 +1,59 @@
...
 import { GlobalState } from 'StoreTypes'
-import { SpotTileContainerOwnProps } from './SpotTileContainer'
+import { getDefaultNotionalValue } from './components/Tile/TileBusinessLogic';
+import { SpotTileData, SpotTileDataWithNotional } from './model';
+import { CurrencyPair } from 'rt-types';
+import { PriceMovementTypes } from './model/priceMovementTypes';
 
-const getCurrencyPair = (state: GlobalState, props: SpotTileContainerOwnProps) =>
-  state.currencyPairs[props.id]
+const DEFAULT_TILE_DATA: SpotTileData = {
```
sizov: these defaults were lifted from view - this is to make sure that we would not duplicate such code across multiple views

src/client/src/apps/MainRoute/widgets/spotTile/selectors.ts
```diff
@@ -1,19 +1,59 @@
...
-const getSpotTileData = (state: GlobalState, props: SpotTileContainerOwnProps) =>
-  state.spotTilesData[props.id]
+function getSpotTileDataWithNotional(spotTileData: SpotTileData | undefined, currencyPair: CurrencyPair): SpotTileDataWithNotional {
+  // TODO: instead of creating items in the selector, consider creating them in the
+  // reducer (when tiles are created, which in turn happens when reference data is created)
+  if (!spotTileData) {
+    spotTileData = DEFAULT_TILE_DATA
+  }
+
+  const validNotional = typeof spotTileData.notional === 'undefined' ?
```
sizov: this is where we make sure that we have a default notional or amended one

src/client/src/apps/MainRoute/widgets/spotTile/selectors.ts
```diff
@@ -1,19 +1,59 @@
...
+  };
+}
+
+const getSpotTileData = (state: GlobalState, currencyPairId: string) =>
+  state.spotTilesData[currencyPairId]
 const selectSpotTileData = createSelector(
-  [getSpotTileData],
-  spotTileData => spotTileData,
+  [getSpotTileData, getCurrencyPair],
+  (spotTileData: SpotTileData | undefined, currencyPair: CurrencyPair) => getSpotTileDataWithNotional(spotTileData, currencyPair)
```
sizov: combining two selector functions

src/client/src/apps/MainRoute/widgets/spotTile/spotTileDataReducer.ts
```diff
@@ -3,9 +3,8 @@ import { SpotTileActions, TILE_ACTION_TYPES } from './actions'
 import { PriceMovementTypes } from './model/priceMovementTypes'
 import { SpotTileData } from './model/spotTileData'
 
-export interface SpotTileState {
-  [currencyPair: string]: SpotTileData
-}
+// we want to let compiler know that values for certain keys might be missing
+export type  SpotTileState = { [currencyPair: string]: SpotTileData | undefined }
```
sizov: this is where `| undefined` was added to point out that there might be missing `SpotTileData ` for certain currency

src/client/src/apps/MainRoute/widgets/spotTile/components/Tile/Tile.tsx
keithbloom: Can you paste this explanation in to the `spotTile.ts` file

src/client/src/apps/MainRoute/widgets/spotTile/spotTileDataReducer.ts
keithbloom: minor: This could be defined using the Record<> type

```
export type  SpotTileState = Record<string, SpotTileData | undefined>
```

src/client/src/apps/MainRoute/widgets/spotTile/components/Tile/Tile.tsx
sizov: done

src/client/src/apps/MainRoute/widgets/spotTile/spotTileDataReducer.ts
sizov: done

PR #1285, opened 2019-11-12
Split IServiceHostFactory to remove no-op implementations

PiRXlv:


ilikejames:
We implemented a git hook to verify that commit messages use the conventional commits approach, but i guess you are `--no-verify`-ing ;)
Can you amend the commits, something like...

```
refactor(servers): Split IServiceHostFactory to remove no-op implementations
```

...would be nice.

rikoe:
@ilikejames the issue you mention - is it related to being from a forked PR? Our process should work for forks, if it is.

PR #1288, opened 2019-11-12
fix(installer): use tags for openfin installer configuration

ilikejames:
Not 100% loving this solution but its based on the existing process. 
Perhaps in future we could have the demo installer downloaded from the demo site. And the dev installer downloaded from the dev site etc. 

This change:
 - ensures the master build commit is tagged with `env-dev`.
 - updates the openfin config to use these magic tags rather than branches.
 - adds error handling to the generation of installers. 

Currently openfin are having difficulties generating osx installers:

`Something went wrong. Please contact support@openfin.co`

image

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -173,6 +173,18 @@ jobs:
           environment: dev
       - push_images:
           images: *push_images_list
+      - run:
+          name: Make new env-dev version
+          command: |
+            # Tag the new version with env-dev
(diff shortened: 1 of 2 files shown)


ilikejames:
Thinking about this further, I'm not sure this is the best approach. 
We are not enforcing that there should be these tags present, so there can be the occasion where a user opens the app and there is no available path to config.

rikoe:
I am going to decline this PR and just point the installers directly at master for now.

PR #1289, opened 2019-11-14
Fix: ARTP-852 - Point OpenFin dev installers to master branch

rikoe:
* Change installer script to properly print extension while generating
* Do not try to generate Mac installer for launcher, which is Windows-only
* Removed .DS_Store file inadvertently checked in despite being in .gitignore file


src/client/lib/downloadInstaller.js
```diff
@@ -37,21 +37,21 @@ Make sure the files are available on their respective locations when distributin
...
   return Promise.all(installerDownloads)
 }
 
 // File name + github branch for each json manifest
 const INSTALLERS_TO_CREATE = [
-  { manifestName: 'dev', branch: 'develop' },
-  { manifestName: 'launcher-dev', branch: 'develop' },
-  { manifestName: 'demo', branch: 'master' },
-  { manifestName: 'launcher-demo', branch: 'master' },
+  { manifestName: 'dev', branch: 'master', os: 'win' },
```
sizov: is it a bit odd that we crate two installers (dev AND demo) from the same branch ?

sizov: i thought that dev and demo relate to the branches..but maybe they are not :/ confused

sizov: ah, everything is clear now (after looking at actual manifests)

PR #1290, opened 2019-11-14
fix: updates caching policy in nginx, disables service worker (ARTP-854)

keithbloom:
The client is not always retrieving the most recent version. This change fixes this by:
- Configuring the caching policy in nginx
- Disabling the service worker

### DIFF
--- a/src/client/src/index.tsx
+++ b/src/client/src/index.tsx
@@ -39,7 +39,7 @@ async function init() {
     // If you want your app to work offline and load faster, you can change
     // unregister() to register() below. Note this comes with some pitfalls.
     // Learn more about service workers: https://bit.ly/CRA-PWA
-    serviceWorker.register()
+    serviceWorker.unregister()
   }
 }
 
--- a/src/services/nginx/default.conf
+++ b/src/services/nginx/default.conf
@@ -6,8 +6,16 @@ upstream broker {
     server broker:8000;
 }
 
+map $sent_http_content_type $expires {
+    default                    off;
+    text/html                  epoch;
+    text/css                   max;
+    application/javascript     max;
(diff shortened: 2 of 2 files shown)


src/services/nginx/default.conf
```diff
@@ -6,8 +6,16 @@ upstream broker {
     server broker:8000;
 }
 
+map $sent_http_content_type $expires {
+    default                    off;
+    text/html                  epoch;
+    text/css                   max;
+    application/javascript     max;
+}
```
keithbloom: These settings ensure that the main `html` page has `no-cache` set which forces the browser to check the `etag` and `Last-Modified` fields. 

For `css` and `javascript` we set the expiry to some distant time in the future as they are immutable so the browser can keep them for ever.

src/client/src/index.tsx
```diff
@@ -39,7 +39,7 @@ async function init() {
     // If you want your app to work offline and load faster, you can change
     // unregister() to register() below. Note this comes with some pitfalls.
     // Learn more about service workers: https://bit.ly/CRA-PWA
-    serviceWorker.register()
+    serviceWorker.unregister()
```
keithbloom: The service worker is a nice optimization but it does add another layer of complexity to the caching policy. In my limited tests it appeared that it cached the `index.html` file which is bad as it would never get updated. So instead of digging into the configuration I chose to remove it as we have a small number of assets to cache.

sizov: since it is disabled, should we remove this code completely?

keithbloom: The call to `unregister()` will remove the service worker from any clients that have it installed.

PR #1306, opened 2019-11-21
Add UI structure documentation

rikoe:
Doing PR for @rafgarciaa's changes after rebasing.


sizov:
question: what do we need this documentation for? for me it looks like a recipe for having outdated docs in a few months. I don't think anyone would remember to change docs when they do refactoring in any of the components that are described in docs

sizov:
I would remove any mentions of the components from these docs and leave only general principles related to business functionality.

rikoe:
@sizov are you okay with merging this, or do you think we should make changes/leave it unmerged?

PR #1311, opened 2019-11-25
fix: fixing tear-off appearance (ARTP-825)

sizov:
Fixing an issue where old tear-off were closing when new ones were created. This was due to react hooks re-executing themselves in places where we don't expect them to do so

### DIFF
--- a/src/client/src/rt-components/tear-off/ExternalWindow.tsx
+++ b/src/client/src/rt-components/tear-off/ExternalWindow.tsx
@@ -55,7 +55,8 @@ const ExternalWindow: FC<ExternalWindowProps> = ({
         externalWindow.close()
       }
     }
-  }, [config, onUnload, platform.window])
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [])
 
   return null
 }
--- a/src/client/src/rt-platforms/openFin/adapter/window.ts
+++ b/src/client/src/rt-platforms/openFin/adapter/window.ts
@@ -99,12 +99,15 @@ export const openDesktopWindow = async (
     }
   }
 
+  const windowName = config.name || generateRandomName();
+  console.info(`Creating Openfin window: ${windowName}`);
+
   //TODO: move to openfin V2 version (based on promises) once they fix their bug related to getting current window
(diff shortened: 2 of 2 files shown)


src/client/src/rt-components/tear-off/ExternalWindow.tsx
```diff
@@ -55,7 +55,8 @@ const ExternalWindow: FC<ExternalWindowProps> = ({
         externalWindow.close()
       }
     }
-  }, [config, onUnload, platform.window])
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
sizov: any change to `config, onUnload, platform.window` was executing script which was closing the window

sizov: I don't like this fix since we should chose a different strategy rather than disabling TS rules. But this is how it was before, let's leave it until we come up with a better approach

src/client/src/rt-platforms/openFin/adapter/window.ts
```diff
@@ -99,12 +99,15 @@ export const openDesktopWindow = async (
     }
   }
 
+  const windowName = config.name || generateRandomName();
+  console.info(`Creating Openfin window: ${windowName}`);
+
   //TODO: move to openfin V2 version (based on promises) once they fix their bug related to getting current window
   // (in V2 call to ofWindow.getWebWindow() returns undefined - thus we are forced to use old callback APIs)
```
rikoe: do we need to follow this up with OpenFin @sizov ? Have you spoken to them about it at all?

sizov: I haven't spoken about it. How would it be better to reach them out on this?

src/client/src/rt-components/tear-off/ExternalWindow.tsx
keithbloom: I agree that we should find a way of removing the tslint exception. It seems bad that calling `externalWindow.close()` here can close a different window.

sizov:
> I agree that we should find a way of removing the tslint exception. It seems bad that calling externalWindow.close() here can close a different window.

it is actually closing not because of some other window was closed. This is because new props were passed into component and react hook's unsubscription function closed it

PR #1314, opened 2019-11-27
fix: handles service disconnections

keithbloom:
This change fixes a bug in handling service disconnections whereby the client fails to re-subscribe to a service after a network disconnection. This is what causes the pricing service to show the "pricing unavailable" message as the prices are now stale.

### DIFF
--- a/src/client/src/rt-system/ServiceStubWithLoadBalancer.ts
+++ b/src/client/src/rt-system/ServiceStubWithLoadBalancer.ts
@@ -105,15 +105,17 @@ export default class ServiceStubWithLoadBalancer {
               },
             })
 
-            // There must be a way to do this without an inner-subsribe
+            // There must be a way to do this without an inner-subscribe
             const subscription = subscribeTopic$.subscribe(obs)
 
+            // This will detect when a service has gone down an emit an error on to stream. This allows any subscribers
+            // to handle the timeout and `retryWhen` would be good option
             const detectInstanceTimout = this.serviceInstanceDictionaryStream
               .pipe(
                 map(currentStatus => currentStatus.getServiceInstanceStatus(service, serviceInstanceStatus.serviceId)),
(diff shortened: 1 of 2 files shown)


src/client/src/rt-system/ServiceStubWithLoadBalancer.ts
```diff
@@ -105,15 +105,25 @@ export default class ServiceStubWithLoadBalancer {
               },
             })
 
-            // There must be a way to do this without an inner-subsribe
+            // There must be a way to do this without an inner-subscribe
             const subscription = subscribeTopic$.subscribe(obs)
 
+            // This will detect when the service has stopped sending status updates and force an error on the stream
+            // Forcing the error will cause the anyone using this to unsubscribe and then re-subscribe when the service
```
rikoe: minor: surplus "the"

src/client/src/rt-system/ServiceStubWithLoadBalancer.ts
```diff
@@ -105,15 +105,25 @@ export default class ServiceStubWithLoadBalancer {
...
+            // service were unaware of the change of status and showed only stale data.
+            //
+            // We need a fix that handles the service disconnections separately from network disconnections
             const detectInstanceTimout = this.serviceInstanceDictionaryStream
               .pipe(
                 map(currentStatus => currentStatus.getServiceInstanceStatus(service, serviceInstanceStatus.serviceId)),
                 filter(currentStatus => !(currentStatus && currentStatus.isConnected)),
               )
-              .subscribe(() => console.log('Topic timed out: ' + topicName))
+              .subscribe(() => obs.error('Topic timed out: ' + topicName))
```
sizov: any chance to write a test for it

sizov: > error will cause the anyone using this to unsubscribe and then re-subscribe when the service is available again.

do they need to do anything new? are they already re-subscribing?

PR #1315, opened 2019-11-28
feat: change colour of the unavailable price tile state (ARTP-840)

keithbloom:
This adds a third state to `TileNotification` so it can now display `success`, `info` and `warning`. The `info` state is used to display the unavailable prices message uses the background colour from the disconnected state.

Update
-------

I found that we have a style-guide for ReactiveTrader (http://web-dev.adaptivecluster.com/styleguide) which includes guidance on Accents & Functional Colours. Using this I have now changed the background to use the `aware` colour

image

This does mean that the colours for the success and failure have changed slightly as they are using this new palette (they are still green and red)

image

image


src/client/src/apps/MainRoute/widgets/spotTile/components/notifications/TileNotification.tsx
```diff
@@ -2,7 +2,7 @@ import React from 'react'
 import { styled } from 'rt-theme'
 import { Button, Icon, TileBaseStyle } from '../styled'
 
-type AccentColor = 'red' | 'green'
+type AccentColor = 'red' | 'green' | 'grey'
```
rikoe: What I am wondering is whether it is worthwhile naming these after their states, as opposed to the actual colour (e.g. info, warning, success) - so that the styling/UX can be independent from the behaviour.

keithbloom: Hmm, good point. I'll take a look what is involved.

PR #1316, opened 2019-11-28
fix glue42 exception

sizov:


src/client/src/rt-components/tear-off/ExternalWindow.tsx
```diff
@@ -43,8 +42,7 @@ const ExternalWindow: FC<ExternalWindowProps> = ({
     const getWindow = async () => {
       externalWindow = await platform.window.open(config, release)
       if (externalWindow) {
-        externalNativeWindow = await externalWindow.getNativeWindow()
```
sizov: this is what I fixed in recent refactor - I think this should have been a listener on the external window, but it was a listener on a global window. So I fixed it to listen to external. Unfortunately, this lead to introducing a new API - `getNativeWindow`. This API was not implemented in glue42 because `any` was used in the typing (see change in `glue.ts`). THus we were not able to see a compilation error which would have showed us that `getNativeWindow` was not used there. Lesson learned - try to get rid of all `any` - they are evil :)

src/client/src/rt-platforms/glue/glue.ts
```diff
@@ -35,9 +35,8 @@ export class Glue42 implements Platform {
       })
   }
 
-  // TODO how can we remove the any?
```
sizov: any is evil :)

src/client/src/rt-components/tear-off/ExternalWindow.tsx
sizov: to sum up - I removed the new `getNativeWindow` api in favor of old approach - listening to global window (whatever reason for that was)

PR #1317, opened 2019-11-28
feat: removes external chart from spot tile (ARTP-875)

keithbloom:
* Removes the button from the spot tile which appears on hover to launch an external chart
* Fixes a typo


src/client/src/apps/MainRoute/widgets/spotTile/epics/combineSpotTileEpics.ts
```diff
@@ -22,9 +21,5 @@ export default ({ platform }: ApplicationDependencies) => {
     epics.push(publishPriceUpdateEpic, publishTradeExecutedEpic, closePositionEpic)
   }
 
-  if (platformHasFeature(platform, 'app')) {
-    epics.push(displayCurrencyChartEpic)
-  }
-
```
rikoe: Seeing this I am now wondering if we are doing the right thing in removing this entirely instead of just disabling the feature.

I am not currently sure if the Finsemble platform uses this feature as well. I guess it's easy enough to restore if we decide to pull it back in.

PR #1326, opened 2019-12-17
Replacing crossbar.io with rabbitMQ

PiRXlv:
Things left to do:
1. fix client side tests
~2. fix server side tests~ (done)
3. add docker file for RabbitMQ (with plugins) (should be done)
4. review how load balancing works 
5. test more scenarios around loosing connection and reconnecting
6. update documentation


src/client/src/rt-system/AutoBahnConnection.ts
```diff
@@ -1,8 +1,12 @@
 import { Connection } from 'autobahn'
```
rikoe: Instead of editing AutobahnConnection interface and implementation, can we create a "new" interface without the name Autobahn in it? Everything else stays the same. E.g. just `Connection` for the interface, and `RabbitStompConnection` for the implementation, or something like that...

rikoe:
Close PR for now, can reopen later.

PR #1332, opened 2019-12-28
feat(client): upgrade glue42 dependencies

rikoe:
- Alpha versions with no security vulnerabilities
- Forced change to GDObject typing


rikoe:
Closing temporarily until Glue42 publish non-alpha versions of their dependencies, which we are waiting for.

PR #1336, opened 2020-01-02
feat(Client): browser header displays launch env

mattydono:
Feat:

Browser header will now display the following alongside its respective environments:

localhost: Reactive Trader Cloud (LOCAL) 
dev: Reactive Trader Cloud (DEV)
uat: Reactive Trader Cloud (UAT)
demo: Reactive Trader Cloud

undefined: Reactive Trader Cloud (UNKNOWN)


src/client/src/index.tsx
```diff
@@ -11,9 +11,26 @@ const SimpleLauncher = lazy(() => import('./apps/SimpleLauncher'))
...
+  demo: '',
+}
+
 async function init() {
   console.info('BUILD_VERSION: ', process.env.REACT_APP_BUILD_VERSION)
 
+  const env = getEnv()
+  document.title = `${document.title} ${envTitles[env]}`
+
+  console.log('HERE', process.env.REACT_APP_BROKER_HOST)
```
rikoe: @mattydono is this meant to be here? If so, a better log message than "HERE" will be needed 😉

mattydono: It was not :) sorry this has been fixed alongside refactored code!

Sharakai:
> Minor feat:
> 
> extract then apply env to browser header

No such thing as a "minor feat". Don't down-sell any additions or changes you make. It's simply a "feature", "fix", etc.

src/client/src/index.tsx
```diff
@@ -11,9 +11,28 @@ const SimpleLauncher = lazy(() => import('./apps/SimpleLauncher'))
 
 const urlParams = new URLSearchParams(window.location.search)
 
+function getEnv() {
```
Sharakai: This is a utility function that can likely be re-used across the codebase. If there's a utility package somewhere, move it to there.

src/client/src/index.tsx
```diff
@@ -11,9 +11,28 @@ const SimpleLauncher = lazy(() => import('./apps/SimpleLauncher'))
...
+  const serviceUrl = process.env.REACT_APP_BROKER_HOST || window.location.host;
+  if (serviceUrl.includes('localhost')) return 'localhost';
+  const envMatch = serviceUrl.match(/web-([a-zA-Z]+)\.adaptivecluster\.com/);
```
rikoe: @mattydono with regex's I tend to prefer a named capture group as opposed to using array index matching.

src/client/src/index.tsx
```diff
@@ -11,9 +11,28 @@ const SimpleLauncher = lazy(() => import('./apps/SimpleLauncher'))
...
+  dev: '(DEV)',
+  uat: '(UAT)',
+  demo: '',
+}
+
 async function init() {
   console.info('BUILD_VERSION: ', process.env.REACT_APP_BUILD_VERSION)
 
+  const env = getEnv()
+  if (env !== undefined) document.title = `${document.title} ${envTitles[env]}`
```
rikoe: What will happen if `env` is undefined, will the title just revert to "Reactive Trader Cloud"?

mattydono: yes, unless you wanted this to be set to something specific?

rikoe: I am actually unsure - should we set it to "Undefined" or "Unknown" to indicate the issue? Otherwise it is the same as for Demo 🤔

src/client/src/index.tsx
mattydono: this has now been updated in latest commit and is significantly easier to read as opposed to the index matching!

mattydono:
> Have you tested this by changing the `REACT_APP_BROKER_HOST` to the different values?
> 
> **dev**: `web-dev.adaptivecluster.com`
> **demo**: `web-demo.adaptivecluster.com`
> **uat**: `web-uat.adaptivecluster.com`
> **local**: Run `npm run start:local` or whatever the local-dev script is

src/client/src/index.tsx
mattydono: Yep tested with the different values!

I've also changed the name of the function to something more appropriate for its move to the rt-util folder: getEnvironment

src/client/src/index.tsx
mattydono: The browser will now display (UNKNOWN) if the environment returns undefined in order to differentiate itself from Demo

src/client/src/index.tsx
```diff
@@ -4,16 +4,27 @@ import { BrowserRouter, Route, Switch } from 'react-router-dom'
 import { GlobalStyle } from 'rt-theme'
 import * as serviceWorker from './serviceWorker'
 import { getSymphonyPlatform } from 'rt-platforms'
+import { getEnvironment } from 'rt-util/getEnvironment'
```
rikoe: 👍

src/client/src/index.tsx
```diff
@@ -4,16 +4,27 @@ import { BrowserRouter, Route, Switch } from 'react-router-dom'
...
+  dev: '(DEV)',
+  uat: '(UAT)',
+  demo: '',
+}
+
 async function init() {
   console.info('BUILD_VERSION: ', process.env.REACT_APP_BUILD_VERSION)
 
+  const env = getEnvironment()
+  document.title = `${document.title} ${env !== undefined ? envTitles[env] : '(UNKNOWN)'}`
```
Sharakai: nit: Just to keep the `(UNKNOWN)` alongside all the other titles, you could do:

```ts
const envTitles = {
  localhost: '(LOCAL)',
  dev: '(DEV)',
  uat: '(UAT)',
  demo: '',
  unknown: '(UNKNOWN)',
}
...
document.title = `${document.title} ${envTitles[env || 'unknown']}`;
```

src/client/src/rt-util/getEnvironment.ts
```diff
@@ -0,0 +1,8 @@
+export function getEnvironment(): string | undefined {
+    const serviceUrl = process.env.REACT_APP_BROKER_HOST || window.location.host;
+
+    if (serviceUrl.includes('localhost')) return 'localhost';
+
+    const envMatch = /web-(?<env>[a-zA-Z]+)\.adaptivecluster\.com/.exec(serviceUrl)
+    return envMatch ? envMatch['groups'].env : undefined
+}
```
Sharakai: Note: Need to configure `prettier` and/or `tslint` to always add a new line at EOF to prevent weird git diffs

src/client/src/rt-util/getEnvironment.ts
```diff
@@ -0,0 +1,8 @@
+export function getEnvironment(): string | undefined {
+    const serviceUrl = process.env.REACT_APP_BROKER_HOST || window.location.host;
+
+    if (serviceUrl.includes('localhost')) return 'localhost';
+
+    const envMatch = /web-(?<env>[a-zA-Z]+)\.adaptivecluster\.com/.exec(serviceUrl)
+    return envMatch ? envMatch['groups'].env : undefined
```
Sharakai: Shouldn't use a static key as a dynamic key.
```suggestion
    return envMatch ? envMatch.groups.env : undefined
```

mattydono: this produces a ts error:

'groups' doesn't exist on RegExpExecArray

Sharakai: https://github.com/microsoft/TypeScript/issues/29465#issuecomment-455489648

src/e2e/src/tests/trade.spec.ts
```diff
@@ -35,7 +44,8 @@ describe('UI Tests for Reactive Trader Cloud Web Application', async () => {
 
   it('Verify page title', async () => {
     const title = await browser.getTitle()
-    expect(title).toBe('Reactive Trader Cloud')
+    const isValid = envTitles.includes(title)
+    expect(isValid).toBeTruthy()
```
Sharakai: You typically know which environment your test is running in, so can be very specific about what to expect rather than saying "One of all of these".

PR #1337, opened 2020-01-02
Fix RFQ Button

zhouyuhe:
Change the RFQ position:  RFQ button is now aligned with the price unavailable boxes


Sharakai:
Looks like you have 2 merge commits in the commit history of this branch, which is a little odd. I'd recommend `rebase`ing instead of merging.

Makes sure this is squashed and the commit message corrected when this is pushed into the `master` branch

src/client/package.json
```diff
@@ -20,7 +20,7 @@
   },
   "scripts": {
     "precommit": "lint-staged",
-    "start": "cross-env REACT_APP_BUILD_VERSION=$(git describe) react-scripts start",
+    "start": "react-scripts start",
```
Sharakai: Q: How come this was changed? Looks unrelated to the issue at hand.

src/client/src/apps/MainRoute/widgets/spotTile/components/notifications/TileBooking.tsx
```diff
@@ -30,6 +30,8 @@ const BookingPill = styled.div<{ disabled: boolean; altStyle: boolean; color: st
     !disabled &&
     `
   cursor: pointer;
+  position: relative;
+  top:-8px;
```
Sharakai: You're using absolute units to position this, but padding is in relative units.
Would reducing the `padding` fix this, rather than having to use a negative placement?

Sharakai: Make sure that this looks okay at multiple zoom levels.
I recommend 75%, 100%, and 150%, and 200% at a minimum.

src/client/package.json
rikoe: We changed it to get `npm start` working on @zhouyuhe 's machine - I added that bit, but it only seems to work on Mac/Unix 😞

zhouyuhe: Because with `REACT_APP_BUILD_VERSION=$(git describe)` it shows the error, and I cannot run the app

src/client/src/apps/MainRoute/widgets/spotTile/components/notifications/TileBooking.tsx
zhouyuhe: Good spot, will change it

rikoe:
Closing as we are revisiting the design.

PR #1341, opened 2020-01-06
fix(client): align RFQ functionality with Zeplin designs (ARTP-899)

zhouyuhe:
- move the RFQ button into the correct div
- RFQ button now follow the zeplin design in non-analytic style (no zeplin in analytic style)
image

### DIFF
--- a/src/client/src/apps/MainRoute/widgets/spotTile/components/PriceControls/PriceControls.tsx
+++ b/src/client/src/apps/MainRoute/widgets/spotTile/components/PriceControls/PriceControls.tsx
@@ -115,14 +115,6 @@ const PriceControls: React.FC<Props> = ({
       isAnalyticsView={isAnalyticsView}
     >
       <PriceMovement priceMovementType={priceMovement} spread={spreadValue} />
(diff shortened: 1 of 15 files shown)


src/client/src/apps/MainRoute/widgets/spotTile/components/PriceControls/PriceControls.tsx
```diff
@@ -47,6 +55,7 @@ const PriceControls: React.FC<Props> = ({
     currencyPair.pipsPosition,
     currencyPair.ratePrecision,
   )
+  const userError = Boolean(inputValidationMessage)
```
mozmorris: I can see you have matched the prop by naming this `userError` which is fine. If we were to refactor this a little including the `<TileBookingSwitch />` component, a nice style is to prefix boolean values with `has`, `is`, or `should` eg `hasUserError` or simply `hasError`

zhouyuhe: good spot, I will change it

PR #1342, opened 2020-01-07
fix(client): remove start script that doesn't work on windows

rikoe:


src/client/package.json
```diff
@@ -20,7 +20,7 @@
   },
   "scripts": {
     "precommit": "lint-staged",
-    "start": "cross-env REACT_APP_BUILD_VERSION=$(git describe) react-scripts start",
```
rikoe: Command expansion only works on Mac and Linux, the default command executer for Windows is `cmd.exe`, which doesn't support this.

PR #1343, opened 2020-01-07
allow rfq state tiles to pop out

ianchan035:


src/client/src/apps/MainRoute/widgets/spotTile/components/TileSwitch.tsx
```diff
@@ -63,7 +63,7 @@ const TileSwitch: React.FC<Props> = ({
     >
       {({ notional, userError }: TileSwitchChildrenProps) => (
         <>
-          <TileControls canPopout={isRfqStateNone && canPopout} onPopoutClick={onPopoutClick} />
+          <TileControls canPopout={isRfqStateNone || isRfqStateCanRequest && canPopout} onPopoutClick={onPopoutClick} />
```
rikoe: What possible RFQ states are there? I think we should make the tile pop-outable regardless of RFQ state... so just "canPopout" (if that is always true for RFQs)

Also I don't think this is doing what you think? Remember that and takes precedence over or, so this means:
`isRfqStateNone || (isRfqStateCanRequest && canPopout)` not `(isRfqStateNone || isRfqStateCanRequest) && canPopout)`, which I assume is what you want?

ianchan035: the states include `isRfqStateReceived` , `isRfqStateExpired`, `isRfqStateCanRequest`, `isRfqStateRequested` and `isRfqStateNone`

rikoe: So the tile should be pop-outable in all the states, not just some of them, especially since there doesn't seem to be a restriction on dragging out a tile in any state.

There is of course a question of whether it is possible to maintain the state when the tile is popped out, since it is just a URL 🤔 .

@mozmorris what do you think? If transitioning the state is a problem, we may disable the drag-out to match the behaviour of the pop-out button (i.e. doesn't work in some states), but the right solution would be maybe be using local storage to ensure the state is transitioned when dragged/popped out?

ianchan035: https://weareadaptive.atlassian.net/secure/RapidBoard.jspa?rapidView=148&projectKey=ARTP&modal=detail&selectedIssue=ARTP-833 i think this could be related?

mozmorris: I don't think transitioning will be a problem here. I'm just wondering while it was disabled in the first place.

I'd be happy to just switch this to `<TileControls canPopout onPopoutClick={onPopoutClick} />`.

ianchan035: I just tested it, when it is in RFQ state and try to pop the tile, the state is lost in the popped out window

mozmorris: So it sounds like @rikoe's original suggestion of `(isRfqStateNone || isRfqStateCanRequest) && canPopout` is what we need?

mozmorris: ...and conflicts need fixing.

PR #1344, opened 2020-01-07
update to layouts 2.0

limerentfox:
## Context
Openfin has released a major update of its [layouts service](https://developers.openfin.co/docs/platform-api). The new version comes bundled and does not need to be installed as a separate package. 

With [Layouts 2.0](https://cdn.openfin.co/docs/javascript/14.78.46.23/Layout.html) the premise of window layout management is entirely different. Instead of snapping and docking windows, Openfin introduces the concept of `views`. A platform can house multiple views. 
image

Think of them as subpanels within an application. For those of you who use VSCode, these views are the same as the tabs in the editor. You can drag, resize, stack, tear out and reorder them.

## MVP
image

In this demo, RT's main app window and it's subpanels (analytics, blotter, spot tiles), have been converted into tab format. You can grab and resize each panel, lift it and rearrange and even tear it out of the main window. 

## Note
The code needs cleaning and some decisions need to be made. 

### Known Issues
Popped out spot tiles are currently running off layouts 1.0 and use the 'snap and dock' functionality.

Any updates to the view don't fully propagate until you resize the moved panels. Openfin seems to not be refreshing.

Window controls drag regions do not work unless the window is resized.

Tearing any tab out of the main window leads to it being opened in an unstyled default window frame, with default Openfin tab styles.

https://drive.google.com/file/d/16EcBqyBtNDKT20mOcwCyP-X6ON2fa9b4/view?usp=sharing

### UX Considerations
What behavior do we want when we drag a child window out? Should it become its own view and be tabbable?

How do we want to handle saving state? Will there be a workspace dropdown?

### DIFF
--- a/src/client/public/config/openfin/platforms.json
+++ b/src/client/public/config/openfin/platforms.json
@@ -1,0 +1,191 @@
+{
+    "runtime": {
+        "arguments": "--v=1 --inspect",
+        "version": "14.78.48.3"
+    },
+    "shortcut": {
+        "company": "Adaptive Consulting",
+        "description": "Desktop Reactive Trader Cloud",
+        "icon": "http://localhost:3000/static/media/icon.ico",
+        "name": "Reactive Trader Cloud (LOCAL)"
+    },
+    "appAssets": [
+        {
+            "src": "http://localhost:3000/plugin/add-in.zip",
+            "alias": "excel-api-addin",
+            "version": "2.0.0",
+            "forceDownload": true
+        }
+    ],
+    "services": [
+        {
+            "name": "layouts",
+            "config": {
+                "features": {
+                    "snap": true,
+                    "dock": true,
+                    "tab": false
+                }
+            },
+            "manifestUrl": "https://cdn.openfin.co/services/openfin/layouts/1.2.0/app.json"
+        },
+        {
+            "name": "fdc3",
+            "manifestUrl": "https://cdn.openfin.co/services/openfin/fdc3/0.2.0/app.json"
+        },
+        {
+            "name": "notifications",
+            "manifestUrl": "https://cdn.openfin.co/services/openfin/notifications/0.11.1/app.json"
+        }
+    ],
+    "platform": {
+        "uuid": "platform_prototype_generated",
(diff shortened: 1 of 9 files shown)


src/client/src/apps/MainRoute/routes/ShellRoute.tsx
```diff
@@ -33,6 +35,13 @@ const ShellRoute: React.FC<Props> = ({ header }) => {
   const blotter = useSelector(blotterSelector)
   const analytics = useSelector(analyticsSelector)
 
+  useEffect(() => {
```
limerentfox: Openfin looks for container with `id="layout-container` on `DOMContentLoaded` to load the platform views. React needs to finish loading and rendering before Openfin can look for this container, otherwise nothing loads. To get around this timing issue I manually call `DOMContentLoaded`.

src/client/public/config/openfin/platforms.json
```diff
@@ -0,0 +1,191 @@
...
+                                    "isClosable": true,
+                                    "reorderEnabled": true,
+                                    "title": "",
+                                    "activeItemIndex": 0,
+                                    "content": [
+                                        {
+                                            "type": "component",
+                                            "componentName": "view",
+                                            "componentState": {
+                                                "url": "http://localhost:3000/analytics",
```
limerentfox: Tab headers need to each be set according to their route. Will need to use something like `react-helmet`.

src/client/src/apps/MainRoute/routes/ShellRoute.tsx
```diff
@@ -85,6 +94,45 @@ const ShellRoute: React.FC<Props> = ({ header }) => {
     </StatusBar>
   );
 
+  //@ts-ignore
+  if (window.fin && window.fin.me.isWindow) {
+    const OfTabTheme = createGlobalStyle<{ theme: Theme }>`
+      .lm_tabs {
```
limerentfox: This can be extracted out for easier reuse. In addition to specific classes that we can override the styles of, Openfin provides css variables as well.

PR #1345, opened 2020-01-08
Release/v1.2 - Launcher instance (ARTP-900)

mattydono:
Fix: uuid is now unique to the environment launched

This should allow us to launch a single openfin instance per env.


src/client/src/apps/utils/openfin-utils.ts
```diff
@@ -31,25 +32,27 @@ async function restoreExistingApp(existingApp: Application): Promise<void> {
...
   name,
   url,
+  uuid,
   windowOptions,
 }: ApplicationConfig): Promise<Application> {
   const appOptions: fin.ApplicationOption = {
     name,
     url,
-    uuid: name,
+    uuid: uuid!,
```
mozmorris: Thinking about this now @mattydono, you probably want to fallback to the original `name` arg. e.g `uuid: uuid || name,`

src/client/package.json
```diff
@@ -20,7 +20,7 @@
   },
   "scripts": {
     "precommit": "lint-staged",
-    "start": "cross-env REACT_APP_BUILD_VERSION=$(git describe) react-scripts start",
```
mozmorris: @rikoe I take it we're happy for these other changes to find their way into this release branch?

src/client/src/apps/utils/openfin-utils.ts
mattydono: ah yes that makes a lot more sense!

src/client/package.json
rikoe: yeah that's fine

PR #1352, opened 2020-01-09
Initial work to abstract out FDC3 broadcast

mozmorris:
More to follow, but this should be a non-breaking change

* Support other intents providers
* Support intents providers on multiple platforms
* Remove references to fdc3 from platform
* Replace with `intents` provider object
* Noop intents provider for platforms without support


src/client/src/apps/MainRoute/widgets/blotter/components/Blotter.tsx
```diff
@@ -73,25 +72,14 @@ const Blotter: React.FC<BlotterProps> = props => {
...
-    if (!platform.fdc3.broadcast) {
-      console.error(`Broadcasting is not available on the platform ${platform.name}`)
+    const { currencyPairSelected } = platform.intents
+
+    if (!currencyPair || typeof currencyPairSelected !== 'function') {
       return
     }
-    platform.fdc3.broadcast(getCurrencyPairContext(currencyPair))
+
+    currencyPairSelected(currencyPair)
```
rikoe: I really like this - exposing intents at a platform level seems like a great idea.

src/client/src/rt-intents/fdc3/contexts/getCurrencyPair.ts
```diff
@@ -0,0 +1,11 @@
+import { Context } from 'openfin-fdc3'
+
+const getCurrencyPairContext = (currencyPair: string): Context => {
+  const ticker = `${currencyPair.slice(0, 3)}/${currencyPair.slice(3, 6)}`
+  return {
+    type: 'fdc3.instrument',
+    id: { ticker },
+  }
+}
+
```
rikoe: Nice

src/client/src/rt-platforms/browser/browser.ts
```diff
@@ -39,9 +40,7 @@ export default class Browser implements Platform {
     open: (config: WindowConfig, onClose?: () => void) => openBrowserWindow(config, onClose),
   }
 
-  fdc3 = {
-    broadcast: () => {},
-  }
+  intents = new Noop()
```
rikoe: Is there a way we can do this so that platforms don't have to explicitly set intents to `Noop`? Just so they can opt-in as opposed to having to opt out...

src/client/src/apps/MainRoute/widgets/blotter/components/Blotter.tsx
rikoe: On second thought - what we are doing here is a "context broadcast", not raising an intent, should we distinguish in the API, or is it not necessary?

src/client/src/rt-platforms/browser/browser.ts
mozmorris: I guess we could have another wrapper/abstraction around the methods which check they exist. I mainly did this for typing purposes. I'll leave as is here, but let's iterate on the other branch.

src/client/src/apps/MainRoute/widgets/blotter/components/Blotter.tsx
mozmorris: Will follow up on the other PR.

PR #1354, opened 2020-01-09
fix(client): center currency pair windows (ARTP-915)

fsarda:


mozmorris:
@fsarda looks good, just need to rebase with latest from `master`

PR #1357, opened 2020-01-13
Fix/lost rfqstate in pop out

ianchan035:


src/client/src/apps/MainRoute/widgets/workspace/selectors.ts
```diff
@@ -31,17 +32,25 @@ const makeExternalWindowProps: (key: string, tileLayout?: WindowPosition) => Ext
 
 const getSpotTiles = (state: GlobalState) => state.currencyPairs
 const getSpotTilesLayout = (state: GlobalState) => state.layoutService.spotTiles
+const getSpotTilesData = (state: GlobalState) => state.spotTilesData
+
+const spotTilesDataTypeChecker = (spotTilesData: SpotTileState, key: string) => typeof spotTilesData !== 'undefined' && typeof spotTilesData[key] !== 'undefined'
```
mozmorris: Pretty sure `spotTilesData` is always an object, so you can reduce this to only `typeof spotTilesData[key] !== 'undefined'`.

If that is the case, then do you need this function? See comment below.

src/client/src/apps/MainRoute/routes/SpotRoute.tsx
```diff
@@ -16,16 +29,57 @@ const SpotTileStyle = styled.div`
   margin: 0 auto;
 `
 
-const getTileViewFromQueryStr: (queryStr: string) => TileViews = queryStr => {
-  const parsedQueryString = queryString.parse(queryStr)
+const getFloatFromQuery = (query: string[] | string | null | undefined): number | undefined => {
```
mozmorris: Will Slack you about these :-)

mozmorris: fyi for other reviewers: we're going to ditch these methods.

rikoe:
Can we add some unit test / e2e test coverage for this?

src/client/src/apps/MainRoute/routes/SpotRoute.tsx
```diff
@@ -6,6 +6,19 @@ import { TileViews } from '../widgets/workspace/workspaceHeader'
 import { styled } from 'rt-theme'
 import { InteropTopics, platformHasFeature, usePlatform } from 'rt-platforms'
 import { Subscription } from 'rxjs'
+import { RfqState } from '../widgets/spotTile/components/types'
+
+export interface rfqQueryObject {
```
bhavesh-desai-scratch: Why are some props undefined and some null? Should this live in its own file

src/client/src/apps/MainRoute/routes/SpotRoute.tsx
```diff
@@ -16,16 +29,33 @@ const SpotTileStyle = styled.div`
...
   const tileView = parsedQueryString['tileView'] as TileViews
+
   return !tileView
     ? TileViews.Normal
     : Object.values(TileViews).includes(tileView)
     ? tileView
     : TileViews.Normal
 }
 
+const getRfqDatafromQuery: (query: string) => rfqQueryObject = query => {
```
bhavesh-desai-scratch: should live in its own file

PR #1358, opened 2020-01-13
Fix - reduce memory usage for analytics mode

fsarda:
- Limit graph points to 250
- Use slice indicating start and end index to improve array copy


src/client/src/apps/MainRoute/widgets/spotTile/spotTileDataReducer.ts
```diff
@@ -41,10 +42,18 @@ const spotTileReducer = (
     case TILE_ACTION_TYPES.SPOT_TILE_SUBSCRIBE:
       return state
     case TILE_ACTION_TYPES.SPOT_PRICES_UPDATE:
+      const startIndexSlice =
```
mozmorris: Fran, I think you could use `Math.max` here to keep this nice and tidy:

eg

```javascript
const startIndexSlice = Math.max(1, state.historicPrices.length - HISTORIC_PRICES_MAX_POINTS)
```

PR #1359, opened 2020-01-14
ARTP-930

audreyf1:


src/server/Adaptive.ReactiveTrader.Common/DateUtils.cs
```diff
@@ -20,17 +20,32 @@ public static DateTime FromSerializationFormat(string dateTimeString)
             return date.ToUniversalTime();
         }
 
-        public static DateTime ToWeekday(this DateTime date)
+        public static DateTime UnixTimeStampToDateTime(double unixTimeStamp)
```
rikoe: This looks like a good change to me, but as discussed offline, let's switch to `DateTimeOffset` instead of `DateTime`.

PR #1362, opened 2020-01-15
Dynamic OpenFin installer configs

mozmorris:


src/e2e/Dockerfile
```diff
@@ -15,5 +15,6 @@ ENV CHROME_BIN /usr/bin/chromium-browser
 WORKDIR /client-tests
 
 COPY . .
+RUN npm set unsafe-perm true
```
rikoe: Why has this been added?

src/client/lib/downloadInstaller.js
```diff
@@ -42,24 +45,23 @@ Make sure the files are available on their respective locations when distributin
...
-  { manifestName: 'launcher-demo', branch: 'master', os: 'win' }
+  { type: 'app', env: 'dev', os: 'win' },
+  { type: 'app', env: 'dev', os: 'osx' },
+  { type: 'app', env: 'uat', os: 'win' },
+  { type: 'app', env: 'uat', os: 'osx' },
+  { type: 'app', env: 'demo', os: 'win' },
+  { type: 'app', env: 'demo', os: 'osx' },
+  { type: 'launcher', env: 'dev', os: 'win' },
+  { type: 'launcher', env: 'uat', os: 'win' },
+  { type: 'launcher', env: 'demo', os: 'win' }
```
rikoe: Once we have made this change, my plan is that we don't actually build installers or check them into source control.

We can use direct links like this: `https://install.openfin.co/download/?os=win&config=https%3A%2F%2Fweb-demo.adaptivecluster.com%2Fopenfin%2Fapp.json&fileName=ReactiveTraderCloud-Demo&unzipped=true`

Or even `fins://web-demo.adaptivecluster.com/openfin/app.json`, which works if the OpenFin runtime is already installed.

src/node-server/openfin-config/src/config/launcher.ts
```diff
@@ -0,0 +1,94 @@
+/* eslint-disable @typescript-eslint/ban-ts-ignore */
```
rikoe: There is a lot of commonality between app.ts and launcher.ts - can we be smarter about sharing between the two?

src/node-server/openfin-config/src/config/app.ts
```diff
@@ -0,0 +1,73 @@
+/* eslint-disable @typescript-eslint/ban-ts-ignore */
```
rikoe: Now that we have the OpenFin config here, how do we ensure compatibility between this location and the OpenFin config that exists in the client project and is used for `npm run openfin:local` etc.?

It is worth thinking about how to keep the two in sync effectively, and the developer workflow. E.g. should the base OpenFin config file come from a shared location? Maybe we can have template JSON files that we read in and then use variable substitution to put in the environment-specific bits, instead of outputting the whole object via TypeScript?

Not sure, just thinking out loud.

src/services/nginx/default.conf
```diff
@@ -6,6 +6,10 @@ upstream broker {
     server broker:8000;
 }
 
+upstream openfinconfig {
+    server openfinconfig:8080;
+}
+
```
rikoe: Can we please make the paths `/openfin/app.sjon` and `/openfin/launcher.json`. I think having `config` in there is redundant.

src/e2e/Dockerfile
mozmorris: There is a postinstall script as part `e2e` that downloads and installs the selenium server jars/binaries globally. This step was throwing an error. My understanding is the script is attempting to set the wrong uid. This tells npm we're running with root privileges so set the uid accordingly. Voila, the problem went away.

src/client/lib/downloadInstaller.js
mozmorris: Nice. I did notice we're committing the installers to git and was going to say, this change should make that redundant. I didn't want to delete them just yet those until this was merged.

src/node-server/openfin-config/src/config/launcher.ts
mozmorris: Nice, I'll create a base these can extend.

src/node-server/openfin-config/src/config/app.ts
mozmorris: I like the out loud thinking :-) Let me some not out lout thinking and I'll come back to you.

PR #1364, opened 2020-01-15
feat(client): browser header displays env (ARTP-891)

mattydono:
Re-submission of lost code due to merging error.

- browser header now displays correct deployment env alongside title


mozmorris:
I think you need to update the e2e tests too.

mattydono:
> I think you need to update the e2e tests too.

had this up once I saw the fail :)

PR #1371, opened 2020-01-16
fix(client)non-stop-flashing-blotter (ARTP-901)

ianchan035:


src/client/src/rt-platforms/openFin/adapter/openFin.ts
```diff
@@ -43,6 +43,8 @@ export default class OpenFin implements Platform {
   openFinNotifications = require('openfin-notifications')
 
   constructor() {
+    window.addEventListener('unload', () => this.openFinNotifications.removeEventListener('notification-action'))
```
rwleo: Unfortunately, this doesn't seem to be working. Throw a few `console.log()`s around - in the `notification-action` event listener , in the `highlightTradeEpic`, in the `removeHighlightTradeEpic`, etc. - and you'll see that everything still fires X number of times, where X is the number of loads related to the RTC window.

rwleo: That being said, the cosmetic bug - flashing infinitely as opposed to 3 times - has been tended to. Whether or not we should break off the aforementioned "event listener stuff" into its own ticket and proceed with merging in this PR is a decision I'll leave to @rikoe .

rikoe: Let's fix the issue you describe above I think @Ian887. Thanks @rwleo

mozmorris: 1. `this.openFinNotifications.removeEventListener('notification-action')` almost certainly needs a second arg, a ref to the callback.
2. `this.openFinNotifications.addEventListener` is passed an anonymous function, it needs a ref to the callback

Both these points are related. I haven't checked the `removeEventListener` api for `openFinNotifications`, but I'm guessing it needs that second arg. We should either fix that first or confirm it doesn't need the callback ref.

mozmorris: https://cdn.openfin.co/docs/services/notifications/stable/api/modules/notifications.html#removeeventlistener

```
Removes a listener previously added with addEventListener.

Has no effect if eventType isn't a valid event, or listener isn't a callback registered against eventType.
```

rwleo: Exactly right w/r/t the second arg. That said, while reviewing this PR, I edited the call to `this.openFinNotifications.addEventListener` to use a named function and passed that same function to `this.openFinNotifications.removeEventListener` - this seemed to affect nothing, unfortunately.

PR #1373, opened 2020-01-16
feat(config): osx installer for launcher [ARTP-330]

ianchan035:


mozmorris:
This all works fine @ianchan035 - but the installers haven't been committed to Git. There's a note in the README stating that these are linked to from external sources so we should probably add them in.

PR #1375, opened 2020-01-17
fix(client): ui fix on prices for rfq state (ARTP-923)

fsarda:


fsarda:
> @fsarda can you please add screenshots/gifs of how it looks now to aid the review please.

@rikoe And this how currently looks

spottile_beforePR

And this is how it looks with PR changes

spottile

src/e2e/src/tests/trade.spec.ts
```diff
@@ -109,7 +109,8 @@ describe('UI Tests for Reactive Trader Cloud Web Application', async () => {
     await mainPage.tile.initiateRFQ()
   })
 
-  it('should validate unavailable streaming', async () => {
+  // This test is currently disabled since it does not seem to be a case where it happens anymore
+  xit('should validate unavailable streaming', async () => {
```
rikoe: Shall we just remove it then @fsarda ?

fsarda: great, on it

PR #1383, opened 2020-01-21
feat(client): hide theme toggle on launcher (ARTP-929)

fsarda:


fsarda:
This is not going to be included at the moment.

PR #1395, opened 2020-01-23
ARTP-890

audreyf1:


mozmorris:
Haha, I like the thinking here 😄 

I wonder if it might be better to change the circleci config to deploy the broker first, something like

```

PR #1399, opened 2020-01-27
Fix/analytics view layout (ARTP-934 & ARTP-958)

zhouyuhe:
- Tile analytics view layout  not broken at large widths
- Add padding to priveMovement in analytics tile
image
image


rikoe:
This looks great!

PR #1400, opened 2020-01-27
Restore OpenFin workspace on boot

mozmorris:
* Ensures workspace handlers (eg `setRestoreHandler`)  are setup properly on boot
* Persists workspace object to local storage on relevant application events eg resize, move, minimise, etc

### DIFF
--- a/src/client/src/rt-platforms/openFin/adapter/epics.ts
+++ b/src/client/src/rt-platforms/openFin/adapter/epics.ts
@@ -1,16 +1,102 @@
 import { Action } from 'redux'
-import { map, switchMapTo } from 'rxjs/operators'
-import { ActionsObservable, ofType, StateObservable } from 'redux-observable'
-import { LayoutActions } from 'rt-actions/layoutActions'
-import { setupWorkspaces } from './openFin'
+import { fromEventPattern, merge } from 'rxjs'
+import { map, switchMapTo, mergeMap, delay } from 'rxjs/operators'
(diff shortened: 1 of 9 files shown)


src/client/src/rt-platforms/openFin/adapter/epics.ts
```diff
@@ -1,17 +1,102 @@
...
+  SETUP_ACTION_TYPES,
+  WORKSPACE_ACTION_TYPES,
+  LAYOUT_ACTION_TYPES,
+  LayoutAction,
+} from 'rt-actions'
+import { setupWorkspaces, restoreWorkspace, saveWorkspace } from './workspaces'
+import { isMainWindow, addApplicationEventHandler } from './window'
+import { WorkspaceActions, WorkspaceAction, LayoutActions } from 'rt-actions'
+
+const DELAY_SAVING_WORKSPACE = 2_000
```
rikoe: Why is there a delay?

mozmorris: If you generate the workspaces object immediately after creating/closing/moving/minimising/etc openfin windows,  the object frequently contains empty windows eg url: `about:blank`.

This delay could be reduced, I set it this long to be safe. It's not ideal, but I couldn't reliably get it to work without adding this.

mozmorris: Additionally where I use that delay, I've left a comment.

PR #1401, opened 2020-01-27
fix(client): cancel RFQ button in non-graph view fixed (ARTP-961)

ramirozap:
Before: 
image

After: 
viewport width smaller than 320px: 
image
viewport width between 320px and 400px:
image
viewport width 400px and up:
image

### DIFF
--- a/src/client/src/apps/MainRoute/widgets/spotTile/components/notifications/TileBooking.tsx
+++ b/src/client/src/apps/MainRoute/widgets/spotTile/components/notifications/TileBooking.tsx
@@ -24,6 +24,8 @@ const BookingPill = styled.div<{
 }>`
   display: flex;
   justify-content: center;
+  padding-left: ${({ isAnalyticsView }) => (isAnalyticsView ? '1px' : '6px')};
+  padding-right: ${({ isAnalyticsView }) => (isAnalyticsView ? '1px' : '6px')};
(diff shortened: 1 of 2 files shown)


src/client/src/apps/MainRoute/widgets/spotTile/components/notifications/TileBooking.tsx
```diff
@@ -22,6 +22,8 @@ const BookingPill = styled.div<{
   color: string
   isAnalyticsView: boolean
 }>`
+  padding-left: ${'1px'};
```
mozmorris: Does we need to use template placeholders`${someValue}` here?

rikoe:
Can we have a screenshot of the change on the PR please?

src/client/src/apps/MainRoute/widgets/spotTile/components/notifications/TileBooking.tsx
ramirozap: Yes it is safe, I went with `padding-left: '1px'` as in a regular style object as first approach and that didn't work, thats why I used the templating afterwards.

ramirozap:
> Can we have a screenshot of the change on the PR please?

Description Updated with screenshots

mozmorris:
@ramirozap you'll need to rebase and fix the conflict in `TileBooking.tsx`. An `npm install` is needed after and then update the snapshots.

PR #1409, opened 2020-01-29
Feat/styleguide buttons dropdowns

fsarda:


src/client/src/rt-styleguide/components/Dropdown.tsx
```diff
@@ -0,0 +1,23 @@
...
+  title?: string
+}
+
+export const Dropdown: React.FC<DropdownStyleProps> = props => {
+  const { title, ...rest } = props
+
+  return (
+    <Button {...rest} intent="mute">
+      {`${title} `}
+      <FontAwesomeIcon icon={faAngleDown} />
```
mozmorris: <table>
<tr>
<td>Implementation</td><td>Design</td>
</tr>
<tr>
<td>
Screenshot 2020-01-30 at 09 44 59
</td>
<td>
Screenshot 2020-01-30 at 09 44 26
</td>
</tr>
</table>

A few thoughts here:

* The angle down icon needs some spacing
* In the design the buttons are of equal width. This to me is an implementation detail so we should adhere to the design but implement the css as if we were using the buttons/dropdowns in a project eg. it's the application's concern to make them equal width not the component we're importing
* The hover/disabled states don't match the design

src/client/src/rt-styleguide/components/Dropdown.tsx
```diff
@@ -0,0 +1,23 @@
...
+  disabled?: boolean
+  title?: string
+}
+
+export const Dropdown: React.FC<DropdownStyleProps> = props => {
+  const { title, ...rest } = props
+
+  return (
+    <Button {...rest} intent="mute">
+      {`${title} `}
```
mozmorris: See comment re: icon spacing - we should use css to add the correct spacing.

PR #1418, opened 2020-01-31
Fix/check for undefined time zone

ianchan035:
Some environment may not support IANA-specified zones, this PR is to check if your runtime environment supports IANA zones and would return utc time by default


src/client/src/apps/MainRoute/widgets/blotter/components/blotterUtils.ts
```diff
@@ -19,7 +19,7 @@ import {
 } from '../blotterFields'
 
 function UtcFormatDate(date: Date) {
-  const localZoneName = DateTime.local().zoneName
+  const localZoneName = Info.features().zones ? DateTime.local().zoneName : 'utc'
```
rikoe: Can we log to the console what time zone is detected and used? May be useful for debugging in the future.

PR #1422, opened 2020-02-04
Store workspace view state in URL

mozmorris:
* Store the currency and tile type in the url eg `/ALL/Analytics`, `/GBP/Normal`
* Update nav items to use hyperlinks
* Redirect index path to `/ALL/Analytics`

Other stuff:
* Bump react-router-dom
* re-enable e2e tests and point them at `/ALL/Normal` (prevents charts being displayed, lots less memory used)


mozmorris:
I think I should also point out that when the workspace is restored, it will now remember the previous state of the tiles workspace.

PR #1424, opened 2020-02-05
Fix/style guide navigation

fsarda:
- update hash and active section on scroll
- fix a key warning on dropdown menu
scrollnav

### DIFF
--- a/src/client/src/apps/StyleguideRoute/components/OnePageNavBar.tsx
+++ b/src/client/src/apps/StyleguideRoute/components/OnePageNavBar.tsx
@@ -1,4 +1,4 @@
-import React, { useRef, useEffect, useState } from 'react'
+import React, { useRef, useEffect, useState, useCallback } from 'react'
 import { HashLink as Link } from 'react-router-hash-link'
 import { css } from 'styled-components'
 import { styled } from 'rt-theme'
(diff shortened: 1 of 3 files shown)


src/client/src/apps/StyleguideRoute/components/OnePageNavBar.tsx
```diff
@@ -20,49 +30,96 @@ const OnePageNavBar: React.FC<OnePageNavBar> = props => {
   const [scrollTop, setScrollTop] = useState(window.scrollY)
   const [positionNavBar, setPositionNavBar] = useState(MAX_SCROLL_HEIGHT)
   const [currentSection, setCurrentSection] = useState('')
-  const handleScroll = () => setScrollTop(window.scrollY)
+  const [isSticky, setIsSticky] = useState(false)
+
+  const handleScroll = useCallback(() => {
+    setScrollTop(window.scrollY)
+
+    const getCurrentSection = (): NavSection => {
```
mozmorris: @fsarda there's now quite a lot going on in this event handler e.g creating a function & executing it. Inside that newly created function there's calculations too. The result for me is a fairly janky scroll performance. I also get a warning in the console:

Screen Shot 2020-02-05 at 11 58 25

I'd suggest moving `getCurrentSection` outside of the scroll handler and wrapping it in a useCallback.

rikoe: Looking great @fsarda !

src/client/src/apps/StyleguideRoute/components/OnePageNavBar.tsx
```diff
@@ -20,49 +30,95 @@ const OnePageNavBar: React.FC<OnePageNavBar> = props => {
...
+    })
+
+    return currentSection
+  }, [sections])
+
+  const handleScroll = useCallback(() => {
+    setScrollTop(window.scrollY)
+    const currentSectionScrolled: NavSection = getCurrentSection()
+
+    if (typeof currentSectionScrolled !== 'undefined') {
```
mozmorris: Can we add another conditional check in here. For example:

```javascript
&& !location.hash.includes(currentSectionScrolled.path)
```

or 

```javascript
&& currentSectionScrolled.title !== currentSection
```

I think it will significantly improve performance as calling `history.pushState` is relatively slow; if we can only call it when needed, that will help :-)

mozmorris:
Nice @fsarda, this feels a lot smoother - great work.

PR #1434, opened 2020-02-10
Remove setting default notional in selector [ARTP-991]

mozmorris:
See https://weareadaptive.atlassian.net/browse/ARTP-991 for further details.

**TLDR**: There's little point adding the default notional (static) to the spot tile data (dynamic).

This refactor plus the changes made in #1431 result in a noticeable UI performance improvement.

**Before:**
prePerf

**After:**
postPerf


rikoe:
Should this go into master or 1.3? Want to release 1.3 this week so we can stop making this distinction...

mozmorris:
Let's get it into 1.3, will update.

PR #1438, opened 2020-02-11
fix(client): keep blotter row transition consistent

ianchan035:
This PR is to allow the row to have the same transition of background colour when the theme changes


rikoe:
@ianchan035 should we target this at release/1.3 branch instead? is it a fix to the blotter row highlighting?

ianchan035:
> @ianchan035 should we target this at release/1.3 branch instead? is it a fix to the blotter row highlighting?

This is just I noticed that when the theme is changing the row bg colour does not have the same transition style , this is nothing to do with the highlight

PR #1442, opened 2020-02-12
Feat/launch bar icons style

fsarda:
Change to launchbar app:

image
Look and feel changes

image

image

Adding tooltip with app name


rikoe:
Could you please add some screenshots/gifs @fsarda

zhouyuhe:
image
I think the tooltip should be inline with the bottom of the launch bar

mozmorris:
https://github.com/AdaptiveConsulting/ReactiveTraderCloud/blob/master/src/node-server/openfin-config/src/config/launcher.ts will also need updating 👍

src/client/public/config/openfin/launcher-dev.json
```diff
@@ -39,7 +40,7 @@
     }
   },
   "runtime": {
-    "arguments": "--remote-debugging-port=9222",
+    "arguments": "--remote-debugging-port=9222 --disable-legacy-window",
```
rikoe: Why did we add this switch?

src/client/public/config/openfin/launcher-demo.json
```diff
@@ -3,11 +3,12 @@
     "name": "Reactive Ecosystem Launcher",
     "url": "https://web-demo.adaptivecluster.com/launcher",
     "uuid": "reactive-launcher-demo",
-    "defaultWidth": 650,
-    "defaultHeight": 52,
+    "defaultWidth": 400,
+    "defaultHeight": 120,
```
rikoe: I wouldn't actually change these JSON files in this way. Actual installed applications without these changes point to this file.

src/client/public/config/openfin/launcher-dev.json
fsarda: In order to make window transparent (referring to this docs https://developers.openfin.co/docs/faq#can-i-make-a-transparent-window-in-openfin) 

image

src/client/public/config/openfin/launcher-demo.json
rikoe: The right place to change the config is here: https://github.com/AdaptiveConsulting/ReactiveTraderCloud/blob/master/src/node-server/openfin-config/src/config/launcher.ts

As this config is now served as part of the environment.

Unfortunately the local config does not follow this pattern yet, so please change only launcher-local.json, and the config in the location above.

rikoe: The above comments apply to all the configs for dev, uat, demo etc.

PR #1449, opened 2020-02-18
Fix opening a spot tile in glue

drodriguez-Tick42:
External spot tiles where showing as blank because of the way URLs were created, which appears to have been broken when the app started loading on a non-root path (it seems that / has started redirecting to /ALL/Analytics)

### DIFF
--- a/src/client/src/rt-platforms/glue/window.ts
+++ b/src/client/src/rt-platforms/glue/window.ts
@@ -79,7 +79,7 @@ export const openGlueWindow = async (config: BrowserWindowProps, onClose?: () =>
     relativeTo,
     relativeDirection,
   } = calculatePosition(myWindow, width, height, url)
-  const fullUrl = `${location.href.slice(0, -1)}${url}`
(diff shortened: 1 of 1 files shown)


rikoe:
@drodriguez-Tick42 I recommend that you run using the latest release (see https://github.com/AdaptiveConsulting/ReactiveTraderCloud/releases/tag/v1.3.0), tag `v1.3.0`, as deployed at https://web-demo.adaptivecluster.com.

The code in `master` points to our next release and is under heavy development. The path feature you mention will likely change again before we deploy the release.

drodriguez-Tick42:
> The path feature you mention will likely change again before we deploy the release.

Thanks, we confirm that the spot tiles in v1.3.0 seem to work correctly in Glue, but I believe the fix in this PR makes that code more future-proof and independent of any changes to the base URL (unless the spot URL construction changes, but then we'll need to modify it in another place, so I think that's not related to this PR), let us know if there might be a better solution

PR #1451, opened 2020-02-19
Feat/search update

ianchan035:
Capture
new search behaviour

cancel button if there is text in input


fsarda:
Nice work @ianchan035 ! Animation looks smooth :D 

A couple of things i've observed:
-  when the results container is opened, the border-radius for the launcher is lost
- I can't close the results container once it has been opened, not even pressing 'esc'

fsarda:
Nice work @ianchan035 ! Animation looks smooth :D 

A couple of things i've observed:
-  when the results container is opened, the border-radius for the launcher is lost
- I can't close the results container once it has been opened, not even pressing 'esc'

PR #1456, opened 2020-02-21
fix(openfin): make sure undock icon is available when reloaded (ARTP-1016)

ianchan035:
This PR is to fix the behaviour where after the popped out window tile is reloaded and missing the undock button while it is docked.

### DIFF
--- a/src/client/src/rt-platforms/openFin/adapter/window.ts
+++ b/src/client/src/rt-platforms/openFin/adapter/window.ts
@@ -2,6 +2,7 @@
 import { WindowConfig } from '../../types'
 import { get as _get, last as _last } from 'lodash'
 import { PlatformWindow } from '../../platformWindow'
+import { snapAndDock } from 'openfin-layouts'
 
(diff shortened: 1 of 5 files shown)


src/client/src/rt-platforms/openFin/adapter/window.ts
```diff
@@ -20,7 +21,7 @@ export const openfinWindowStates: { readonly [key: string]: WindowState } = {
   Maximized: 'maximized',
 }
 
-const generateRandomName = function() {
+const generateRandomName = function () {
```
rikoe: Minor: That space looks a bit out of place 😉

src/client/src/rt-platforms/openFin/adapter/window.ts
```diff
@@ -167,3 +168,28 @@ export const addApplicationEventHandler = (event: string) => (handler: Function)
 
   app.addListener(event, handler as () => any)
 }
+
```
rikoe: Minor: Can probably do with just one new-line here...

PR #1459, opened 2020-02-24
Feat/styleguide form buttons tabs

fsarda:
- Fix buttons display
- Fix inputs display
- Tabs are no longer on styleguide as they are not used
image

### DIFF
--- a/src/client/src/apps/StyleguideRoute/components/FormGrid.tsx
+++ b/src/client/src/apps/StyleguideRoute/components/FormGrid.tsx
@@ -67,7 +67,7 @@ const FormGrid = () => {
       <ActiveNumberInput type="number" placeholder="Prompt" value="10" label="" />
       <Input type="number" placeholder="Prompt" value="10" disabled label="" />
 
-      <div>text</div>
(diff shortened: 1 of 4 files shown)


src/client/src/apps/StyleguideRoute/components/FormGrid.tsx
```diff
@@ -55,36 +56,22 @@ const FormGrid = () => {
...
-      <Input type="text" placeholder="Prompt" disabled label="label" />
-      <Input type="number" placeholder="Prompt" disabled label="label" />
-
-      <div>Populated</div>
-      <Input type="text" placeholder="Prompt" value="Text" label="label" />
-      <Input type="number" placeholder="Prompt" value="1000" label="label" />
+      <div></div>
+      <div></div>
+      <div></div>
+      <div></div>
```
rikoe: Why the empty divs?

fsarda: needed an empty line but fixed with a margin

PR #1461, opened 2020-02-24
restructure server projects

audreyf1:


src/docker-compose.yml
```diff
@@ -42,6 +42,8 @@ services:
       args:
         - CI=true
         - REACT_APP_BUILD_VERSION=${BUILD_VERSION}
+    depends_on:
+      - servers
```
audreyf1: The client takes a huge time to build and freezes on "Creating an optimized production build..." and sometimes stays there forever in local, so I'd rather have it building last...

PR #1478, opened 2020-03-02
Artp 1014 doc

audreyf1:


docs/client.md
```diff
@@ -4,48 +4,51 @@ The trading client GUI is a single page app (SPA) built using Typescript, React,
...
 
 We deploy Reactive Trader to two enviroments:
 
-* [Demo](https://web-demo.adaptivecluster.com/) 
-The most stable version of the application it always reflects the `master` branch
-* [Dev](https://web-dev.adaptivecluster.com/) The most up to date version of the application it always reflects the `develop` branch
-
+- [Demo](https://web-demo.adaptivecluster.com/)
+  The most stable version of the application it always reflects the `master` branch
+- [Dev](https://web-dev.adaptivecluster.com/) The most up to date version of the application it always reflects the `develop` branch
```
rikoe: This info is out of date. We now have three environments - Demo, Dev and UAT, and there is no longer a develop branch. Instead deployments are managed through continuous deployment (for Dev), and the `env-uat` and `env-demo` tags (although you probably don't need to get into that here - you can just reference tag-based deployment via CircleCI.

docs/client.md
```diff
@@ -4,48 +4,51 @@ The trading client GUI is a single page app (SPA) built using Typescript, React,
...
 
 ## Starting the GUI
 
 Clone the repo and install the necessary node modules:
 
 ```sh
 npm install  # Install Node modules listed in ./package.json
-npm start    # Compile and launches the webpack dev server. By default, the client connects to a demo environment.
+npm start    # Compile and launch the webpack dev server. By default, the client connects to a demo environment.
```
rikoe: I believe by default the client now connects to the dev environment (no longer demo) - please confirm!

docs/server.md
```diff
@@ -1,19 +1,26 @@
-Reactive Trader Cloud - Server
-===============
-
-* [Getting Started on Windows](./setup/windows-setup.md)
-* [Getting Started on Linux](./setup/linux-setup.md)
-* [Getting Started on Mac OS](./setup/macos-setup.md)
+# Reactive Trader Cloud - Server
```
rikoe: Do you think it would make sense to collapse the various getting started documents into one, and instead point people towards the setup instructions for the various tools that are environment-specific?

Because once you have installed crossbar, eventstore, dotnetcore etc. for your environment, the commands for running things up should be exactly the same regardless of environment.

This way we could de-duplicate the instructions across the environment-specific documents, and thereby prevent changes not being made everywhere at the same time in the future.

docs/setup/windows-setup.md
```diff
@@ -1,30 +1,37 @@
...
+
 For the web client follow these [instructions](../client.md)
 
 ## Running with Visual Studio 2019 on Windows
 
 Install Visual Studio 2019 Pro or Community. Upon running the installation, be sure to install the following optional components:
-  - .NET Framework 4.6.2
-  - .NET Core 2.1 development tools
+
+- .NET Framework 4.6.2
```
rikoe: Do we really need .NET Framework? It is not needed for the other environments...

docs/client.md
audreyf1: yes, just checked, I will update that,,,

PR #1482, opened 2020-03-04
Feat/new header

ianchan035:
This PR is to update the header in order to align with the new design

Capture

### DIFF
--- a/src/client/src/rt-components/route-wrapper/RouteWrapper.tsx
+++ b/src/client/src/rt-components/route-wrapper/RouteWrapper.tsx
@@ -1,6 +1,8 @@
 import React, { ReactNode } from 'react'
+import { useParams } from 'react-router'
 import { styled } from 'rt-theme'
 import { Platform, usePlatform } from 'rt-platforms'
+import { currencyFormatter } from 'rt-util'
 
 const RouteStyle = styled('div')<{ platform: Platform }>`
   width: 100%;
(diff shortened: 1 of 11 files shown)


src/client/src/rt-components/route-wrapper/RouteWrapper.tsx
```diff
@@ -22,14 +24,31 @@ type WindowRole = 'main' | 'sub'
...
+  const { symbol } = useParams()
+
   const { PlatformHeader, PlatformControls, PlatformRoute, window } = platform
+
+  let formattedCurrency
+
+  if (!title && symbol) {
+    formattedCurrency = currencyFormatter(symbol)
+  }
+
```
fsarda: Can't we just pass symbol formatted as title? seems to me that this component shouldn't know about params. What do you think?

src/client/src/rt-components/route-wrapper/RouteWrapper.tsx
```diff
@@ -22,14 +24,31 @@ type WindowRole = 'main' | 'sub'
...
   const Header = windowType === 'main' ? PlatformControls : null
   const subheader =
-    windowType === 'sub' ? <PlatformHeader popIn={window.close} minimize={window.minimize} /> : null
+    windowType === 'sub' ? (
+      <PlatformHeader
+        popIn={window.close}
+        minimize={window.minimize}
+        title={title || formattedCurrency}
```
fsarda: same here

PR #1505, opened 2020-03-24
Artp 861 rabbitmq

audreyf1:
Replacing autobahn lib by rxstomp
Refactor broker connection and rpc for client and node services
Remove load balancing from the client since it's handled by rabbitmq
Update docker and kubernetes deployment configuration
Clean server unused files and fix integration tests
Update documentation
Merge Ivars work


src/client/src/rt-system/__tests__/connectionStream.test.ts
```diff
@@ -18,184 +19,51 @@ describe('createConnection$', () => {
...
-    createConnection$(mockConnection)
+    connectionStream$(mockConnection)
 
     // assert
     expect(mockConnection.open).toHaveBeenCalledTimes(0)
     expect(mockConnection.close).toHaveBeenCalledTimes(0)
     expect(mockConnection.onopen).toHaveBeenCalledTimes(0)
     expect(mockConnection.onclose).toHaveBeenCalledTimes(0)
   })
-
```
rikoe, line 33: Do we need equivalents for those tests or did they only apply to Autobahn?

audreyf1: Well spotted, I need to do something about this test.
With stomp, it works differently than Autobahn which provided an AutobahnSession class.
Now, we don't have a session with stomp, we just activate an endpoint with the broker address and port and we make rpc requests or rpc stream requests to this endpoint.

PR #1510, opened 2020-03-26
feat(client): user modal and dropdown added to header (ARTP-1058)

ramirozap:
Kapture 2020-03-26 at 13 19 35


fsarda:
I have only one comment but is more related to UX @ramirozap @edclayforthcarr. How can i display to select theme when there is no user logged in?
image

PR #1526, opened 2020-04-10
Minimum Viable Progressive Web App

pmilla1606:
This PR:

- Adds a service worker with the minimum requirements for installability and registers it.
- Updates the manifest.json to include higher res icons to meet installability requirements.

**Important**:
- To get this to work in Chrome, you need to enable chrome://flags/#bypass-app-banner-engagement-checks
- Edge (Chromium) should work out of the box

RTC Running in `standalone` mode
Screen Shot 2020-04-10 at 8 18 20 AM

Chrome default install prompt. We can offer our own install button/prompt/flow if we want.
Screen Shot 2020-04-10 at 8 21 53 AM

`chrome://apps` showing Reactive Trader
Screen Shot 2020-04-10 at 8 18 43 AM

### DIFF
--- a/src/client/src/serviceWorker.js
+++ b/src/client/src/serviceWorker.js
@@ -31,7 +31,6 @@ export function register(config) {
 
     window.addEventListener('load', () => {
       const swUrl = `${process.env.PUBLIC_URL}/rtc-sw.js`
-      console.log(process.env, isLocalhost, swUrl, config)
       if (isLocalhost) {
         // This is running on localhost. Let's check if a service worker still exists or not.
         checkValidServiceWorker(swUrl, config)
--- a/src/client/src/serviceWorker.js
+++ b/src/client/src/serviceWorker.js
@@ -30,8 +30,8 @@ export function register(config) {
     }
 
     window.addEventListener('load', () => {
-      const swUrl = `${process.env.PUBLIC_URL}/service-worker.js`
-
+      const swUrl = `${process.env.PUBLIC_URL}/rtc-sw.js`
+      console.log(process.env, isLocalhost, swUrl, config)
       if (isLocalhost) {
         // This is running on localhost. Let's check if a service worker still exists or not.
         checkValidServiceWorker(swUrl, config)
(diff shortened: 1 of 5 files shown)


src/client/src/serviceWorker.js
```diff
@@ -55,7 +54,7 @@ export function register(config) {
 function registerValidSW(swUrl, config) {
   navigator.serviceWorker
     .register(swUrl)
-    .then(registration => {
+    .then((registration) => {
```
rikoe, line 57: Minor: Is there a reason you changed these parameters to be wrapped in brackets @pmilla1606, or was that done by prettier? I can't remember what our convention is, but I prefer single parameters not to have brackets!

src/client/src/serviceWorker.js
```diff
@@ -30,8 +30,7 @@ export function register(config) {
     }
```
rikoe, line 30: Is there a reason this file is JavaScript and not TypeScript?

rikoe: Also, should this be in a PWA-specific "platform", as opposed to the main codebase? Not sure tbh. It's probably here for a reason.

src/client/src/serviceWorker.js
```diff
@@ -30,8 +30,7 @@ export function register(config) {
...
 
     window.addEventListener('load', () => {
-      const swUrl = `${process.env.PUBLIC_URL}/service-worker.js`
-
+      const swUrl = `${process.env.PUBLIC_URL}/rtc-sw.js`
```
rikoe, line 33: What is the difference bewteen service-workes.js and rtc-sw.js @pmilla1606? Do we need both?

pmilla1606: `serviceWorker.js` is the default created by `create-react-app` which is in charge of managing service workers. `rtc-sw.js` is a minimal service worker that will allow us to meet the installability requirements. I think it's a good idea to keep them separate especially if we extend this in the future.

src/client/src/serviceWorker.js
pmilla1606: JS because it's not compiled as part of the our source. It's in the public directory (same as the create-react-app default one).

src/client/src/serviceWorker.js
pmilla1606: yup - that was prettier!

PR #1529, opened 2020-04-14
Artp 861 rabbitmq client

audreyf1:


src/client/src/rt-components/loadable/Loadable.tsx
```diff
@@ -44,9 +44,9 @@ const Loadable: React.FC<Props> = ({
   minHeight,
 }) => {
   useEffect(() => {
-    onMount && onMount()
+    if (status === ServiceConnectionStatus.CONNECTED) onMount && onMount()
```
rikoe, line 47: I am not sure I quite parse this statement?

The `onMount` is already a single-line if statement, as in `if (onMount !== undefined) onMount()`

Should it be `if ((status === ServiceConnectionStatus.CONNECTED) && onMount) onMount()`?

src/client/src/rt-system/ServiceStub.ts
```diff
@@ -39,14 +39,16 @@ export class ServiceStub {
...
+  createRequestResponseOperation<TResponse, TPayload>(
+    service: string,
+    operationName: string,
     payload: TPayload,
   ): Observable<TResponse> {
     const dto: SubscriptionDTO<TPayload> = {
       payload,
       Username: this.userName,
     }
+    const remoteProcedure = service + '.' + operationName
```
rikoe, line 51: Here and below two different string concatenation is used. Can we make it consistent please?

I prefer `` const remoteProcedure = `${service}.${operationName}\` `` I think.

src/client/src/rt-components/loadable/Loadable.tsx
audreyf1: yes, I think you're right

src/client/src/rt-components/loadable/Loadable.tsx
```diff
@@ -44,9 +44,9 @@ const Loadable: React.FC<Props> = ({
   minHeight,
 }) => {
   useEffect(() => {
-    onMount && onMount()
+    if (status === ServiceConnectionStatus.CONNECTED && onMount) onMount()
```
rikoe, line 47: I think you will have an issue with operator precedence here: `===` has lower precedence than `&&` I think. See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Operator_Precedence

audreyf1: Interesting link, thanks Riko.
In the table === appears before && which means it has highest precedence than && ?
But to make it clearer I can add parenthesis.

audreyf1: Prettier removes the additional parenthesis, so it means it's advised not to put them when it's not necessary...

rikoe: doh, I probably have it the wrong way round then. yup, I read it upside down 🤦

PR #1532, opened 2020-04-15
ARTP-1115: fix buy/sell buttons color on tile hover

rafgarciaa:
ezgif com-video-to-gif

Updated spot tile state on hover in both light and dark theme


rikoe:
I don't actually think this was the ask. The issue is that the background of the tile on hover should be light, not dark. See https://web-demo.adaptivecluster.com for the correct behaviour on tile hover.

tile hover

PR #1556, opened 2020-04-28
ARTP-1130: Remove reactive trader in browser header

rafgarciaa:
Screen Shot 2020-04-28 at 3 33 10 PM

### DIFF
--- a/src/client/src/apps/MainRoute/components/app-header/Header.tsx
+++ b/src/client/src/apps/MainRoute/components/app-header/Header.tsx
@@ -4,10 +4,6 @@ import { styled } from 'rt-theme'
 import LoginControls from './LoginControls'
 import Logo from './Logo'
 import ThemeSwitcher from './theme-switcher'
-import { isFinsemble, isOpenFin, isGlue42, isSymphony } from 'rt-platforms/getPlatformAsync'
-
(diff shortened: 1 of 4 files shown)


src/client/src/apps/MainRoute/components/app-header/Header.tsx
```diff
@@ -4,6 +4,9 @@ import { styled } from 'rt-theme'
 import LoginControls from './LoginControls'
 import Logo from './Logo'
 import ThemeSwitcher from './theme-switcher'
+import { isFinsemble, isOpenFin, isGlue42, isSymphony } from 'rt-platforms/getPlatformAsync'
+
+const isBrowser = !isFinsemble && !isOpenFin && !isGlue42 && !isSymphony
```
rikoe, line 9: Can we please move this check into the platform module so that we don’t need to import everything here?

In fact we may need to reverse this. I am not 100% sure, but I think we may need to do this the other way around. I.e. only show Reactive Trader in the title bar for the OpenFin platform, and don’t show it in the other cases.

In other words the title UX element should only exist and be added in the OpenFin platform code.

I need to take another look, and confirm what the app will look like under all the platforms. 

@pmilla1606 , @rafgarciaa does this make sense?

PR #1568, opened 2020-05-05
bug(finsemble): fixes bug preventing finsemble from loading (ARTP-1136)

pmilla1606:
Screen Shot 2020-05-05 at 3 31 39 PM

- Renames `isParentAppLauncher` to `isParentAppOpenfinLauncher` for clarity
- Creates `/launcherUtils` to contain `isParentAppOpenfinLauncher` since the exports of `/openfin` were causing the error

### DIFF
--- a/src/client/src/rt-platforms/openFin/adapter/launcherUtils.ts
+++ b/src/client/src/rt-platforms/openFin/adapter/launcherUtils.ts
@@ -1,0 +1,8 @@
+export async function isParentAppOpenfinLauncher() {
+  const app = await fin.Window.getCurrent()
+  const {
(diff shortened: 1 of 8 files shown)


src/client/src/rt-platforms/openFin/adapter/launcherUtils.ts
```diff
@@ -0,0 +1,8 @@
+export async function isParentAppOpenfinLauncher() {
+  const app = await fin.Window.getCurrent()
```
rikoe, line 2: @pmilla1606  should we add an extra check for `fin` here, since it’s not always called from an OpenFin context?

PR #1578, opened 2020-05-07
Update RTC, RTA and Launcher Icons

pmilla1606:


pmilla1606:
@rikoe thanks - updated the shortcut icons

PR #1581, opened 2020-05-08
fix(build): make chartiq and token optional

audreyf1:
In the .npmrc I didn't manage to make the environment NPM_TOKEN optional. I wanted to have .npmrc ignored by git so that it would only exist upstream but then the client docker file that is using it fails. When people get the latest master their npm will complain about this NPM_TOKEN variable not defined...

And npm ci fails even if chartiq is marked as optional in package-lock.json. 
npm install works better but fails on react-script.

### DIFF
--- a/src/client/Dockerfile
+++ b/src/client/Dockerfile
@@ -7,11 +7,11 @@ ARG REACT_APP_BUILD_VERSION
 
 COPY .npmrc .npmrc
 COPY ./package*.json ./
-RUN npm ci
+RUN npm install
 RUN rm -f .npmrc
 COPY . .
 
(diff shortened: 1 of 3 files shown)


src/client/Dockerfile
```diff
@@ -7,7 +7,7 @@ ARG REACT_APP_BUILD_VERSION
 
 COPY .npmrc .npmrc
 COPY ./package*.json ./
-RUN npm ci
+RUN npm install
```
audreyf1, line 10: npm ci fails when trying to download the private package. I don't know how to fix this so I put back npm install at the moment...

PR #1608, opened 2020-05-27
ARTP-1174: add area specific title to popped out windows

rafgarciaa:
Screen Shot 2020-05-27 at 12 58 55 AM


src/client/src/rt-util/getAppName.ts
```diff
@@ -1,10 +1,31 @@
 import { getEnvironment } from 'rt-util'
+import { currencyFormatter } from './'
+import { appTitles } from '../index'
```
rikoe, line 3: I am not sure we should import the app titles here? Can we just move them here?

rafgarciaa: Yeah I can move it all here @rikoe

PR #1614, opened 2020-06-02
Glue42 Core

pmilla1606:
This PR adds Glue42 Core to RTC.

- Only load G42C when using query param: `~/?waitFor=GLUE_CORE`
- Adds new Platform type
- Adds new Platform route
- Adds Glue CLI and CLI output (story needed for including this in the our build process)


src/client/package.json
```diff
@@ -67,9 +71,9 @@
     "openfin-notifications": "0.11.1",
     "polished": "^3.4.1",
     "query-string": "^6.8.2",
-    "react": "^16.8.6",
+    "react": "^16.13.1",
     "react-copy-to-clipboard": "^5.0.1",
-    "react-dom": "^16.8.6",
+    "react-dom": "^16.13.1",
```
rikoe, line 76: Should we revert the react dependencies?

src/client/src/rt-platforms/glue/glueCore.ts
```diff
@@ -0,0 +1,91 @@
...
+    /**
+     * In order to integrate Glue42 with channels the clicked symbol needs to be published to the channel.
+     */
+    publishToChannel: (symbol: string) => {
+      try {
+        window.glue.channels.publish({ symbol: symbol.replace('/', '') })
+      } catch (err) {
+        console.warn(err.message)
+      }
+    },
```
rikoe, line 69: I think we said remove this bit for now?

src/client/src/rt-platforms/glue/glueCore.ts
```diff
@@ -0,0 +1,91 @@
...
+  }
+
+  notification = {
+    notify: (message: object) => {
+      if ('Notification' in window) {
+        if (Notification.permission === NotifyPermission.granted) {
+          sendNotification(message)
+        }
```
rikoe, line 77: How about the notification, does it actually work, or shall we remove it for now as well?

PR #1627, opened 2020-06-18
ARTP-1190: disable snap and dock from launcher config

rafgarciaa:
Disables the snap and dock feature from the launcher only. Main application and its children still have snap and dock capabilities.

ezgif com-video-to-gif (1)


rafgarciaa:
@rikoe I think prettier removed the extra white space

PR #1630, opened 2020-06-23
feat(platform): set new pwa platform api

rafgarciaa:


src/client/src/rt-platforms/types.ts
```diff
@@ -1,5 +1,5 @@
 export type PlatformName = 'browser' | 'openfin' | 'finsemble'
-export type PlatformType = 'browser' | 'desktop'
+export type PlatformType = 'browser' | 'desktop' | 'pwa'
```
rikoe, line 2: Is this right? Based on the platforms we currently have, I would expect to see browser, openfin, finsemble, glue42, glue42core for platform name.

I think type is still just browser and desktop, as PWA is just another version of a desktop app?

rafgarciaa: @rikoe yup, removed `pwa` and set it to `desktop`

PR #1632, opened 2020-07-03
fix(client): pwa display on ios (ARTP-1228)

steinalex:
**Changes:**
- Remove the iOS banner on a PWA instance
- Remove code for Android specific modal (as we now install via native Android)

Tested on iPad Mini and Android 10

### DIFF
--- a/src/client/src/apps/MainRoute/components/app-header/PWAInstallPrompt/PWALaunchButton.tsx
+++ b/src/client/src/apps/MainRoute/components/app-header/PWAInstallPrompt/PWALaunchButton.tsx
@@ -1,14 +1,15 @@
 import React, { Dispatch, SetStateAction } from 'react'
-import { PWABanner, InstallButton, MobileDevice } from './PWAInstallPrompt'
+import { PWABanner, InstallButton } from './PWAInstallPrompt'
 import { usePWABannerPrompt } from './usePWABannerPrompt'
+import { isiOS } from 'apps/utils'
 interface InstallLaunchProps {
   state: string | null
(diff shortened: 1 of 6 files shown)


src/client/src/rt-system/environment.ts
```diff
@@ -17,8 +17,9 @@ export default class Environment {
 
   static isPWA() {
     return (
-      window.matchMedia('(display-mode: standalone)').matches ||
-      (window.navigator as Navigator).standalone
+      (window.matchMedia('(display-mode: standalone)').matches ||
```
steinalex, line 20: Updated this function to return a `boolean`

src/client/src/apps/MainRoute/components/app-header/PWAInstallPrompt/PWALaunchButton.tsx
```diff
@@ -1,14 +1,15 @@
...
 interface InstallLaunchProps {
   state: string | null
   setIsModalOpen: Dispatch<SetStateAction<boolean>>
 }
 export const PWALaunchButton: React.FC<InstallLaunchProps> = ({ state, setIsModalOpen }) => {
-  const [prompt, promptToInstall, device] = usePWABannerPrompt()
+  const [prompt, promptToInstall] = usePWABannerPrompt()
 
-  if (!prompt && device === MobileDevice.iOS && state === PWABanner.Hidden) {
+  if (!prompt && isiOS && state === PWABanner.Hidden) {
```
steinalex, line 12: The initial plan was to have an `Android` specific modal, but then I figured how to install the native way. Hence the only test now is `iOS`

PR #1635, opened 2020-07-09
Data tag to indicate most recent blotter row (ARTP-1236)

iowayankee:


src/client/src/rt-types/trade.ts
```diff
@@ -11,6 +11,7 @@ export interface Trade {
   valueDate: Date
   status: any
   highlight?: boolean
+  latest?: boolean
```
rikoe, line 14: We are extending the data model with properties only required for testing, I am not sure this is ideal?

Ideally if we can restrict testing annotations to just the data-qa attributes that would be preferable.

src/client/src/apps/MainRoute/widgets/blotter/components/Blotter.tsx
```diff
@@ -62,7 +62,8 @@ const getRowClass = ({ data }: { data: Trade }) => {
 }
 
 const Blotter: React.FC<BlotterProps> = props => {
-  const { canPopout, rows } = props
+  let { canPopout, rows } = props
+  rows = rows.map((a, i) => (i === 0 ? { ...a, latest: true } : a))
```
rikoe, line 66: What is the impact of this extra step on performance?
Is there a way (1) to use react key attributes to identify the right row (2) if there is no other way to do this, is it possible to enable this code while tests are running only, and not in production?

steinalex:
Taking a look at the generated DOM, there's a few things we could target that are generated by `ag-grid`.

By using a query selector, you could query `col-id='trade-id` followed by another specific selector. A query selector will always return the first element it matches, so in this case will always be the top row :confetti_ball: 

Give it a try and see what you think: `document.querySelector("[col-id='tradeId']" + ".ag-cell").textContent`

Martin should just be able to use this selector this within his test, so no code change needed

PR #1636, opened 2020-07-14
Blotter

MartinLavigneAdaptive:


src/e2e/src/pages/blotter.component.ts
```diff
@@ -1,70 +1,53 @@
...
+    return trade
+  }
+
+  async getTradeDirection(num: string): Promise<ElementFinder> {
+    const trade = await this.root.element(by.qa(`${num}-direction`))
+    return trade
+  }
+
+  async getLatestTrade() {
+    const latestTrade = this.root.element(by.css("[col-id='tradeId']" + '.ag-cell'))
```
iowayankee, line 48: Nice!

src/e2e/src/pages/blotter.component.ts
```diff
@@ -1,70 +1,98 @@
...
-          )
-        ),
-        tradeBackGroundColour: root.element(
-          by.css(
-            '[data-qa="shell-route__blotter-wrapper"] .ag-body .ag-row:nth-child(1) [col-id="statusIndicator"]'
-          )
-        )
-      }
-    }
+    // this.tradesTable = {
```
steinalex, line 27: Remove the commented code please :smile:

src/e2e/src/pages/blotter.component.ts
```diff
@@ -1,70 +1,98 @@
...
+  }
+
+  async getTradeDealtCCY(num: string): Promise<ElementFinder> {
+    const trade = await this.root.element(by.qa(`${num}-dealtCurrency`))
+    return trade
+  }
+
+  async getTradeNotional(num: string): Promise<ElementFinder> {
+    const trade = await this.root.element(by.qa(`${num}-notional`))
+    return trade
```
steinalex, line 79: To make this even cleaner, you could just return the await without the `const` assignment.

`return await this.root.element(by.qa(`${num}-notional`))`

Same can be applied to the other functions

src/e2e/src/pages/blotter.component.ts
```diff
@@ -1,70 +1,98 @@
...
 import { mapAsync } from '../utils/async.utils'
-import { waitForElementToBePresent } from '../utils/browser.utils'
+import { waitForElementToBePresent, waitForElementToBeVisible } from '../utils/browser.utils'
 
 export interface BlotterHeader {
   id: string
   label: string
 }
 
+type TradeProperties =
```
steinalex, line 10: If we're no longer using the constructor, these typings can be removed, as well as on line `24`

MartinLavigneAdaptive: This is actually used by a utils file, so this specific comment won't be addressed

PR #1638, opened 2020-07-22
feat: added cpu and memory requests for each pod ARTP-1112

mtolla:
Added requests specifications for each pod. 
I had some issues setting the limits while using Kubernetes locally. The CPU usages I faced are higher from the one in our deployments and the pods were slow and tend to restart without hitting the memory limit.

### DIFF
--- a/src/services/kubernetes/broker-deployment.yaml
+++ b/src/services/kubernetes/broker-deployment.yaml
@@ -21,6 +21,6 @@ spec:
             - containerPort: 15674
           resources:
             requests:
-              memory: "80M"
+              memory: "120M"
               cpu: "100m"
 
--- a/src/services/kubernetes/broker-deployment.yaml
+++ b/src/services/kubernetes/broker-deployment.yaml
@@ -19,3 +19,8 @@ spec:
             - containerPort: 5672
             - containerPort: 15672
             - containerPort: 15674
+          resources:
+            requests:
+              memory: "80M"
+              cpu: "100m"
+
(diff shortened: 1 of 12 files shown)


src/services/kubernetes/broker-deployment.yaml
```diff
@@ -19,3 +19,8 @@ spec:
             - containerPort: 5672
             - containerPort: 15672
             - containerPort: 15674
+          resources:
+            requests:
+              memory: "80M"
+              cpu: "100m"
```
rikoe, line 25: Should we give the broker more resources? It is the main bottleneck.

src/services/kubernetes/web-deployment.yaml
```diff
@@ -17,3 +17,7 @@ spec:
           image: reactivetradercloud/nginx:${BUILD_VERSION}
           ports:
             - containerPort: 3000
+          resources:
+            requests:
+              memory: "10M"
+              cpu: "15m"
```
rikoe, line 23: Why so little memory for the client, is it because nginx doesn't need a lot?

mtolla: In our deployments, we normally register values under 10MB except for some spikes around 15 to 25. I think this one is correct

src/services/kubernetes/broker-deployment.yaml
mtolla: You are right, the broker memory parameter is too low, let me fix that. The CPU value is a little bit higher, I think I swapped the values while editing 😅 . 
I could leave the CPU to 100, in that way we are always sure we will have the resources.

PR #1642, opened 2020-08-03
fix(client): add fdc3 api to app config (ARTP-1269)

steinalex:
**Changes:**
- Add FDC3 API to application config

**Notes:**
Tested on Linux Ubuntu and Windows 10

### DIFF
--- a/src/client/src/apps/utils/openfin-utils.ts
+++ b/src/client/src/apps/utils/openfin-utils.ts
@@ -33,7 +33,7 @@ export async function createOrBringToFrontOpenFinApplication({
   name,
   url,
   uuid,
-  windowOptions
+  windowOptions,
 }: ApplicationConfig): Promise<Application> {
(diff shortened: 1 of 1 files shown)


src/client/src/apps/utils/openfin-utils.ts
```diff
@@ -47,14 +47,15 @@ export async function createAndRunOpenFinApplication({
...
+  windowOptions,
 }: ApplicationConfig): Promise<Application> {
   const appOptions: fin.ApplicationOption = {
     name,
     url,
     uuid: uuid || name,
     nonPersistent: true,
-    mainWindowOptions: windowOptions
+    mainWindowOptions: windowOptions,
+    fdc3Api: true,
```
rikoe, line 58: Is it possible to add this option per application in applicationConfigurations.tsx instead, so that it's more explicit?

It also means it wouldn't be automatically enabled for all openfin apps, we probably want to do it on an opt-in basis.

PR #1643, opened 2020-08-05
ARTP-1109 Replace nsgate with ingress

mtolla:
Added configuration that will replace `nsgate` with multiple Kubernetes ingresses (one for each namespace).
Since we will have multiple ingresses, I created a pre-shared certificate for the TLS certificate.


sergei-ivanov:
I'd advise a cautious approach: deploy the ingress in all environments, flip the DNS entries to point to the ingress public endpoint, leave it there for some time, then if no issues surface, undeploy nsgate and remove from the codebase for good.

PR #1654, opened 2020-09-29
Artp 1302 popped out tiles dont have tabs

spcfran:
Torn off windows now look like this:

image

Chrome styles (buttons and and title font) will be addressed in a future ticket


rikoe (approved):
This is great! I think we should try and get to a point where we can use the environment specific titles like in the current version,  but it's absolutely fine to do that in a separate PR:

image

PR #1656, opened 2020-10-03
Chore  improve sc debugging experience

spcfran:
`styled-components` includes a babel macro that makes the styled components to have meaningful names based on the variable they are assigned to. 

This makes for a much cleaner React dev tools component tree as well as class names on the DOM nodes.

This PR contains the setup for that to work, as well as replaces all imports of the proxyed `styled` from `rt-theme` with direct imports from `styled-components/macro`, necessary for the macro to work.


src/client/src/rt-theme/index.ts
```diff
@@ -12,8 +11,11 @@ export { themes, getThemeColor } from './themes'
 export type { Theme, TouchableIntentName, ColorProps, ThemeSelector } from './themes'
 export { colors } from './colors'
 export { default as GlobalStyle } from './globals'
-
-export const styled: ThemedStyledInterface<Theme> = baseStyled
-
 export { ThemeName, ThemeProvider, ThemeConsumer, useTheme } from './ThemeContext'
 export { default as ThemeStorageSwitch } from './ThemeStorageSwitch'
+
+// Make all styled-component functions (e.g. `styled`, `css`) typed with Theme
```
spcfran, line 17: This is the current recommended way to get typed interpolations in styled components, rather than the proxy approach we previously had

rikoe (approved):

PR #1663, opened 2020-10-13
ARTP-1313 Tile scrolling behavior fix

mhamdar:
The simple fix here seems to be to use parent element height as opposed to viewport in OverflowScroll (vh -> %). Scrolling now works as expected and I don't think this change causes unexpected side-issues.

### DIFF
--- a/src/client/src/apps/MainRoute/routes/styled.tsx
+++ b/src/client/src/apps/MainRoute/routes/styled.tsx
@@ -4,7 +4,7 @@ import { rules } from 'rt-styleguide'
 export const OverflowScroll = styled.div`
   overflow-y: scroll;
   ${rules.touchScroll};
-  height: 100%;
+  height: 100%
 `
 
 export const Wrapper = styled.div`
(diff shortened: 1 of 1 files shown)


src/client/src/apps/MainRoute/routes/styled.tsx
```diff
@@ -4,7 +4,7 @@ import { rules } from 'rt-styleguide'
 export const OverflowScroll = styled.div`
   overflow-y: scroll;
   ${rules.touchScroll};
-  height: 100vh;
+  height: 100%;
```
rikoe, line 7: Have you tested this both in the browser and OpenFin? I think this change was made to make it work in OpenFin.

mhamdar: It seems to work in OpenFin as well, though there is a separate issue of the scrollbar not appearing at an early enough width,

3977ef01a4169d2fe308da89507c2e71
/

mhamdar: Works on both platforms now:

7e508e0520d938176d1f2de64af1b69f

rikoe (approved):
👏

PR #1665, opened 2020-10-14
fixed tradebutton colors

Mark-Adaptive:
The lines that I removed where overwriting the pre-existing hover behavior of the trade buttons which is the desired behavior (sell button turns red on hover and buy button turns blue on hover). I simply removed those lines and now the desired behavior is achieved according to the zeplin mocks. 
image
image

### DIFF
--- a/src/client/src/apps/MainRoute/widgets/spotTile/components/PriceButton/styled.tsx
+++ b/src/client/src/apps/MainRoute/widgets/spotTile/components/PriceButton/styled.tsx
@@ -54,20 +54,13 @@ export const TradeButton = styled.button<{
   pointer-events: none;
     `
       : `
-  .spot-tile:hover & {
-    background-color: ${theme.core.darkBackground};
-  }
-  .spot-tile:hover &:hover {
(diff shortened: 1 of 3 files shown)


rikoe:
@Mark-Adaptive looks like the unit tests are failing. The snapshot used for the test may need to be updated to reflect your changes...

Mark-Adaptive:
> @Mark-Adaptive looks like the unit tests are failing. The snapshot used for the test may need to be updated to reflect your changes...

Oh I see. Do you happen to know the name of the testing file that I need to update?

rikoe:
> Oh I see. Do you happen to know the name of the testing file that I need to update?

Looks like src/apps/MainRoute/widgets/spotTile/components/Tile/Tile.test.tsx - it's in the failed build logs.

Same thing should happen when you run the tests locally (which is recommend before each PR!)

Mark-Adaptive:
I updated the snapshots and pushed the change and the testing seems to pass on my end but fails on github. Not sure why. 
image

rikoe (approved):

PR #1666, opened 2020-10-16
Add trademark

Mark-Adaptive:
I am unable to access the Launcher so I'm not sure if the last two screenshots on the Kanban card have the TM logo added but if they draw from the same place as the others (getAppName()) then they should work as well. I also added TM to every instance of the word "Reactive Trader" in the codebase. I'm not sure if this is completely necessary and I can revert it back if need be. I believe the only relevant area that needed to be changed was in getAppName() and maybe one other place. 

Here are the screenshots from my browser. You can see the TM reflected in the pop-out browser header at the very top: 
image
image
image
image
image
image

### DIFF
--- a/src/client/src/rt-platforms/glue/reactive-trader.json
+++ b/src/client/src/rt-platforms/glue/reactive-trader.json
@@ -22,7 +22,7 @@
   {
     "title": "Reactive Trader™ Analytics",
     "type": "window",
-    "name": "reactive-trader-™-analytics",
+    "name": "reactive-trader-analytics",
     "hidden": true,
     "details": {
       "url": "http://localhost:3000/analytics",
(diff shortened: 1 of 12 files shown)


src/client/src/rt-platforms/glue/reactive-trader.json
```diff
@@ -1,8 +1,8 @@
 [
   {
-    "title": "Reactive Trader",
+    "title": "Reactive Trader™",
     "type": "window",
-    "name": "reactive-trader",
+    "name": "reactive-trader-™",
```
rikoe, line 5: These `name`s are identifiers I think, they don't need the TM unless they will actually be displayed

rikoe (changes requested):

src/client/src/rt-platforms/glue/reactive-trader.json
Mark-Adaptive: Okay, just removed TM from all the "name" fields.

rikoe (approved):

PR #1667, opened 2020-10-21
fix(client): trademark-to-appear-on-tab (ARTP-1316)

Mark-Adaptive:
I fixed the issue with the TM not appearing in the browser tab. I believe MainRoute.tsx was the culprit but I added TM to a few more areas in the code just in case there are other cases were TM does not appear. 
image

### DIFF
--- a/src/client/src/apps/SimpleLauncher/SimpleLauncher.tsx
+++ b/src/client/src/apps/SimpleLauncher/SimpleLauncher.tsx
@@ -82,7 +82,7 @@ export const SimpleLauncher: React.FC = () => {
             <PricingServiceProvider value={pricingService}>
               <ReferenceDataProvider value={referenceData}>
                 <PlatformProvider value={platform}>
-                  <Helmet title={getAppName('Reactive Launcher')} />
+                  <Helmet title={getAppName('Reactive Launcher™')} />
                   <Launcher />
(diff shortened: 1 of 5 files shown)


src/client/src/apps/SimpleLauncher/SimpleLauncher.tsx
```diff
@@ -82,7 +82,7 @@ export const SimpleLauncher: React.FC = () => {
             <PricingServiceProvider value={pricingService}>
               <ReferenceDataProvider value={referenceData}>
                 <PlatformProvider value={platform}>
-                  <Helmet title={getAppName('Reactive Launcher')} />
+                  <Helmet title={getAppName('Reactive Launcher™')} />
```
rikoe, line 85: Remove this one is only the phrase "Reactive Trader" has a trademark 😄

rikoe (changes requested):

src/client/src/apps/SimpleLauncher/SimpleLauncher.tsx
Mark-Adaptive: Haha my bad, got rid of the TM on that line. Should be good now.

rikoe (approved):

PR #1678, opened 2020-10-26
feat(browser): use portals for tile popouts (ARTP-1317)

mhamdar:
This is an attempt at using React Portals to handle tile pop outs, avoiding the loading time needed with the current solution. I don't consider this merge-ready but I am making a draft PR so that the general approach can be discussed.
676a7993340ad61b623376688e58a67e

This approach maintains tile view / theme sync across windows. Some current outstanding issues:
1) I'm not sure how we want to handle the address bar:
4f56ba39df963c4795ebaa315923002e
Since we wouldn't be using routes anymore, we don't actually need anything here, but it doesn't seem right to leave it as 'about:blank' either.

2) There is currently a bug that you can reproduce as follows: pop out the entire tile workspace, and from that popped out window, pop out an individual tile. Now, if you pop the workspace back in, the popped-out tile stops updating, but its window doesn't automatically close. This also applies to closing the main / full application view while a tile is popped out.

### DIFF
--- a/src/client/src/rt-components/tear-off/TearOff.tsx
+++ b/src/client/src/rt-components/tear-off/TearOff.tsx
@@ -190,8 +190,9 @@ const TearOff: React.FC<TearOffProps> = ({ render, externalWindowProps, tornOff,
   }
 
   if (tornOff) {
+    console.info('extWP', externalWindowProps!.config!.name!)
     // use portal only for browser tiles
-    if (['LiveRates', 'analytics'].includes(externalWindowProps!.config!.name!)
+    if (['LiveRates', 'analytics', 'blotter'].includes(externalWindowProps!.config!.name!)
       || platform.name !== 'browser') {
       return (
         <ExternalWindow onUnload={popIn} {...externalWindowProps} />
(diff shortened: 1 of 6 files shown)


src/client/src/rt-components/tear-off/TearOff.tsx
```diff
@@ -1,11 +1,22 @@
...
+import React, { useCallback, useRef, useState, useEffect } from 'react'
+import ReactDOM from 'react-dom'
 import ExternalWindow, { ExternalWindowProps } from './ExternalWindow'
 import styled from 'styled-components/macro'
 import { useDispatch } from 'react-redux'
 import { usePlatform } from 'rt-platforms'
 import { LayoutActions } from 'rt-actions'
+import { RouteWrapper } from 'rt-components'
+
+const SpotTileStyle = styled.div`
```
stewart-adaptive, line 10: TearOff should be generic and not contain anything specific to the wrapped component, I dont think spot tile stuff should be here

src/client/src/rt-components/tear-off/TearOff.tsx
```diff
@@ -102,7 +190,24 @@ const TearOff: React.FC<TearOffProps> = ({ render, externalWindowProps, tornOff,
   }
 
   if (tornOff) {
-    return <ExternalWindow onUnload={popIn} {...externalWindowProps} />
+    console.info('extWP', externalWindowProps!.config!.name!)
+    // use portal only for browser tiles
+    if (['LiveRates', 'analytics', 'blotter'].includes(externalWindowProps!.config!.name!)
```
stewart-adaptive, line 195: Shouldn't this cover all popouts not just spot tiles?

PR #1681, opened 2020-10-28
fix: paths for new subwindows opened from launcher (ARTP-1323)

stewart-adaptive:
host url prefix is added by the openfin-platform window.ts we were adding it twice

headers were not correct due to missing openfin-sub-window-frame in launchers defaultWindowOptions


chuck-adaptive (approved):
👍

PR #1682, opened 2020-10-28
ARTP-1326 disable RFQ when max exceeded

mhamdar:
Disabling RFQ booking when the user has entered a notional value that exceeds our hardcoded limit.
e9b5eacee82045dee5cce7f92c336944
6dc888456f4f1fa7798a55d6c421942c


stewart-adaptive:
is it ok to still show the initiate RFQ button in this state?

mhamdar:
> is it ok to still show the initiate RFQ button in this state?

mhamdar:
> is it ok to still show the initiate RFQ button in this state?

I think this is the approach we had discussed (@rikoe ?), what did you have in mind?

chuck-adaptive (approved):
Pending UX decision, 👍

rikoe (approved):

PR #1684, opened 2020-10-29
fix: removed open fin specific code from RouteWrapper and AppLayout

stewart-adaptive:
Removed PlatformHeader from the RouteWrapper, its never rendered as the check for windowType === 'sub' && !isChildView never === true

Moved check of isParentAppOpenfinLauncher into OpenFinSubWindowFrame and deleted from OpenFinWindowFrame
The check is used to determine if headerControlHandlers should show a close or popin button on the subframe windows

subwindow from launcher = close button
subwindow from !launcher = popin button

isParentAppOpenfinLauncher was throwing an exception, replaced check of fin.Window.getCurrent().getParentApplication with window.fin.me.uuid

### DIFF
--- a/src/client/src/rt-components/route-wrapper/RouteWrapper.tsx
+++ b/src/client/src/rt-components/route-wrapper/RouteWrapper.tsx
@@ -1,9 +1,7 @@
-import React, { ReactNode, useEffect, useState } from 'react'
+import React, { ReactNode, useEffect } from 'react'
 import styled from 'styled-components/macro'
-import { Platform, usePlatform, isParentAppOpenfinLauncher } from 'rt-platforms'
-import { getAppName } from 'rt-util'
+import { Platform, usePlatform } from 'rt-platforms'
 import { useTheme, ThemeName } from 'rt-theme'
-import Helmet from 'react-helmet'
 
 const RouteStyle = styled('div')<{ platform: Platform }>`
   width: 100%;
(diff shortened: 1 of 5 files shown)


src/client/src/rt-components/route-wrapper/RouteWrapper.tsx
```diff
@@ -31,27 +29,12 @@ interface RouteWrapperProps {
...
-  const { children, windowType = 'main', title } = props
-  const [fromLauncher, setFromLauncher] = useState<boolean>(false)
+  const { children, windowType = 'main' } = props
   const platform = usePlatform()
   const theme = useTheme()
 
-  const { PlatformHeader, PlatformFooter, PlatformControls, PlatformRoute, window } = platform
-
-  useEffect(() => {
-    isParentAppOpenfinLauncher()
```
rikoe, line 47: @stewart-adaptive this particular bit - `isParentAppOpenfinLauncher` - was added, I think, for the OpenFin windows that you can launch by clicking on the buttons like "Open Live Rates", "Open Trades" and "Open Tile".

In those cases the launcher is directly launching the routes themselves, disconnected from the main app (i.e. they have a "close" button, not a "pop back in" button.

Can you confirm that with this change that functionality is still working?

image

stewart-adaptive: you are right, Ill look into a fix

stewart-adaptive: Updated and fixed close button issue, image

PR #1685, opened 2020-10-30
Artp 1288 make version url clickable

Mark-Adaptive:
I made the version number on the connections dialog visible and clickable to redirect to the github page. I was not exactly sure how the REACT_APP_VERSION .env variable works so I just left it as is in this commit but on my local machine I tested I changed the value of REACT_APP_VERSION (the variable that is being displayed as the version number on the connections dialog) to "v1.5.2" and clicked it and it redirected me properly to the github page. I included screenshots of this as well as the tests passing below. If you want the REACT_APP_VERSION variable modified to something else please let me know and I will change it and send out another commit. I created a few redundant commits in the process of rebasing the code so my apologies for that. 

image
image
image
image

### DIFF
--- a/src/client/.env
+++ b/src/client/.env
@@ -1,6 +1,6 @@
 NODE_ENV=development
 REACT_APP_ENV=default
-REACT_APP_VERSION=$npm_package_version
+REACT_APP_VERSION=v1.5.2
 REACT_APP_NAME=$npm_package_name
 REACT_EDITOR=code
 REACT_APP_GREENKEY_URL=ws://localhost:8888/client/ws/speech
--- a/src/client/.env
+++ b/src/client/.env
@@ -1,6 +1,6 @@
 NODE_ENV=development
 REACT_APP_ENV=default
-REACT_APP_VERSION="$npm_package_version"
+REACT_APP_VERSION=$npm_package_version
 REACT_APP_NAME=$npm_package_name
 REACT_EDITOR=code
 REACT_APP_GREENKEY_URL=ws://localhost:8888/client/ws/speech
--- a/src/client/.env
+++ b/src/client/.env
@@ -1,6 +1,6 @@
 NODE_ENV=development
 REACT_APP_ENV=default
-REACT_APP_VERSION=v1.5.2
+REACT_APP_VERSION=$npm_package_version
 REACT_APP_NAME=$npm_package_name
 REACT_EDITOR=code
 REACT_APP_GREENKEY_URL=ws://localhost:8888/client/ws/speech
--- a/src/client/.env
+++ b/src/client/.env
@@ -1,6 +1,6 @@
 NODE_ENV=development
 REACT_APP_ENV=default
-REACT_APP_VERSION=$npm_package_version
+REACT_APP_VERSION="$npm_package_version"
 REACT_APP_NAME=$npm_package_name
 REACT_EDITOR=code
 REACT_APP_GREENKEY_URL=ws://localhost:8888/client/ws/speech
--- a/src/client/src/apps/MainRoute/layouts/AppLayout.tsx
+++ b/src/client/src/apps/MainRoute/layouts/AppLayout.tsx
@@ -12,15 +12,11 @@ export interface Props {
 }
 
 const AppLayout: React.FC<Props> = ({ before, header, body, footer, after }) => {
-  // TODO - move to openfin-platform
-  //@ts-ignore
-  const isChildView = window.fin && window.fin.me && window.fin.me.isView
-
   return (
     <AppLayoutRoot data-qa="app-layout__root">
       {before}
 
-      {!isChildView && header !== null && <Header controls={header} />}
+      {header !== null && <Header controls={header} />}
 
       <Body>{body}</Body>
 
--- a/src/client/src/apps/MainRoute/widgets/footer-version/FooterVersion.tsx
+++ b/src/client/src/apps/MainRoute/widgets/footer-version/FooterVersion.tsx
@@ -15,10 +15,10 @@ const FooterVersion: FC = () => (
     <Link
       href={
         'https://github.com/AdaptiveConsulting/ReactiveTraderCloud/releases/tag/' +
-        process.env.REACT_APP_ENV
+        process.env.REACT_APP_VERSION
       }
     >
-      {process.env.REACT_APP_ENV}
+      {process.env.REACT_APP_VERSION}
     </Link>
   </Wrapper>
 )
--- a/src/client/src/apps/MainRoute/widgets/footer-version/FooterVersion.tsx
+++ b/src/client/src/apps/MainRoute/widgets/footer-version/FooterVersion.tsx
@@ -1,4 +1,4 @@
-import React, { FC } from 'react'
+import React, { FC, useState } from 'react'
 import styled from 'styled-components/macro'
 
 export const Wrapper = styled.div`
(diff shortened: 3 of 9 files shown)


src/client/src/apps/MainRoute/widgets/footer-version/FooterVersion.tsx
```diff
@@ -6,7 +6,21 @@ export const Wrapper = styled.div`
...
+export const Link = styled.a`
+  color: inherit;
+  text-decoration: inherit;
+`
+const FooterVersion: FC = () => (
+  <Wrapper>
+    <Link
+      href={
+        'https://github.com/AdaptiveConsulting/ReactiveTraderCloud/releases/tag/' +
+        process.env.REACT_APP_VERSION
```
rikoe, line 18: @Mark-Adaptive mmm, I think we need a way to distinguish "non-tagged" versions from tagged ones, otherwise the URL will result in a 404.

e.g. v1.2.1 will exist, but not v1.2.1-13-43gfr33.

Not sure currently how to do this. Three options I can think of at the moment:
1. Pass another environment variable that indicates this.
2. Check if the version URL exists (i.e. doesn't lead to a 404) by calling an API/trying a fetch.
3. Make all links go to the last official version, i.e. extract just the first part of the version number (the v1.2.1 part), which should exist.

Of these, 2 seems the most reliable to me. Let's discuss on standup today. This is the GitHub REST API for querying tags: https://developer.github.com/v3/git/tags/. Since it's a public repo, that should work fine, but not sure how straightforward it is.

In any case - the logic should be:
1. retrieve version from env var
2. check whether version exists as a tag on github
2.1 if true, print version is clickable link
2.2 if false, just print the version

Mark-Adaptive: I have been trying to implement solution 2 but I am receiving the following error when I try to fetch the link: 
"Access to fetch at 'https://github.com/AdaptiveConsulting/ReactiveTraderCloud/releases/tag/v1.5.2' from origin 'http://localhost:3000' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource. If an opaque response serves your needs, set the request's mode to 'no-cors' to fetch the resource with CORS disabled." 
Do you happen to know any work arounds for this?

Mark-Adaptive: Should have the proper behavior now: 
Clickable when version number does exist in the github releases:
image
Not Clickable when it doesn't exist: 
image

src/client/.env
```diff
@@ -1,6 +1,6 @@
 NODE_ENV=development
 REACT_APP_ENV=default
-REACT_APP_VERSION=$npm_package_version
+REACT_APP_VERSION=v1.5.2
```
rikoe, line 3: Can we revert this?

src/client/src/apps/MainRoute/layouts/AppLayout.tsx
```diff
@@ -12,15 +12,11 @@ export interface Props {
...
-  // TODO - move to openfin-platform
-  //@ts-ignore
-  const isChildView = window.fin && window.fin.me && window.fin.me.isView
-
   return (
     <AppLayoutRoot data-qa="app-layout__root">
       {before}
 
-      {!isChildView && header !== null && <Header controls={header} />}
+      {header !== null && <Header controls={header} />}
```
rikoe, line 19: These changes are from a different PR so I am not sure what is going on here? You may need to rebase your changes on the latest ones from master?

src/client/src/apps/MainRoute/widgets/footer-version/FooterVersion.tsx
```diff
@@ -1,12 +1,51 @@
...
   opacity: 0.59;
   font-size: 0.75rem;
 `
+export const Link = styled.a`
+  color: inherit;
+  text-decoration: inherit;
+`
+const FooterVersion: FC = () => {
+  const [versionExists, setVersionExists] = useState<boolean | void>(false)
+  const gitTagExists = async (gitTag: string | undefined) => {
```
rikoe, line 15: I'm not sure embedding and calling this function inside the React component is a great idea. It will probably re-run on render.

What I would suggest is (1) extract the code outside of the react component into a named function (2) call it from within the component if you need, but if you do so, then use a useEffect hook.

However, is it necessary to call this inside the component? Could we do it once when the application starts up? The version is not likely to change once the application is launched!

src/client/src/apps/MainRoute/widgets/footer-version/FooterVersion.tsx
```diff
@@ -1,12 +1,51 @@
...
+    const exists = await fetch(
+      'https://api.github.com/repos/AdaptiveConsulting/ReactiveTraderCloud/releases'
+    )
+      .then(response => response.json())
+      .then(data => {
+        for (let i = 0; i < data.length; i++) {
+          if (data[i].tag_name === gitTag) {
```
rikoe, line 22: This is not wrong, but there are more readable ways to filter arrays in JS. E.g. `Array.prototype.find`.

Also DON'T mix async-await keywords with then and catch promise callbacks. Use one or the other.

src/client/src/apps/MainRoute/widgets/spotTile/components/Tile/Tile.test.tsx
```diff
@@ -30,7 +30,7 @@ const defaultParams: Parameters<typeof getDerivedStateFromUserInput>[0] = {
       mid: 1.48357,
       priceMovementType: 'Down' as PriceMovementTypes,
       symbol: 'EURCAD',
-      valueDate: '2019-04-07T00:00:00Z',
+      valueDate: new Date(2019, 3, 7).toISOString(),
```
rikoe, line 33: These look like more changes that aren't part of your work, but for which the commits have somehow been pulled in? The one below as well?

rikoe (changes requested):

src/client/src/apps/MainRoute/widgets/footer-version/FooterVersion.tsx
Mark-Adaptive: Would you recommend defining the gitTagExists async function outside of FooterVersion and then using a useEffect to get it's value on mount of the FooterVersion component? I did actually attempt to do this earlier but I kept running into this issue that the async function can only return a Promise whereas ideally I'd like it to simply return a boolean (true or false depending on whether the version number exists in the releases) and use that boolean to determine whether to display a clickable link or not. If I were to define the gitTagExists outside of my functional component, how would I then return a boolean true or false from it to use in my functional component? Currently I simply set State within the gitTagExists function but I agree that that isn't the optimal solution.

src/client/src/apps/MainRoute/widgets/footer-version/FooterVersion.tsx
Mark-Adaptive: Oh I see. If I don't use "then" will I be able to simply return a boolean from my async function rather than a promise?

src/client/src/apps/MainRoute/widgets/spotTile/components/Tile/Tile.test.tsx
Mark-Adaptive: Yup seems to be an issue with the rebase I did. I will just create a new branch from my master and copy over my changes to the FooterVersion.tsx file.

src/client/.env
Mark-Adaptive: Yup I'll fix it.

src/client/src/apps/MainRoute/layouts/AppLayout.tsx
Mark-Adaptive: I'll create the new branch from master and only modify FooterVersion.tsx.

PR #1692, opened 2020-11-05
ARTP-1234: Enhance browser notifications

mhamdar:
Similarly to OpenFin, when you click on a trade notification from the browser, highlight that trade in the blotter.


src/client/src/rt-platforms/browser/utils/sendNotification.ts
```diff
@@ -8,18 +8,31 @@ export interface NotificationMessage {
...
+    data: tradeNotification,
   }
 
-  /* eslint-disable @typescript-eslint/no-unused-vars */
-  // @ts-ignore
-  const notification = new Notification(title, options)
+  if (registration) {
+    registration.showNotification(title, options)
+  } else {
+    delete options.actions
```
stewart-adaptive, line 34: can this else be removed, seems to serve no purpose

src/client/src/apps/MainRoute/Router.tsx
```diff
@@ -1,60 +1,97 @@
...
+const { highlightTradeAction } = BlotterActions
+
+export const Router: FC = () => {
+  const dispatch = useDispatch()
+  const switchHighlight = (trade: Trade, on: boolean) => ({
+    ...trade,
+    highlight: on,
+  })
+
+  const trades = useSelector((state: GlobalState) => state.blotterService.trades)
```
stewart-adaptive, line 23: Can highlights be moved to be under the tiles route?

src/client/src/rt-platforms/browser/utils/sendNotification.ts
mhamdar: Instantiating `const notification = ... ` actually displays a notification, so that else block is a fallback mechanism to still display a notification (without blotter highlighting functionality) if the service worker doesn't register for whatever reason.

src/client/src/apps/MainRoute/Router.tsx
mhamdar: I had that same thought, but it seems when the eventListener is only bound to a specific route, the SW message doesn't propagate the way you might expect it to, and the highlight doesn't work.

(There might be a way to do it, but it wouldn't be as simple as just moving over the code there, and I'm not sure what changes would have to be made to get it done).

src/client/src/rt-platforms/browser/utils/sendNotification.ts
```diff
@@ -8,18 +8,31 @@ export interface NotificationMessage {
...
+let registration: ServiceWorkerRegistration
+navigator.serviceWorker &&
+  navigator.serviceWorker.ready.then(reg => {
+    registration = reg
+  })
+
 export const sendNotification = ({ tradeNotification }: NotificationMessage) => {
   const status = tradeNotification.status === 'done' ? 'Accepted' : 'Rejected'
   const title = `Trade ${status}: ${tradeNotification.direction} ${tradeNotification.dealtCurrency} ${tradeNotification.notional}`
   const body = `vs. ${tradeNotification.termsCurrency} \nRate ${tradeNotification.spotRate}    Trade ID ${tradeNotification.tradeId}`
```
rikoe, line 20: Shouldn't the title and body text of the notifications change as per the designs linked in the JIRA?
image
This was the main part of the JIRA

src/client/src/apps/MainRoute/Router.tsx
rikoe: I agree with @stewart-adaptive - this really clutters up the Router file. What I would suggest is if there isn't a clear way to get the highlighting to work for now, it is put on ice at the moment, and updating the design of the notifications (new icon, text, etc.) takes priority.

rikoe (changes requested):

mhamdar:
notif

rikoe (approved):

PR #1701, opened 2020-11-25
Artp 1333 1360

chuck-adaptive:
# Changes

* hide the ReactiveTrader popout icons - these controls must be handled by the OpenFin Platform API. Doing so allows the main window to intelligently fill the space on the main window when a view is popped out.
* decorate platform controls to use branded icons
* Add lock icon to toggle platform tabs/layout lock. Toggling tabs locks the existing layout

There is no "pop-in" functionality in the Platform API. To restore a popped out view, drag it into place on the main window.

This does not effect the spot tile popouts. Those should function the same as they did before.

# Images

image

image

image


src/client/src/rt-platforms/openfin/components/ToggleOpenFinLayoutLock.tsx
```diff
@@ -0,0 +1,30 @@
...
+import styled from 'styled-components/macro'
+import { toggleLockedLayout } from './utils'
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
+import { faLock, faLockOpen } from '@fortawesome/free-solid-svg-icons'
+import { library } from '@fortawesome/fontawesome-svg-core'
+
+library.add(faLock, faLockOpen)
+
+const OpenFinOnlyButton = styled('button')`
+  ${isOpenFin ? '' : 'display: none;'}
```
garyadaptive, line 12: Is this necessary? This is in the OpenFin platform adapter, so presumably isn't rendered in other platforms?

src/client/src/rt-platforms/openfin/components/index.ts
```diff
@@ -3,3 +3,4 @@ export { OpenFinLogo } from './OpenFinLogo'
 export { default as OpenFinBrowserLink } from './OpenFinBrowserLink'
 export { OpenFinWindowFrame } from './OpenFinWindowFrame'
 export { OpenFinSubWindowFrame } from './OpenFinSubWindowFrame'
+export { default as ToggleOpenFinLayoutLock } from './ToggleOpenFinLayoutLock'
```
garyadaptive, line 6: Ah, I missed that this is being exported and rendered in `Header`. That's a leaky abstraction; is it possible for this to be rendered inside `OpenFinSubWindowFrame` or will that involve too big a refactor? At a minimum we should mark this as tech debt and create a ticket.

garyadaptive:
I made a minor comment, but otherwise looks really good. Nice one! I'll give it a thorough test and then I'm happy to approve, pending whatever we agree on the `<ToggleOpenFinLayoutLock />` feedback.

chuck-adaptive:
Closing this PR in favor of the other one, which resolves the above

PR #1704, opened 2020-11-30
fix(launcher): set permissions to launch Limit Checker (ARTP-1363)

garyadaptive:
For reasons I don't yet understand, we need to set the permissions on both the parent and child windows (perhaps the IPC call is delegated via the platform window - I will clarify this with OpenFin). By setting the permission on both windows, the Limit Checker can now be opened.

### DIFF
--- a/src/client/public/openfin/launcher.json
+++ b/src/client/public/openfin/launcher.json
@@ -23,11 +23,6 @@
           "width": 4,
           "height": 4
         },
-        "permissions": {
-          "System": {
-            "launchExternalProcess": true
-          }
-        },
         "saveWindowState": false,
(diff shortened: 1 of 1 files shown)


chuck-adaptive (commented):
I am getting the permissions error when launching

image

garyadaptive:
> I am getting the permissions error when launching
> 
> image

I think that's outside of RT's remit and probably something to do with your local Windows security settings. I also get warnings about unsigned certs, but I get the option to ignore it. Your security settings might be stricter than mine. Either way, I _think_ that error is nothing to do with our code.

chuck-adaptive (approved):

PR #1715, opened 2020-12-10
New tiles for merge

slcfv3:
Tiles part for new architecture


src/new-client/src/services/currencyPairs.ts
```diff
@@ -104,5 +105,9 @@ export {
   useFilteredSymbols,
   useCurrencies,
   onSelectCurrency,
```
iprado78, line 107: Should we standardize on always exporting the whole bind tuple and a naming convention for the export?  e.g., `export const currencyPairRxBinding = bind(...)` - @josepot ?

iprado78: This would be for a later PR, just initiating the discussion.

src/new-client/src/types/Side.ts
```diff
@@ -0,0 +1,4 @@
+export enum Side {
```
iprado78, line 1: I removed this folder because it's a duplicate of the Direction enum we have in the trades service that we can use instead.

src/new-client/.env.development
```diff
@@ -1,4 +1,4 @@
-REACT_APP_BROKER_HOST=web-dev.adaptivecluster.com
+REACT_APP_BROKER_HOST=web-demo.adaptivecluster.com
```
iprado78, line 1: Just keep this as a local change / don't commit it

PR #1722, opened 2020-12-11
fix: use k8s job to populate the event store (ARTP-1183)

garyadaptive:
This removes the eventstore initialisation from the CI build and instead uses the existing (but currently unused) Kubernetes job. This change will mean that the eventstore is seeded at deploy time, against the running instance in k8s, instead of committing a modified image to docker.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -51,7 +51,7 @@ jobs:
         run: |
           # TODO: create new environment automatically
           # Update/create each deployment file in src/services/kubernetes
-          for f in ./src/services/kubernetes/*.yaml; do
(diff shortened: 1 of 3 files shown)


.github/workflows/deploy.yml
```diff
@@ -51,7 +51,7 @@ jobs:
         run: |
           # TODO: create new environment automatically
           # Update/create each deployment file in src/services/kubernetes
-          for f in ./src/services/kubernetes/*.yaml; do
+          for f in $(find ./src/services/kubernetes -type f -name "*.yaml"); do
```
garyadaptive, line 54: This is where the magic happens. There is a job under services/kubernetes/initialise/initialise-job.yaml that is not currently picked up by the existing file glob. This will now traverse into this directory and pick up this file to be applied.

PR #1725, opened 2020-12-11
feat(new-client): tiles new features

slcfv3:


src/new-client/src/services/tiles.ts
```diff
@@ -89,3 +96,12 @@ export const tilesSubscription$ = concat(
       ),
   ),
 )
+
+const [selectedTileViewInput$, onSelectTileView] = createListener<TileView>()
```
iprado78, line 100: Ah, interesting, wasn't aware of this this utility yet.  @josepot - preference for using this rather than manually creating Subjects that we would pass into bind?

iprado78 (approved):

src/new-client/src/services/tiles.ts
josepot: I don't have a strong opinion on this. I _think_ that the code that `createListener` produces is a bit more declarative and sometimes even more concise, it also clearly separates the producer from the observable... On the other hand, though, I'm not sure if it's worth it to increate the API surface for something that trivial.

I was hopping that others would tell me whether they find it useful or not 🙂 .

PR #1727, opened 2020-12-11
refactor(new-client): rename 'blotter' to 'trades' and complete set f…

iprado78:
…ilter implementation  (656)


src/new-client/src/App/Trades/TradesGrid/SetFilter/SetFilter.tsx
```diff
@@ -25,31 +29,63 @@ export const SetFilter = forwardRef<unknown, IFilterParams>(
 
     const [filterSearchText, setFilterSearchText] = useState("")
     const [selectedFilters, setSelectedFilters] = useState(
-      () => new Set<string>(),
+      () => new Set<string>(filterOptions),
     )
     const inputRef = useRef<HTMLInputElement>(null)
 
+    const resetFilters = useCallback(
```
iprado78, line 36: Linter actually complains about removing because it's a dependency of an effect.  Ultimately, we can solve with @josepot recommendation to lift some of the behavior in the useEffects into streams.

slcfv3 (approved):

PR #1733, opened 2020-12-15
Refactor: upgrade and improve position bubble chart

jrothweiler:
This PR generally improves the position bubble chart code as a part of moving it over to the new client. The first step of this process was to upgrade our version of D3, the library powering the bubble physics simulation, to v6 (we were previously on v3). From there, I simplified a lot of the code by removing calculations that we were doing manually and using D3 to achieve the same. The types still aren't great in a few places, but I tried to fix the low hanging fruit in this regard.

### DIFF
--- a/src/new-client/package.json
+++ b/src/new-client/package.json
@@ -31,7 +31,7 @@
     "@types/react-dom": "^16.9.8",
     "ag-grid-community": "^24.1.0",
     "ag-grid-react": "^24.1.1",
-    "d3": "3.5.12",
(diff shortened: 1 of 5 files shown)


josepot (dismissed):
Amazing work @jrothweiler 🎉 ! Thank you so much!

iprado78 (dismissed):

josepot:
Sorry that I keep doing this, but I'm going to merge this before we get more conflicts.

PR #1734, opened 2020-12-16
refactor: change tiles to responsive(668)

slcfv3:
-Change the tiles components to responsive instead of adaptive.
-Add dynamic chart feature.


src/new-client/src/App/LiveRates/Tiles/TilePanel.tsx
```diff
@@ -0,0 +1,79 @@
...
+          className="spot-tile"
+          data-qa="analytics-tile__spot-tile"
+          data-qa-id={`currency-pair-${currencyPair.symbol.toLowerCase()}`}
+        >
+          <TileHeader ccyPair={currencyPair} date={date} />
+
+          <TileBodyWrapper isAnalyticsView={isAnalytics}>
+            {isAnalytics ? (
+              <LineChartWrapper isTimerOn={isTimerOn}>
+                <AnalyticsTileChart history={historicPrices} />
```
josepot, line 56: It would be better if the `AnalyticsTileChart` was using the `useHistoricalPrices` hook internally, for 2 reasons: 
- It's data that only the `AnalyticsTileChart` cares about, so better encapsulation.
- It wouldn't trigger unnecessary React updates to the other components.

src/new-client/src/App/LiveRates/Tiles/TilePanel.tsx
```diff
@@ -0,0 +1,79 @@
...
+              </LineChartWrapper>
+            ) : null}
+
+            <PriceControlWrapper>
+              <PriceControls
+                currencyPair={currencyPair}
+                priceData={priceData}
```
josepot, line 63: Same here, `priceData` it's only relevant to the `PriceControls` component.

src/new-client/src/App/LiveRates/Tiles/TilePanel.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+interface Props {
+  id: string
+}
+
+export const TilePanel: React.FC<Props> = memo(({ id }) => {
+  const currentView = useSelectedTileView()
+  const isAnalytics = currentView === TileView.Analytics
+  const currencyPairs = useCurrencyPairs()
+  const currencyPair = currencyPairs[id]
```
josepot, line 33: It would be better to pass the `currencyPair` through props... Either that or have a `useCurrencyPair` hook that accepts the `id` and returns the currencyPair of that id only when there are relevant changes to the given currenyPair.

josepot (approved):
Great work @slcfv3 🎉 ! Thanks a lot!

I'm going to merge this and afterwords I will open another PR addressing the comments that I've made in this PR and other small things.

Let's have a quick catch up when you are around and we can have a look at those changes together.

PR #1739, opened 2020-12-21
refactor: LiveRates refactor

josepot:
I'm sorry to send such a large refactor in one commit. Main differences:
- Better colocation of both state logic:
  - We no longer have a "tiles" service, we now have a "prices" service, instead
  - The logic for the `selectedCurrency` and the `selectedView` has now been moved into the `LivesRates` folder
  - Similarly, the logic for the "filtered symbols" has now been moved into the `Tiles` component.
  - Removed `utils` file and colocated those transformations into the only components that needed them.
  - Avoid prop-drilling by moving service related logic closer to the components that consume it.
 
- Refactored components so that they are better colocated/organized: less components, less "styled.tsx" files, etc.
- Better stream composition so that each component only exports the streams that they need, all the way up to `Subscribe`.


src/new-client/src/App/LiveRates/LiveRates.tsx
```diff
@@ -16,15 +20,25 @@ const LiveRateWrapper = styled(Wrapper)`
...
   height: 100%;
 `
 
+const liveRates$ = merge(tiles$, mainHeader$)
+
 export const LiveRates: React.FC = () => (
   <LiveRateWrapper>
     <OverflowScroll>
-      <MainPanel />
+      <div data-qa="workspace__tiles-workspace">
```
iprado78, line 33: Do you need three layers of wrapper? div/OverflowScrollLiveRateWrapper

slcfv3 (approved):

iprado78 (approved):

src/new-client/src/App/LiveRates/LiveRates.tsx
josepot: Of course not 🙂 .

However, the refactored code is (for now) supposed to produce the exact same DOM elements (and css styles) than we had before... (Unless there is a major reason not to).

FYI: once we have refactored all the services, feature-components, tests, etc We will look into properly leaveraging a better desing-system, but until then it's ok to keep producing the same DOM structure and css styles.

PR #1742, opened 2020-12-22
feat(new-client)/connection-status-component(705)

slcfv3:
Screenshot 2020-12-22 151536
-connection status components can display multiple services' connection status now

### DIFF
--- a/src/new-client/src/App/Footer/Footer.tsx
+++ b/src/new-client/src/App/Footer/Footer.tsx
@@ -1,6 +1,6 @@
 import ContactUsButton from "./ContactUsButton"
 import StatusBar from "./StatusBar"
-import { StatusButton } from "./StatusButton/StatusButton"
+import { StatusButton } from "./StatusButton"
 
 export const Footer: React.FC = () => (
   <StatusBar>
--- a/src/new-client/src/App/Footer/StatusButton/StatusButton.tsx
+++ b/src/new-client/src/App/Footer/StatusButton/StatusButton.tsx
@@ -1,7 +1,6 @@
 import { SyntheticEvent, useRef, useState, useEffect } from "react"
-import { of } from "rxjs"
 import { bind } from "@react-rxjs/core"
-import { map, timeoutWith } from "rxjs/operators"
+import { delay, map } from "rxjs/operators"
 import styled from "styled-components/macro"
 import { ServiceConnectionStatus, ServiceStatus } from "services/connection"
 import {
@@ -15,7 +14,8 @@ import {
 import { Root, Button } from "../common-styles"
 import Service from "./Service"
 import { usePopUpMenu } from "utils/usePopUpMenu"
-import { status$, ServiceInstanceStatus } from "services/status"
+import { ServiceInstanceStatus, status$ } from "services/status"
+import { of } from "rxjs"
 
 export const Wrapper = styled.div`
   color: ${(props) => props.theme.textColor};
@@ -85,54 +85,13 @@ const selectAll = (event: SyntheticEvent) => {
   input.select()
 }
 
-const getInstanceNumber = (
-  record: Record<string, ServiceInstanceStatus>,
-  type: string,
-) => {
-  return Object.values(record).filter(
-    (x) => x.isConnected && x.serviceType === type,
-  ).length
-}
-
-const timeoutThreshold = 3000
-const requestTimeoutLogger$ = of(
-  ["blotter", "reference", "execution", "pricing", "analytics"].map(
-    (serviceType) =>
-      ({
-        serviceType,
-        connectionStatus: ServiceConnectionStatus.DISCONNECTED,
-        connectedInstanceCount: 0,
-      } as ServiceStatus),
-  ),
-)
-
-export const [useServiceStatus, serviceStatus$] = bind(
-  status$.pipe(
-    map((record) => {
-      const nameList: Set<string> = new Set()
-      const result: ServiceStatus[] = []
-      const recordValues = Object.values(record)
-      recordValues.forEach((next) => {
-        const newStatus = {
-          serviceType: next.serviceType,
-          connectedInstanceCount: getInstanceNumber(record, next.serviceType),
-          connectionStatus: getInstanceNumber(record, next.serviceType)
-            ? ServiceConnectionStatus.CONNECTED
-            : ServiceConnectionStatus.DISCONNECTED,
-        }
-        if (nameList.size !== nameList.add(newStatus.serviceType).size) {
-          result.push(newStatus)
-        }
-      })
-      return result
-    }),
-    timeoutWith(timeoutThreshold, requestTimeoutLogger$),
-  ),
-  [] as ServiceStatus[],
+const [useServiceStatus, serviceStatus$] = bind<ServiceInstanceStatus[]>(
+  status$.pipe(map(Object.values)),
+  [],
 )
 
 export const [useApplicationStatus, applicationStatus$] = bind(
-  serviceStatus$.pipe(map((status) => getApplicationStatus(status))),
+  of(ServiceConnectionStatus.CONNECTED).pipe(delay(1500)),
   ServiceConnectionStatus.CONNECTING,
 )
 
@@ -140,7 +99,7 @@ export const StatusButton: React.FC = () => {
   const url = "https://web-demo.adaptivecluster.com"
   const ref = useRef<HTMLDivElement>(null)
   const { displayMenu, setDisplayMenu } = usePopUpMenu(ref)
-  const services: ServiceStatus[] = useServiceStatus()
+  const services = useServiceStatus()
 
   const appUrl = url
   const appStatus = useApplicationStatus()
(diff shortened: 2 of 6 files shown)


src/new-client/src/App/Footer/StatusButton/serviceInstanceCollection.ts
```diff
@@ -0,0 +1,72 @@
...
+    const x = this.serviceInstanceCollections.get(serviceType)
+
+    if (x) {
+      return x.getServiceNumberOfInstances()
+    }
+
+    return undefined
+  }
+
+  getStatusOfServices(): ServiceConnectionInfo {
```
iprado78, line 54: As it happens, I've spent a good chunk of time today looking at this code in the original client.   Think you can make this method a little simpler with fromEntries and our mapObject utility: 

```
  return mapObject(
     Object.fromEntries(this.serviceInstanceCollections),
    (instanceCollection) =>  ({
      serviceType: instanceCollection.serviceType,
      connectedInstanceCount: instanceCollection.getServiceNumberOfInstances(),
      connectionStatus: instanceCollection.getServiceNumberOfInstances()
        ? ServiceConnectionStatus.CONNECTED
        : ServiceConnectionStatus.DISCONNECTED,
   })
  )
```

The keys of the `this.serviceInstanceCollections` Map are the serviceTypes:

https://github.com/AdaptiveConsulting/ReactiveTraderCloud/pull/1742/files#diff-431d7047cc2d8c8daac07b0dca1679abc5e0d376488d73c194d6addc8299b500R104
)

src/new-client/src/App/Footer/StatusButton/serviceStatus.ts
```diff
@@ -0,0 +1,134 @@
...
+        addHeartBeatToServiceInstanceStatus(heartBeatTimeout),
+        scan<ServiceInstanceStatus, ServiceInstanceCollection>(
+          (serviceInstanceCollection, next) =>
+            serviceInstanceCollection.update(next),
+          new ServiceInstanceCollection(serviceInstanceStatus.key),
+        ),
+      ),
+    ),
+    scan<ServiceInstanceCollection, ServiceCollectionMap>(
+      (serviceCollectionMap, serviceInstanceCollection) => {
```
iprado78, line 103: See comment below on line 132

src/new-client/src/App/Footer/StatusButton/serviceStatus.ts
```diff
@@ -0,0 +1,134 @@
...
+  statusUpdate$: Observable<RawServiceStatus>,
+  heartBeatTimeout: number,
+) => {
+  return statusUpdate$.pipe(
+    map(convertFromRawMessage),
+    groupBy((serviceInstanceStatus) => serviceInstanceStatus.serviceType),
+    mergeMap((serviceInstanceStatus) =>
+      serviceInstanceStatus.pipe(
+        addHeartBeatToServiceInstanceStatus(heartBeatTimeout),
+        scan<ServiceInstanceStatus, ServiceInstanceCollection>(
```
iprado78, line 95: See comment below on line 132

src/new-client/src/App/Footer/Footer.tsx
```diff
@@ -1,10 +1,14 @@
...
 import { StatusButton } from "./StatusButton/StatusButton"
+import { serviceStatus$ } from "./StatusButton/serviceStatus"
+import { Subscribe } from "@react-rxjs/core"
 
 export const Footer: React.FC = () => (
   <StatusBar>
-    <ContactUsButton />
-    <StatusButton />
+    <Subscribe source$={serviceStatus$}>
+      <ContactUsButton />
```
josepot, line 10: Why is the `ContactUsButton` inside `Subscribe`?

src/new-client/src/App/Footer/Footer.tsx
```diff
@@ -1,10 +1,14 @@
 import ContactUsButton from "./ContactUsButton"
 import StatusBar from "./StatusBar"
 import { StatusButton } from "./StatusButton/StatusButton"
```
josepot, line 3: Remember to use index files to export the correct contracts of each module.

I was expecting to find something like: `import { StatusButton, statusButton$} from './StatusButton'`

src/new-client/src/App/Footer/StatusButton/StatusButton.tsx
```diff
@@ -85,7 +86,7 @@ export const StatusButton: React.FC = () => {
   const url = "https://web-demo.adaptivecluster.com"
   const ref = useRef<HTMLDivElement>(null)
   const { displayMenu, setDisplayMenu } = usePopUpMenu(ref)
-  const services: ServiceStatus[] = []
+  const services: ServiceStatus[] = useServiceStatus()
```
josepot, line 89: I would much rather to have a hook defined just before the `StatusButton` component like: `useApplicationStatus`.

src/new-client/src/App/Footer/StatusButton/StatusButton.tsx
```diff
@@ -85,7 +86,7 @@ export const StatusButton: React.FC = () => {
...
 
   const toggleMenu = useCallback(() => {
```
josepot, line 91: We shouldn't be using `useCallback` for things like this. Also, since the next value depends on the previous one: the toggle function should use the "function overload" of `setState`, like this: `() => { setDisplayMenu(prev => !prev) }` and I would much rather if that was defined directly in the `onClick` event-handle of `Button`.

josepot (changes requested):
Hi @slcfv3 

Please keep in mind that the goal is to adapt the previous functionality to the standards of the new architecture. In that sense, it's important that we get rid of those things that are no longer needed and to draw the correct boundaries between each module.

This is how I was envisioning this:
- Under services, let's add a new service named `status`, this service should expose a shared-observable of `Record<string, ServiceInstanceStatus>`, where the key of the `Record` is the ID of the `ServiceInstanceStatus`. Please, when creating this observable, let's make sure that we don't inherit all the bogus code from the old services 🙂 . Ping me if you want me to help you wit that, but in a nutshell the only thing that needs to happen is: use the `watch` function to get the status updates and then scan them into a `Record` of `ServiceInstanceStatus`, that's it.

- The status button is composed of 2 different components:
  - The button itself, which derives its state (connected, connecting, disconnected) from the `status$` stream that it's imported from services. However, on bind we should use the default state of `connecting`, so that we don't have to use `Subscribe`
  - The list of statusses which also derive their state from the `status$` stream that's imported from services. However, on `bind` we should use an empty list so that we don't have to use `Subscribe`

src/new-client/src/services/status.ts
```diff
@@ -0,0 +1,38 @@
...
+
+const convertFromRawMessage = (
+  serviceStatus: RawServiceStatus,
+): ServiceInstanceStatus => {
+  return {
+    serviceType: serviceStatus.Type,
+    serviceId: serviceStatus.Instance,
+    timestamp: serviceStatus.TimeStamp,
+    serviceLoad: serviceStatus.Load,
+    isConnected: true,
```
josepot, line 27: I don't see the point of adding the `isConnected` property, given the fact that its value will always be `true` looks misleading...

src/new-client/src/App/Footer/StatusButton/StatusButton.tsx
```diff
@@ -81,28 +85,81 @@ const selectAll = (event: SyntheticEvent) => {
...
+            ? ServiceConnectionStatus.CONNECTED
+            : ServiceConnectionStatus.DISCONNECTED,
+        }
+        if (nameList.size !== nameList.add(newStatus.serviceType).size) {
+          result.push(newStatus)
+        }
+      })
+      return result
+    }),
+    timeoutWith(timeoutThreshold, requestTimeoutLogger$),
```
josepot, line 129: I don't think that this is accurate, if the initial connection takes longer than 3 secs, that doesn't mean that we are disconnected from those services... It means that we have not been able to connect, we don't even know if those are the available services.

src/new-client/src/App/Footer/StatusButton/StatusButton.tsx
```diff
@@ -81,28 +85,81 @@ const selectAll = (event: SyntheticEvent) => {
...
+          connectionStatus: getInstanceNumber(record, next.serviceType)
+            ? ServiceConnectionStatus.CONNECTED
+            : ServiceConnectionStatus.DISCONNECTED,
+        }
+        if (nameList.size !== nameList.add(newStatus.serviceType).size) {
+          result.push(newStatus)
+        }
+      })
+      return result
+    }),
```
josepot, line 128: I'm pretty sure that the only transformation that's needed here is `map(Object.values)`

src/new-client/src/App/Footer/StatusButton/StatusButton.tsx
```diff
@@ -81,28 +85,81 @@ const selectAll = (event: SyntheticEvent) => {
...
+    timeoutWith(timeoutThreshold, requestTimeoutLogger$),
+  ),
+  [] as ServiceStatus[],
+)
+
+export const [useApplicationStatus, applicationStatus$] = bind(
+  serviceStatus$.pipe(map((status) => getApplicationStatus(status))),
```
josepot, line 135: By the looks of it, it seems that we are conflating the connection status with the health of the services, but those are 2 different things:

- The connection status has to do with the socket connection: connected, disconnected, connecting...
- The health of the services has to do with whether the services are healthy or not, which I suppose it depends on whether there are more than zero instances of each up and running.

We should clarify what the business requirement for this button is: whether it has to display the connection status or the health of the services, or both.

josepot (changes requested):

PR #1744, opened 2020-12-28
Fix/stomp client

iprado78:
Two changes to our stomp client: 

1.  Rolled back rx-stomp so that the automatic reconnects succeed.  Issue opened on the library here: https://github.com/stomp-js/rx-stomp/issues/277
2. Move `take(1)` inside our `getRemoteProcedureCall$` so that it always completes (the underlying rpc stream does complete, but we because we wrap in a `combineLatest` with two source streams that don't, our stream needs to be manually completed).

### DIFF
--- a/src/new-client/src/services/client/client.ts
+++ b/src/new-client/src/services/client/client.ts
@@ -1,5 +1,5 @@
 import { combineLatest, Observable } from "rxjs"
-import { map, switchMap } from "rxjs/operators"
+import { map, switchMap, take } from "rxjs/operators"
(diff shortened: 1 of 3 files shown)


josepot:
Before we merge this in, let's try to use the latest version of the library with the `connectionTimeout` option set to zero, and let's see if that does the trick.

iprado78:
> Before we merge this in, let's try to use the latest version of the library with the `connectionTimeout` option set to zero, and let's see if that does the trick.

It fixed the issue

josepot (approved):

PR #1746, opened 2020-12-29
chore(new-client): remove recharts

josepot:
What this PR accomplishes:
- It brings the total bundle size of the App down from 1.71MB to 1.38MB
- It fixes a couple of issues that I've found with the legends of the Profits & Losses historical chart
- We are now able to use css for changing the color of the line of the Tile, rather than using React state for the hover
- It gets rid of all those annoying warnings that we were getting in the console due to Recharts

What's missing:
- The tooltip that's displayed while hovering the Profits & Losses historical chart (coming soon)

PS: I can't believe that there are more deletions than additions on the diff PR after removing recharts. I'm pretty sure that if we didn't take into account the lines from the lock file that wouldn't be the case, though... but I still think that's hilarious 😂 .


src/new-client/src/App/Analytics/ProfitAndLoss/LineChart/dataPoints$.ts
```diff
@@ -0,0 +1,17 @@
...
+
+export const dataPoints$ = history$.pipe(
+  map((historical) => {
+    const points = historical.map(
+      (entry) => [new Date(entry.timestamp), entry.usPnl] as [Date, number],
+    )
+    const xRange = [points[0][0], points[points.length - 1][0]] as const
+
+    const yValues = points.map(([, y]) => y)
+    const yRange = [Math.max(...yValues), Math.min(...yValues)] as const
```
iprado78, line 13: Could use d3.extent too

```
    const xRange = extent(points.map(([x]) => x)) as [Date, Date]
    const yRange = extent(points.map(([_, y]) => y)).reverse() as [number, number]
```

src/new-client/src/App/LiveRates/Tiles/AnalyticsTile.tsx
```diff
@@ -1,13 +1,32 @@
...
+const VIEW_BOX_WIDTH = 200
+const VIEW_BOX_HEIGHT = 81
+
+export const xScale = scaleTime().range([0, VIEW_BOX_WIDTH])
+export const yScale = scaleLinear().range([0, VIEW_BOX_HEIGHT])
+
+const [useHistoricalPath, analyticsTile$] = bind((symbol: string) =>
+  getHistoricalPrices$(symbol).pipe(
+    map((prices) => {
+      const points = prices.map(
```
iprado78, line 18: Merit an abstraction?

iprado78 (approved):

src/new-client/src/App/Analytics/ProfitAndLoss/LineChart/dataPoints$.ts
josepot: Nice! Thanks!

src/new-client/src/App/LiveRates/Tiles/AnalyticsTile.tsx
josepot: I was on the fence on this one... I've added a new commit that abstracts the main transformations of the "historical chart", in a way that they can be easily composed. I think that's worth it, let me know what you think about it. Thanks!

PR #1747, opened 2020-12-30
chore(server): add connection timeout to node bot

chuck-adaptive:
Adds `connectionTimeout` to the existing bot config

Also adds `depends_on` to bot (depending on the broker)

### DIFF
--- a/src/docker-compose.yml
+++ b/src/docker-compose.yml
@@ -103,6 +103,8 @@ services:
     image: ${DOCKER_USER}/bot:${BUILD_VERSION}
     build:
       context: ./server/node/bot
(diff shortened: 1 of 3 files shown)


garyadaptive:
Looks good - I think this should work well.

The issue I mentioned in standup relates to this: https://stackoverflow.com/questions/31746182/docker-compose-wait-for-container-x-before-starting-y/41854997#41854997. But we achieve essentially the same effect with your solution, as we're using the built-in reconnection behaviour of RxStomp and it has the advantage of working in k8s as well.

garyadaptive (approved):

PR #1748, opened 2021-01-04
chore(new-client): remove ag-grid dependency (707)

iprado78:
This PR is the first of what will be several related to removing our ag-grid dependency (and shrinking the bundle size drastically as a result).   The scope of this PR simply renders our trades to an html data table (with most of the existing styles).   ToDo PRs:

1.  Sorting by column
2. Quick text / universal filter
3. Column filtering (will need UX)
4. Export to excel
5. Responsiveness (we should settle on approach)
6. Virtualization


src/new-client/src/App/Trades/TradesHeader/AppliedFilters.tsx
```diff
@@ -36,20 +32,8 @@ const FilterIcon = styled("i")`
...
-      ),
-    ),
-  ),
-  [],
-)
-
 export const AppliedFilters: React.FC = () => {
-  const api = useGridApi()
-  const filteredColDefs = useFilteredColDefs()
+  const filteredColDefs: any[] = []
```
slcfv3, line 36: Can we avoid any type here?

src/new-client/src/App/Trades/TradesHeader/QuickFilter.tsx
```diff
@@ -55,11 +54,10 @@ const QuickFilterClearIcon = styled("i")`
 export const QuickFilter: React.FC = () => {
   const quickFilterInput = useRef<HTMLInputElement>(null)
   const [quickFilterText, setQuickFilterText] = useState<string>("")
-  const api = useGridApi()
 
   useEffect(() => {
-    api.setQuickFilter(quickFilterText.length ? quickFilterText : null)
-  }, [quickFilterText, api])
+    console.log("ToDo")
```
slcfv3, line 59: Why do we need to console log "todo" in multiple places?

PR #1750, opened 2021-01-05
feat(new-client)/connection-status-component(705)

slcfv3:
-connection status components can display multiple services' connection status now
-service health and application connection status are independent: service health only depends on how many nodes are loaded, while application connection depends on the status of socket
-when connection is lost, status bar will show disconnected and when connection is back, app will reconnect and connection bar will show 'connecting --- connected'

### DIFF
--- a/src/new-client/src/services/status.ts
+++ b/src/new-client/src/services/status.ts
@@ -1,0 +1,61 @@
+import { watch$ } from "services/client"
+import { delay, map, mergeAll, scan, switchMap, take } from "rxjs/operators"
+import { split } from "@react-rxjs/utils"
+import { merge, of } from "rxjs"
+import { bind, shareLatest } from "@react-rxjs/core"
+
+export interface RawServiceStatus {
+  Type: string
+  Instance: string
+  TimeStamp: number
+  Load: number
+}
+
+export interface ServiceInstanceStatus {
+  serviceType: string
+  serviceId: string
+  timestamp: number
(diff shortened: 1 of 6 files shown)


garyadaptive (approved):

src/new-client/src/services/status.ts
```diff
@@ -0,0 +1,61 @@
+import { watch$ } from "services/client"
+import { delay, map, mergeAll, scan, switchMap, take } from "rxjs/operators"
+import { split } from "@react-rxjs/utils"
+import { merge, of } from "rxjs"
+import { bind, shareLatest } from "@react-rxjs/core"
+
+export interface RawServiceStatus {
```
josepot, line 7: "raw" interfaces shouldn't be exported.

src/new-client/src/services/status.ts
```diff
@@ -0,0 +1,61 @@
...
+        }
+      } else {
+        result[item.Type] = {
+          ...result[item.Type],
+          serviceLoad: result[item.Type].serviceLoad - 1,
+        }
+      }
+      return result
+    }, {} as Record<string, ServiceInstanceStatus>),
+    shareLatest(),
```
josepot, line 59: `bind` already applies the `shareLatest` enhancer, so this one is redundant.

slcfv3: No problem, I will delete these two in my next PR

PR #1751, opened 2021-01-05
Feature/new grid

iprado78:
This PR is the first of what will be several related to removing our ag-grid dependency (and shrinking the bundle size drastically as a result). The scope of this PR:

1.  Renders our trades to an html data table (with most of the existing styles)
2. Adds sorting by column to the table

ToDo:
1. Quick text / universal filter
2. Export to excel
3. Column filtering (will need UX)
4. Responsiveness (we should settle on approach)
5. Virtualization


src/new-client/src/App/Trades/services.ts
```diff
@@ -0,0 +1,113 @@
...
+import { map, scan, startWith } from "rxjs/operators"
+import { trades$ } from "services/trades"
+import { ColField } from "./TradesGrid"
+
+type SortDirection = "ASC" | "DESC"
+
+class TableSort {
+  constructor(public direction?: SortDirection, public field?: ColField) {}
+}
+
```
slcfv3, line 12: I am not quite familiar with class, just wondering if it's possible to create a new interface for it instead of a class.

slcfv3 (approved):

src/new-client/src/App/Trades/services.ts
iprado78: The benefit of the class over the interface is that it automatically types the objects it creates so you don't have to do typecasting in a lot of situations where you would with an object literal.  I started with an interface but then I was writing `{ direction: ..., field: ...} as TableSort` everywhere.

PR #1755, opened 2021-01-07
feat(new-client): export excel button(713)

slcfv3:
-Export the trades new table to csv file

### DIFF
--- a/src/new-client/src/App/Trades/TradesHeader/ExcelButton.tsx
+++ b/src/new-client/src/App/Trades/TradesHeader/ExcelButton.tsx
@@ -1,4 +1,7 @@
+import { map, take } from "rxjs/operators"
 import styled from "styled-components/macro"
+import { tableTrades$ } from "../services"
+import { colConfigs } from "../TradesGrid/TradesGrid"
 
 const ExcelIcon = () => (
   <svg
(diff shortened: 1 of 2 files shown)


src/new-client/src/App/Trades/TradesHeader/ExcelButton.tsx
```diff
@@ -40,10 +44,47 @@ const Button = styled("button")`
...
+          return res
+        }),
+      ),
+    ),
+  ),
+  [],
+)
 
 export const ExcelButton: React.FC = () => {
+  const trades = useExportTable()
```
iprado78, line 65: Because this component doesn't render any of this trade data to the UI, we shouldn't actually need this hook, which brings the trades into the component state.  In fact, the whole `download_csv` function (which I would rename to  `downloadCsv` for consistency) should be moveable up and out of the component, since it doesn't use any component state.   So the idea would be:  

1. No need for `bind`, just define `const exportTable$ = tableTrades$.pipe(...)` directly
2. Add the `take(1)` after your `map`
3. Inside `downloadCsv`, do 
```
exportTable$.subscribe((trades) => {
.... your csv code
})
```

Because it's a `take(1)` and all of this happens outside the component, the Observable won't hang around in memory  between clicks.

iprado78 (approved):

PR #1756, opened 2021-01-07
feat(new-client): virtualize trades(721)

slcfv3:
-visualize the trades with react-window so that our table can do lazy loading and the performance is improved.


src/new-client/src/App/Trades/TradesGrid/TradesGrid.tsx
```diff
@@ -8,6 +8,7 @@ import {
 import { capitalize } from "utils/capitalize"
 import { format as formatDate } from "date-fns"
 import { useTableTrades, sortFieldSelections$, useTableSort } from "../services"
+import { FixedSizeList as List } from "react-window"
```
iprado78, line 11: In our case, I would import as "VirtualizedTable" so that it reads semantically for what it is doing

src/new-client/src/App/Trades/TradesGrid/TradesGrid.tsx
```diff
@@ -145,48 +146,70 @@ const DATE_FORMAT = "dd-MMM-yyyy"
...
+      </TableHead>
+      {children}
+    </Table>
+  )
+}
+
+export const TradesGrid: React.FC = () => {
+  const trades = useTableTrades()
+
+  const Row = ({ index, style }: RowParam) => (
```
iprado78, line 189: Since this doesn't depend directly on any state from the TradesGrid, you can/should declare it outside of the TradesGrid component.  That way the Row component is created once, rather than on every render.  Also, following the naming pattern, call it "VirtualizedTableRow"

src/new-client/src/App/Trades/TradesGrid/TradesGrid.tsx
```diff
@@ -145,48 +146,70 @@ const DATE_FORMAT = "dd-MMM-yyyy"
...
+    </TableBodyRow>
+  )
+
+  return (
+    <TableWrapper>
+      <List
+        className="List"
+        height={150}
+        itemCount={trades.length}
+        itemSize={30}
```
iprado78, line 206: Do we know if it can accept units other than pixels??--specifically, `rem` values?

iprado78 (commented):
Looks good.  Have mostly minor comments.  I am going to pull down the code and run it just look at how it is behaving UX/UI wise.  Also, FYI, in terms of the terminology on the PR message, it's "virtualize" not "visualize".

src/new-client/src/App/Trades/TradesGrid/TradesGrid.tsx
slcfv3: I tried to do it yesterday but how can I help the Row component to consume the Rxjs observable outside the components? Should I call the useTableTrades in Row component?

src/new-client/src/App/Trades/TradesGrid/TradesGrid.tsx
slcfv3: I don't think it's doable since height attribute only accepts number type.

src/new-client/src/App/Trades/TradesGrid/TradesGrid.tsx
iprado78: From what I read in the docs just how, I believe you pass the trades array into the List/VirtualizedTable via the `itemData` prop.  It's then available to your row via the 'data' prop.  So you should be able to access each trade inside the row component as  `data[index]`.  Because it's typed as an `any` you'll like have to cast it as `Trade`.

src/new-client/src/App/Trades/TradesGrid/TradesGrid.tsx
```diff
@@ -145,47 +183,79 @@ const DATE_FORMAT = "dd-MMM-yyyy"
...
+                )
+              ) : (
+                <span className="spacer" />
+              )}
+            </TableHeadCell>
           ))}
-        </tbody>
+        </TableHeadRow>
+        <VirtualizedTable
+          className="List"
```
iprado78, line 250: What's using this className?

src/new-client/src/App/Trades/TradesGrid/TradesGrid.tsx
```diff
@@ -145,47 +183,79 @@ const DATE_FORMAT = "dd-MMM-yyyy"
...
+interface RowParam {
+  data: Trade[]
+  index: number
+  style: React.CSSProperties
+}
+
+const VirtualizedTableRow = ({ data, index, style }: RowParam) => (
+  <TableBodyRow key={data[index].tradeId} style={style}>
+    <StatusIndicator status={data[index].status}>
+      0
```
iprado78, line 195: Why "0"?

src/new-client/src/App/Trades/TradesGrid/TradesGrid.tsx
```diff
@@ -8,98 +9,114 @@ import {
...
   border-spacing: 0;
 `
-const TableHead = styled.thead`
-  font-size: 0.675rem;
-  text-transform: uppercase;
-`
-const TableHeadRow = styled.tr`
+
+const TableHeadRow = styled.div`
   vertical-align: center;
```
iprado78, line 102: This is dead code now that this is not a table.  I would check which of the other styles are no longer relevant.

src/new-client/src/App/Trades/TradesGrid/TradesGrid.tsx
```diff
@@ -145,47 +183,79 @@ const DATE_FORMAT = "dd-MMM-yyyy"
...
+  return rem * parseFloat(getComputedStyle(document.documentElement).fontSize)
+}
+
 export const TradesGrid: React.FC = () => {
   const trades = useTableTrades()
   const tableSort = useTableSort()
+  const [tableHeight, setTableHeight] = useState(55)
+  const containerPercent = useSelectedHeight()
+
+  useEffect(() => {
```
iprado78, line 216: If you have to do something like this, I would use  useLayoutEffect.  For the reasons described here: 

https://daveceddia.com/useeffect-vs-uselayouteffect/

In general, you are doing all of this to get one value, the dynamic table height.  What I would I do, in that case, is define one custom hook, `useTableHeight`, composed of Observable streams that emits the right table-height.  Also, where are the number values that you are passing into your functions--`0.01`, `3.5`, `5`, `2`, etc.--coming from?

At the end of the day, though, the question is do we need to do all this work?  What were the approaches css approaches you tried before dynamically calculating the height with JS?

iprado78 (changes requested):
On the UX side:

1. If you shrink the table horizontally, you get overlap on the first column
2. Table headers for numeric columns no longer aligned (problem in master, but it's looking worse now)
3. Space between table footer and table too big (again, problem in master, but worse on this branch)

src/new-client/src/App/Trades/TradesGrid/TradesGrid.tsx
slcfv3: Yeah as I said, this version works but not that elegant, since those values like 0.01, 3.5  are from rem values of other element. I do want to use CSS way to use all the available space in that area, but the react-window library needs to use absolute styling. Still thinking about a better way, can we try to use some other library to virtualize the grid?

PR #1757, opened 2021-01-14
feat(new-client): refine the stylings(722)

slcfv3:
-refined the stylings of new-client to match old client's stylings

### DIFF
--- a/src/new-client/public/index.html
+++ b/src/new-client/public/index.html
@@ -7,8 +7,6 @@
       content="width=device-width, initial-scale=1, shrink-to-fit=no, user-scalable=no"
     />
 
-
-
     <!-- PWA -->
     <meta id="themeColor" name="theme-color" content="#2e3543" />
     <meta name="apple-mobile-web-app-capable" content="yes" />
--- a/src/new-client/src/App/Analytics/styled.tsx
+++ b/src/new-client/src/App/Analytics/styled.tsx
@@ -18,7 +18,7 @@ export const AnalyticsInnerWrapper = styled.div<{ inExternalWindow?: boolean }>`
   width: 100%;
   height: 100%;
   max-width: 60rem;
-  min-width: 20rem;
+  min-width: 22rem;
   margin: auto;
   overflow-y: auto;
   overflow-x: hidden;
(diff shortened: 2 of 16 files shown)


src/new-client/public/index.html
```diff
@@ -7,7 +7,7 @@
       content="width=device-width, initial-scale=1, shrink-to-fit=no, user-scalable=no"
     />
 
-
+    <link href="%PUBLIC_URL%/fontawesome/css/all.min.css" rel="stylesheet" />
```
iprado78, line 10: Font-awesome icons via css were removed so that we don't ship all of these webfonts to the browser (which will include all the icons we don't need).  We should be able to get all the icons we need from the React icons library that was added.  See this PR: https://github.com/AdaptiveConsulting/ReactiveTraderCloud/pull/1752

iprado78 (changes requested):
TodDos:

1. Pull in the Executions commit.   There are some differences in styling (biggest is the font color of the confirmation messages on the light theme).
2. Responsiveness of the filters on the Live Rates menu: checkout the behavior of https://web-demo.adaptivecluster.com/ at 915 px width.  Our filters start to wrap then disappear.  Existing app transforms to drop-down.
3. On load, our tiles show the hover-state styles over the prices and then fade out.  Was this intentional?  It's not in the existing app.
4.  On the light theme, you don't see the light gray background over the prices when the mouse enters the tile.
5. On the dark theme, the contact us email address is the wrong color
6. In the connections panel, the version number is missing from the bottom left
7. On the profit and loss chart, font-size of the text should be 10 not 12px

slcfv3:
> TodDos:
> 
> 1. Pull in the Executions commit.   There are some differences in styling (biggest is the font color of the confirmation messages on the light theme).
> 2. Responsiveness of the filters on the Live Rates menu: checkout the behavior of https://web-demo.adaptivecluster.com/ at 915 px width.  Our filters start to wrap then disappear.  Existing app transforms to drop-down.
> 3. On load, our tiles show the hover-state styles over the prices and then fade out.  Was this intentional?  It's not in the existing app.
> 4. On the light theme, you don't see the light gray background over the prices when the mouse enters the tile.
> 5. On the dark theme, the contact us email address is the wrong color
> 6. In the connections panel, the version number is missing from the bottom left
> 7. On the profit and loss chart, font-size of the text should be 10 not 12px

Point 6 doesn't need to be fixed, as long as we put a version code in .env, it will be there. Others are all fixed.

iprado78 (approved):

PR #1758, opened 2021-01-14
feat(new-client): filter on base currency now(723)

slcfv3:


src/new-client/src/services/currencyPairs.ts
```diff
@@ -82,7 +82,7 @@ export const [useCurrencies, currencies$] = bind(
     map((currencyPairs) => [
       ...new Set(
         Object.values(currencyPairs)
-          .map((currencyPair) => [currencyPair.base, currencyPair.terms])
+          .map((currencyPair) => [currencyPair.base])
           .flat(),
       ),
```
iprado78, line 87: Did you test?  This changes the filter options displayed in the header, but not the filtering of tiles (see for example filtering on USD, you'll get tiles where it's in the `terms` part of the pair).  You'll need to make changes to `useFilteredSymbols`.  Also, you longer need to wrap in array and flatten.

iprado78 (changes requested):

src/new-client/src/services/currencyPairs.ts
slcfv3: Yeah, I know, I double checked with Gary and observed the logic of old client, our filter will return both 'AUD/USD' and 'EUR/AUD' when we choose AUD. Do you want to return 'AUD/USD' only?

slcfv3: Changed the filtering logic but still need to double check with Josep before we merge it in.

iprado78 (approved):

PR #1759, opened 2021-01-18
chore(new-client): deal with execution timeout situation

slcfv3:
EUR/JPY request is intentionally set a 5 seconds delay in backend, in order to demo timeout situation. Now new client can deal with execution time out situations as the previous client.

### DIFF
--- a/src/new-client/src/App/LiveRates/Tiles/ExecutionStatusAlertSwitch/Response.tsx
+++ b/src/new-client/src/App/LiveRates/Tiles/ExecutionStatusAlertSwitch/Response.tsx
@@ -123,4 +123,34 @@ const Rejected = () => (
   </ResponseContainer>
 )
 
-export { Done, Rejected }
+const TakingTooLong = () => {
+  const symbol = useContext(SymbolContext)
+  const baseTerm = useBaseTerm(symbol)
+  return (
+    <ExecutionStatusAlertContainer status={ExecutionStatus.TakingTooLong}>
+      <CurrencyPairDiv>
+        <AlignedTriangle />
+        {baseTerm}
+      </CurrencyPairDiv>
+      <TradeMessageDiv>
+        Trade execution taking longer than expected
+      </TradeMessageDiv>
+    </ExecutionStatusAlertContainer>
+  )
+}
+
+const RequestTimeout = () => {
+  const symbol = useContext(SymbolContext)
+  const baseTerm = useBaseTerm(symbol)
+  return (
+    <ExecutionStatusAlertContainer status={ExecutionStatus.RequestTimeout}>
+      <CurrencyPairDiv>
+        <AlignedTriangle />
+        {baseTerm}
+      </CurrencyPairDiv>
+      <TradeMessageDiv>Trade execution timeout exceeded</TradeMessageDiv>
+    </ExecutionStatusAlertContainer>
+  )
+}
+
+export { Done, Rejected, TakingTooLong, RequestTimeout }
--- a/src/new-client/src/services/executions.ts
+++ b/src/new-client/src/services/executions.ts
@@ -1,6 +1,6 @@
 import { bind } from "@react-rxjs/core"
 import { collect, split } from "@react-rxjs/utils"
-import { concat, race, Subject, timer } from "rxjs"
+import { concat, race, Subject, timer, OperatorFunction } from "rxjs"
 import {
   exhaustMap,
   filter,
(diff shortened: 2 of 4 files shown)


src/new-client/src/services/executions.ts
```diff
@@ -44,6 +46,8 @@ export enum ExecutionStatus {
   Ready = "Ready",
   Pending = "Pending",
   Done = "Done",
+  ClientTimeout = "ClientTimeout",
```
josepot, line 49: I find a bit difficult to understand what `ClientTimeout` means. I would rather a more descriptive name like: `TakingTooLong`.

src/new-client/src/services/executions.ts
```diff
@@ -143,6 +147,8 @@ export const onExecutionDismiss = (i: MinimalExecution) =>
   executionDismiss$.next(i)
 
 const DISMISS_TIMEOUT = 5000
+const REQEUST_TIMEOUT = 30000
```
josepot, line 150: `REQUEST`

src/new-client/src/services/executions.ts
```diff
@@ -143,6 +147,8 @@ export const onExecutionDismiss = (i: MinimalExecution) =>
...
+const CLIENT_TIMEOUT = 2000
```
josepot, line 151: `TAKING_TOO_LONG`  or something like that

src/new-client/src/services/executions.ts
```diff
@@ -154,6 +160,26 @@ const executionsMap$ = newExecution$.pipe(
...
+                  ),
+                  timer(REQEUST_TIMEOUT).pipe(
+                    mapTo({
+                      ...execution,
+                      status: ExecutionStatus.RequestTimeout,
+                    }),
+                    takeUntil(request),
+                  ),
+                ),
+              ),
```
josepot, line 182: I find this a bit difficult to follow/understand. The fact that the timer for the `REQUEST_TIMEOUT` is defined twice is a bit confusing IMO.

What I suggest is that:

- We improve the `execute` function so that:
  - it takes the `Execution` as its parameter
  - it also maps the response
  - it races that ^ stream with a timer that will map to a RequestTimeout execution

That way the stream returned by `execute` will handle all the logic/side-effects related with the execution itself.

- Then, since it's likely that we will want to apply the behavior of emitting a value if the original stream hasn't emitted in a certain time, let's create a piepable operator named `withTooLong` message that has the following signature:
```ts
const withTooLongMessage: <M, T>(ms: number, message: M) => OperatorFunction<T, M | T>
```
For that you can use the technique described in this [this blog-post](https://ncjamieson.com/calling-publish-with-a-selector/#with-a-selector)

- After having done that, then we just need to put it together:
```ts
exhaustMap((execution) =>                              
  concat(                                              
    execute(execution).pipe(                           
      withTooLongMessage(TAKING_TOO_LONG, {             
        ...execution,                                  
        status: ExecutionStatus.ClientTimeout,         
      }),                                              
    ),                                                 
    race([                                             
      executionDismiss$.pipe(                          
        filter((x) => x.currencyPair === currencyPair),
        take(1),                                       
      ),                                               
      timer(DISMISS_TIMEOUT),                          
    ]).pipe(mapTo(blankExecution(currencyPair))),      
  ),                                                   
),                                                     
```

src/new-client/src/App/LiveRates/Tiles/ExecutionStatusAlertSwitch/Response.tsx
```diff
@@ -123,4 +123,34 @@ const Rejected = () => (
...
+      <CurrencyPairDiv>
+        <AlignedTriangle />
+        {baseTerm}
+      </CurrencyPairDiv>
+      <TradeMessageDiv>Trade execution timeout exceeded</TradeMessageDiv>
+    </ExecutionStatusAlertContainer>
+  )
+}
+
+export { Done, Rejected, ClientTimeout, RequestTimeout }
```
josepot, line 156: I'm not a big fan of having a component per status. However, since this is something that wasn't introduced by this PR, let's address it in a separate one.

josepot (changes requested):
Good work in here @slcfv3 ! I like to see that you are getting more and more familiar with RxJS, nice usage of `publish`! :clap: Now, let's just make it a bit more declarative :wink:

src/new-client/src/App/LiveRates/Tiles/ExecutionStatusAlertSwitch/Response.tsx
slcfv3: No problem, I will modify the things above and save this one for next PR, thanks for your time Josep!

josepot (approved):
Good stuff!! Thanks @slcfv3 !

PR #1764, opened 2021-01-19
WIP: service tests

iprado78:
I've put in comments the various problems with the setup.  Basically, there's no isolation.  There's really one trades$ subscription maintained through all the tests.  At that point, I might as well just directly test $trades rather than the hook.


josepot:
I'm closing this, because this was finished in other PRs, but luckily we have been able to leverage a good amount of the work that @iprado78 did here. So, thanks a lot for this!

PR #1770, opened 2021-01-22
chore(new-client): remove notional hard code(728)

slcfv3:
-now users can customize the notional value to send instead of hard code.


josepot:
Also closing this in favor of #1774

PR #1774, opened 2021-01-26
refactor(new-client): improve tiles and execution

josepot:


src/new-client/src/App/LiveRates/Tile/state.ts
```diff
@@ -0,0 +1,138 @@
...
+            ),
+            emitTooLongMessage(TAKING_TOO_LONG, {
+              status: TileStates.TooLong,
+            }),
+            startWith({ status: TileStates.Started }),
+          ),
+          race([
+            dismiss$.pipe(
+              filter((s) => s === symbol),
+              take(1),
```
slcfv3, line 130: Still kind of confused why take(1) fixed the bug, it seems even the dismiss$ doesn't complete, it will still return a status: ready, right? Why it will make our buttons deactivated?

slcfv3 (approved):

src/new-client/src/App/LiveRates/Tile/state.ts
josepot: The reason is that:

- `dismiss$.pipe(filter((s) => s === symbol))` returns an observable that doesn't complete, because `dismiss$` doesn't complete.
- On the other hand, the observable `timer(DISMISS_TIMEOUT)` emits once and completes immediately.
- Which means that the observable returned by `race` only completes if the `timer` wins the race. Because if the first observable wins, then that one doesn't complete.
- It is important that the observable returned by `race` always completes, because otherwise the inner-observable of `exhaustMap` won't complete. Keep in mind that `exhaustMap` is like the "opposite" of `switchMap`. In the sense that what `exhaustMap` does is to ignore the emissions of its source until its inner-observable hasn't completed.

There were other ways to accomplish the same. For instance:
- `dismiss$.pipe(takeUntil((s) => s === symbol))
- or
```ts
race([
  dismiss$.pipe(filter((s) => s === symbol)),
  timer(DISMISS_TIMEOUT),
]).pipe(take(1))
```
In fact, that that I wrote that, it's quite obvious to me that using `takeUntil` is a lot more declarative for this... So I will change that :slightly_smiling_face: 

Please @slcfv3 ask more questions if I didn't explain this properly or if this explanations leaves your with more doubts than answers. I do realize that this is a lot to digest.

src/new-client/src/App/LiveRates/Tiles.tsx
```diff
@@ -22,21 +22,24 @@ const PanelItems = styled.div`
...
-        : allSymbols.filter((symbol) =>
-            symbol.includes(selectedCurrency as string),
-          )
+      if (selectedCurrency === ALL_CURRENCIES)
+        return Object.values(currencyPairs)
+      const result = { ...currencyPairs }
+      Object.keys(currencyPairs)
+        .filter((symbol) => symbol.includes(selectedCurrency as string))
+        .forEach((symbol) => {
+          delete result[symbol]
```
slcfv3, line 34: it makes sense to separate the All_currencies condition and other condition, but why we have a delete here?

slcfv3 (dismissed):

src/new-client/src/App/LiveRates/Tiles.tsx
josepot: Ups! I was deleting the wrong ones. Thanks for pointing that out!

PR #1775, opened 2021-01-26
feat(new-client): format the new notional input(730)

slcfv3:
-format the notional input in new tile structure


src/new-client/src/App/LiveRates/Tile/Notional.tsx
```diff
@@ -32,27 +30,28 @@ export const CurrencyPairSymbol = styled("span")`
...
-const notional = 1_000_000
-
-export { notionalInput$ }
 export const NotionalInput: React.FC = () => {
-  const symbol = useContext(SymbolContext)
-  const base = useBase(symbol)
+  const { base, symbol } = useTileCurrencyPair()
+  const notional = useNotional(symbol)
+  const formattedNotional = formatAsWholeNumber(parseInt(notional))
+  const [isFocused, setIsFocused] = useState(false)
```
josepot, line 37: Nitpick:

```suggestion
  const [isFocused, setIsFocused] = useState(false)
  const value = isFocused
    ? notional
    : formatAsWholeNumber(parseInt(notional))
```

josepot (approved):
LGTM! I made a small suggestion, so that we don't compute the formatting while the user is updating the value.

PR #1779, opened 2021-01-29
feat(new-client): implement parametric streams for column filters (731)

iprado78:


src/new-client/src/App/Trades/TradesState/filterState.ts
```diff
@@ -21,35 +22,75 @@ export const fieldValuesContainer = Object.freeze(
...
-      return {
-        ...appliedFilters,
-        [field]: filterSet,
-      }
-    }, ClonedFieldValuesContainer()),
-    startWith(ClonedFieldValuesContainer()),
-  ),
+abstract class FilterEvent {
+  constructor(public colField: ColField) {}
+}
```
slcfv3, line 46: Why we need to create abstract class here? To accept parameters?

iprado78: We don't need to, but I think it helps define the interface.  Every column filter event is going to need a column field and something else (in the case of the reset event, you don't need another piece of data, but you need to know you are resetting).  So this lets us organize all the filter events as extending this class (and forex define a function that takes this type as input), but the class itself will never be directly instantiated.

In terms of why use classes at all... I realize we don't do much OOP in this project.  But, with TypeScript IME, it makes it easier to define a big switch working with _instanceof_ .

iprado78:
This PR now does three things:

1. Refactors the column filters to use parametric streams
2. Adds a text search to narrow the set filter options
3. Adds a Select All option

josepot (approved):

PR #1781, opened 2021-02-02
feat(new-client): add number filter to trades(736)

slcfv3:
-align the number columns to right, as old client
-add filters to number column like notional
-fix a small issue of set filter

### DIFF
--- a/src/new-client/src/App/Trades/TradesGrid/NumFilter.tsx
+++ b/src/new-client/src/App/Trades/TradesGrid/NumFilter.tsx
@@ -3,9 +3,9 @@ import { usePopUpMenu } from "utils/usePopUpMenu"
 import styled from "styled-components/macro"
 import {
   NumColField,
-  comparatorConfigs,
   ComparatorType,
   onColFilterEnterNum,
+  useAppliedNumFilters,
 } from "../TradesState"
 
 export const FilterPopupWrapper = styled.span`
@@ -62,11 +62,9 @@ export const FilterOptionMenuWrapper = styled.div`
   }
 `
 
-export const FilterOptionMenu = styled.div<{
-  visible: boolean
-}>`
-  display: ${({ visible }) => (visible ? "block" : "none")};
+export const FilterOptionMenu = styled.div<{ visible: boolean }>`
   border-radius: 4px;
+  display: ${({ visible }) => (visible ? "block" : "none")};
   position: absolute;
   top: 40px;
   left: 0px;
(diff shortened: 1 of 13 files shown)


src/new-client/src/App/Trades/TradesGrid/NumFilter.tsx
```diff
@@ -0,0 +1,230 @@
...
+import { usePopUpMenu } from "utils/usePopUpMenu"
+import styled from "styled-components/macro"
+import {
+  ColField,
+  comparatorConfigs,
+  ComparatorType,
+  onColFilterEnterNum,
+} from "../TradesState"
+
+export const MultiSelectWrapper = styled.span`
```
iprado78, line 11: In this case, it's not a MultiSelect.  This is heading in the direction of becoming a generic styling component that we'll use for all three filters (number/date/set).  Don't worry about making it generic.  I could try that in my PR for the date filter.  But change the name to something like "FilterPopupWrapper".

src/new-client/src/App/Trades/TradesGrid/NumFilter.tsx
```diff
@@ -0,0 +1,230 @@
...
+  padding: 8px 15px 5px 15px;
+  cursor: pointer;
+  display: inline-block;
+
+  @media (max-width: 400px) {
+    display: none;
+  }
+`
+
+export const MultiSelectMenu = styled.div`
```
iprado78, line 28: "FilterPopup"

src/new-client/src/App/Trades/TradesGrid/NumFilter.tsx
```diff
@@ -0,0 +1,230 @@
...
+  max-height: 8rem;
+  overflow-y: auto;
+  top: 0px;
+  right: 50%;
+  background-color: ${({ theme }) => theme.primary.base};
+  padding: 6px;
+  box-shadow: ${({ theme }) => theme.core.textColor} 0px 0px 0.3125rem 0px;
+`
+
+export const MultiSelectOption = styled.div<{
```
iprado78, line 41: Unused

src/new-client/src/App/Trades/TradesGrid/NumFilter.tsx
```diff
@@ -0,0 +1,230 @@
...
+  outline: none;
+  border: none;
+  font-size: 0.75rem;
+  width: 100%;
+  padding: 2px 0;
+  color: ${({ theme }) => theme.core.textColor};
+  border-bottom: 1.5px solid ${({ theme }) => theme.primary[5]};
+  caret-color: ${({ theme }) => theme.primary.base};
+  &:focus {
+    outline: none !important;
```
iprado78, line 130: Eventually, we should reconsider this.  Outline is important for accessibility.  We just need to get good outline styles.

src/new-client/src/App/Trades/TradesGrid/NumFilter.tsx
```diff
@@ -0,0 +1,230 @@
...
+  border-radius: 2px;
+  margin-bottom: 5px;
+
+  &:hover {
+    background-color: ${({ theme }) => theme.core.backgroundHoverColor};
+    font-weight: ${({ selected }) => (selected ? "bold" : "normal")};
+    text-decoration: underline;
+  }
+`
+export const DropdownWrapper = styled.div`
```
iprado78, line 61: "FilterOptionMenuWrapper"

src/new-client/src/App/Trades/TradesGrid/NumFilter.tsx
```diff
@@ -0,0 +1,230 @@
...
+  .dd-placeholder {
+    padding-right: 20px;
+  }
+
+  i {
+    font-size: 10px;
+  }
+`
+
+export const DropdownMenu = styled.div<{
```
iprado78, line 85: "FilterOptionMenu"

src/new-client/src/App/Trades/TradesGrid/NumFilter.tsx
```diff
@@ -0,0 +1,230 @@
...
+  top: 40px;
+  left: 0px;
+  background-color: ${({ theme }) => theme.primary.base};
+  padding: 6px;
+  box-shadow: 0 7px 26px 0 rgba(23, 24, 25, 0.5);
+  z-index: 100;
+  width: 100%;
+`
+
+export const DropdownOption = styled.div`
```
iprado78, line 100: "FilterOption"

src/new-client/src/App/Trades/TradesGrid/NumFilter.tsx
```diff
@@ -0,0 +1,230 @@
...
+    background-color: ${({ theme }) => theme.core.backgroundHoverColor};
+    font-weight: normal;
+    text-decoration: underline;
+  }
+
+  i {
+    padding-left: 20px;
+  }
+`
+const Input = styled.input`
```
iprado78, line 117: "FilterValueInput"

src/new-client/src/App/Trades/TradesGrid/TableHeadCell.tsx
```diff
@@ -34,42 +43,104 @@ const TableHeadCell = styled.th`
...
 export const TableHeadCellContainer: React.FC<{
   field: ColField
 }> = ({ field }) => {
   const [showFilter, setShowFilter] = useState(false)
-  const ref = useRef<HTMLDivElement>(null)
+  const ref = useRef<HTMLTableHeaderCellElement>(null)
   const { displayMenu, setDisplayMenu } = usePopUpMenu(ref)
   const tableSort = useTableSort()
   const { headerName, filterType } = colConfigs[field]
+  const numeric = filterType === "number" && field !== "tradeId"
```
iprado78, line 65: If trade ids have natural arithmetic order (which in our case they do), you might want to filter by aspects of that order.  If we don't want to rely on that (maybe the nature of an idea will change the future), in maybe we make "filterType" optional and there are just some columns we don't allow filtering on.

src/new-client/src/App/Trades/TradesGrid/TableHeadCell.tsx
```diff
@@ -34,42 +43,104 @@ const TableHeadCell = styled.th`
...
+  const [comparator, setComparator] = useState("Equals")
```
iprado78, line 66: This is a bit of a smell.  We shouldn't have to care about the other applied filters (set/date) at this level.  We should just instantiate the right component (NumberFilter//SetFilter/DateFilter) based on the configured filter type for that field and let the component invoke the hook and get the options.  I realize you are solving the bug we had.  But I think we need another solution.  @josepot ?

src/new-client/src/App/Trades/TradesState/filterState.ts
```diff
@@ -5,6 +5,48 @@ import { Trade, trades$ } from "services/trades"
...
+  InRange: "In range",
+}
+
+export interface NumFilterContent {
+  comparator: ComparatorType
+  value1: number | null
+  value2?: number | null
+}
+
+export type DistinctNums = {
```
iprado78, line 33: As we move to having different filter types, I think we need to make these types conditional.   Forex, on my branch:

```
export const colFilterMap = mapObject(colConfigs, (val) => val.filterType)

export type DistinctValues = {
  [K in ColField]: typeof colFilterMap[K] extends "set"
  ? Set<Trade[K]> : null
}```

File: src/new-client/src/App/Trades/TradesState/filterState.ts [hunk 12]
```diff
@@ -47,6 +89,11 @@ interface FilterEvent {
 interface ColFieldToggle extends FilterEvent {
   value: unknown
 }
+
+interface NumFieldToggle extends FilterEvent {
```
iprado78, line 93: I think is less a toggle than a "NumFilterSet".  Toggling "off" is the reset event, right?

src/new-client/src/App/Trades/TradesState/filterState.ts
```diff
@@ -55,6 +102,49 @@ export const [filterResets$, onFilterReset] = createListener(
   (field: ColField) => ({ field } as FilterEvent),
 )
 
+export const [colFilterNum$, onColFilterEnterNum] = createListener(
+  (field: ColField, value: NumFilterContent) =>
+    ({ field, value } as NumFieldToggle),
+)
+
+export const numberFilters$ = mergeWithKey({
+  toggle: colFilterNum$,
```
iprado78, line 111: See above re: event type

iprado78 (changes requested):

src/new-client/src/App/Trades/TradesState/filterState.ts
slcfv3: Do you want me to combine different filters into one observable? Or you just want to combine different types together?

src/new-client/src/App/Trades/TradesGrid/NumFilter.tsx
slcfv3: Yeah, I just mimic the style of old client, but we can discuss it later.

src/new-client/src/App/Trades/TradesState/filterState.ts
iprado78: I don't have it completely worked out.  But something like this:
```
export enum DateFilterType {
  equals = "Equals",
  greaterThan = "Greater Than",
  lessThan = "Less Than",
  notEqual = "Not Equal",
  inRange = "In Range",
}

export enum NumberFilterType {
  equals = "Equals",
  greaterThan = "Greater Than",
  greaterThanOrEqual = "Greater Than or Equal",
  lessThan = "Less Than",
  lessThanOrEqual = "Less Than or Equal",
  notEqual = "Not Equal",
  inRange = "In Range",
}

export type DistinctValues = {
  [K in ColField]: typeof colFilterMap[K] extends "set"
  ? Set<Trade[K]> : null
}

export type AppliedFilters = {
  [K in ColField]: typeof colFilterMap[K] extends "date"
    ? DateFilterType
    : typeof colFilterMap[K] extends "number"
    ? NumberFilterType
    : typeof colFilterMap[K] extends "set"
    ? Set<Trade[K]>
    : never
}```

File: src/new-client/src/App/Trades/TradesGrid/TableHeadCell.tsx [hunk 14]
```diff
@@ -36,46 +41,80 @@ const TableHeadCell = styled.th`
     display: inline-block;
   }
 `
+const FlexWrapper = styled.div<{ headerFirst: boolean }>`
```
iprado78, line 44: Nice to see it was this smple

iprado78 (approved):
Looks good to me.  Check to see if @josepot wants any changes before merge.  Changes can also be potentially captured in my DateFilter PR that will work on top of this, which will hopefully be available by tomorrow.

josepot (approved):

PR #1783, opened 2021-02-03
Fix/col filter selection persistence

iprado78:
Fixes:

1. Import error on ExecutionStatus
2. Set filter selection state dropping
a. This handles https://github.com/AdaptiveConsulting/ReactiveTraderCloud/pull/1781#discussion_r569465572
b. Subscribe wrapper is to avoid "flash"

### DIFF
--- a/src/new-client/src/App/Trades/TradesGrid/SetFilter.tsx
+++ b/src/new-client/src/App/Trades/TradesGrid/SetFilter.tsx
@@ -111,7 +111,7 @@ const [useFilterOptions] = bind(
 )
 
 export const SetFilter: React.FC<SetFilterProps> = ({ field, parentRef }) => {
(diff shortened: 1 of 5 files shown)


src/new-client/src/App/Trades/TradesGrid/SetFilter.tsx
```diff
@@ -111,7 +111,7 @@ const [useFilterOptions] = bind(
 )
 
 export const SetFilter: React.FC<SetFilterProps> = ({ field, parentRef }) => {
-  const selected = useAppliedFieldFilters(field)
+  const selected = useAppliedFieldFilters(field) as Set<string>
```
iprado78, line 114: This is a temporary cheat.  Will get handled in later PR with better types

slcfv3 (approved):

PR #1785, opened 2021-02-04
feat(new-client): add test for emitTooLong

slcfv3:


src/new-client/src/utils/__tests__/emitTooLong.test.ts
```diff
@@ -0,0 +1,35 @@
...
+
+const response = {
+  response: "I am a response",
+}
+
+describe("utils/emitTooLong", () => {
+  it("should emit normal response if response is faster than timer", () => {
+    testScheduler().run(({ expectObservable, cold }) => {
+      const input = cold("    ---(a|)", { a: response })
+      const expectedOutput = "---(b|)"
```
josepot, line 16:
```suggestion
      const input = cold("    2s (a|)", { a: response })
      const expectedOutput = "2s (b|)"
```

It's better if we test the limit :slightly_smiling_face:

src/new-client/src/utils/__tests__/emitTooLong.test.ts
```diff
@@ -0,0 +1,35 @@
...
+      expectObservable(result).toBe(expectedOutput, {
+        b: response,
+      })
+    })
+  })
+
+  it("should emit too long message if response is slowlier than timer", () => {
+    testScheduler().run(({ expectObservable, cold }) => {
+      const input = cold("    3s (a|)", { a: response })
+      const expectedOutput = "2s b 999ms (c|)"
```
josepot, line 27:
```suggestion
      const input = cold("    2s -(a|)", { a: response })
      const expectedOutput = "2s b(c|)"
```

This suggestion is testing the limit and also making sure that the marbles are better aligned, which makes it easier to understand what's happening in each moment.

josepot (approved):

PR #1786, opened 2021-02-08
feat(new-client): status service test(745)

slcfv3:


src/new-client/src/services/status.test.tsx
```diff
@@ -0,0 +1,148 @@
...
+    it("should show the service health again once the service is reconnected", async () => {
+      const mockStream = new BehaviorSubject(mockSource.a)
+      whenWatch("status", mockStream)
+      const { result } = renderUseStatus()
+      expect(result.current).toEqual(
+        alignResultTimestamp(mockResult.a, result.current),
+      )
+
+      await reactAct(async () => {
+        await wait(2001)
```
josepot, line 132: unfortunately this makes tests extremely slow to run, so could you try using [Jest Timer Mocks](https://jestjs.io/docs/en/timer-mocks) instead?

josepot (changes requested):

src/new-client/src/services/status.test.tsx
slcfv3: sure, I will try

josepot (changes requested):
This is good stuff! But could you please create a `status` folder and make sure that we don't export the raw types publicly, basically follow the same approach that we used for the other services that have tests. Thanks!

slcfv3:
> This is good stuff! But could you please create a `status` folder and make sure that we don't export the raw types publicly, basically follow the same approach that we used for the other services that have tests. Thanks!

no problem

josepot (approved):

PR #1791, opened 2021-02-12
feat(new-client):test equals function(750)

slcfv3:


iprado78 (commented):
1 other thing you could check to be exhaustive:  

```

// Inherited properties should not be treated as equivalent

const object1 = {
  a: 1,
  b: 2
}

const object2 = Object.create(object1)

equals(object1, object2)  should be false
```

slcfv3:
> 1 other thing you could check to be exhaustive:
> 
> ```
> 
> // Inherited properties should not be treated as equivalent
> 
> const object1 = {
>   a: 1,
>   b: 2
> }
> 
> const object2 = Object.create(object1)
> 
> equals(object1, object2)  should be false
> ```

yeah, good point. I will add it

iprado78 (approved):

PR #1794, opened 2021-02-16
CI enhancements: faster builds, resiliency, new client deployment

garyadaptive:
## Headline changes

1. Builds for different services are now independently versioned and only rebuilt where source changes are detected.
2. The new client is added to the build pipeline.
3. A parallel deployment is created for the new client, resolvable via a distinct URL (posted to the PR comment thread).
4. All pull requests (on non-fork branches) are automatically deployed to the cluster.
5. Seeding reference data now waits for the eventstore deployment to be ready.

### Independent build versions

The existing implementation calculates a single build version for the entire project on each commit and build all the services with this version. The downside is that CI must build every single service for every single commit, even when they haven't changed. This is both slow and unnecessary.

This PR changes this behaviour to calculate build version for each service. This is done by taking the object hash for the source directory of each service. For example, the build version for the new client is calculated with:

```bash
git ls-tree HEAD -- src/new-client | awk '{ print $3 }'
```

This gives a shasum for the new-client source directory, derived from all of its committed files. Any time a source file inside new-client is changed, the object hash will update accordingly.

The downside to this approach is that there is not a single, predictable version number for a whole deployment. I anticipate that we'll move to a hybrid model where production (and possibly UAT) images are also tagged with the current git tag (via `git describe`) and PR/dev releases are tagged with hash versions.

### New client deployment

The ongoing work to the new client is now deployed to the cluster. A Dockerfile has been added to the new-client directory, which runs install, test and build, and outputs an image based on nginx. This image deployed to the cluster.

The cluster gateway, which fans out traffic to the different deployment namespaces, has some additional logic which detects a `beta` component in the URL and exposes the new client. For example, this PR has two deployments: https://pr1794.lb.adaptivecluster.com (old) and https://pr1794.beta.lb.adaptivecluster.com (new). This is all handled by DNS resolution inside the cluster (kube-dns).

### Seeding reference data - fix race condition

There is currently a race condition when we seed reference data to the eventstore. We seed data via a Kubernetes batch job, which is created at the same time as the eventstore deployment. This means it is possible (and highly likely) that the eventstore will not be ready when we attempt to seed. This results in the frequent "Authentication Error" message and missing data in the deployment. This PR adds a step to wait for the deployment using a readiness probe which checks for the HTTP port of the eventstore to be exposed. This ensures that seeding will not occur too early.

## Future work

- [ ] Derive services and build contexts by parsing docker-compose.yml
- [ ] Move unit tests into a separate workflow
- [ ] Run integration tests in the testing workflow
- [ ] Integrate Lighthouse CI for perf testing the new client
- [ ] Remove client dependency on OpenResty (for openfin manifest template expansion)
- [ ] Push images to GCR (the container registry on Google Cloud) for faster deployments
- [ ] Add capacity to the cluster to accommodate more concurrent deployments
- [ ] Set up KEDA to scale down idle PR deployments


src/new-client/nginx/nginx.conf
```diff
@@ -0,0 +1,63 @@
...
+    uwsgi_temp_path       /var/run/openresty/nginx-uwsgi;
+    scgi_temp_path        /var/run/openresty/nginx-scgi;
+
+    sendfile        on;
+    #tcp_nopush     on;
+
+    #keepalive_timeout  0;
+    keepalive_timeout  65;
+
+    #gzip  on;
```
iprado78, line 60: Reason not to compress at this layer?

src/services/kubernetes/per-deployment/new-client.yaml
```diff
@@ -0,0 +1,48 @@
+apiVersion: v1
```
iprado78, line 1: Not super consequential for a UI, but would it make sense to tie this dynamically to a package.json version number?  Or a concatenation of package.json version and git hash?  @josepot

iprado78 (approved):

src/new-client/nginx/nginx.conf
garyadaptive: I didn't even think about it, to be honest. I copied the nginx config from client. But yeah, I don't think there's any harm in compression. I'll enable it.

src/services/kubernetes/per-deployment/new-client.yaml
garyadaptive: This refers to the kubernetes api version, nothing to do with our apps.

garyadaptive: The BUILD_VERSION you see a few line below is where we inject our version number. In the case of ephemeral PR deployments, this will be a hash calculated from the file tree. In permanent deployments, it uses git-describe (and therefore a derivation of git tags), which tracks the version of the project as a whole.

src/new-client/nginx/nginx.conf
garyadaptive: Mind you, it might be something to do with openresty, which we use to enable templating the openfin manifests. There are a few github issues around the web that suggest compatibility issues. Maybe that's the reason we disabled it in the first place. I'd like to get rid of openresty in the future as it complicates an otherwise static deployment (that could just be in a storage bucket behind a cdn).

PR #1797, opened 2021-02-16
test(new-client): test tile component too long execution(753)

slcfv3:
This test works well so far but will take over 5s to finish since I added timeout option in waitFor(), should I use sinon instead for this case?


josepot:
> should I use sinon instead for this case?

Yes! Most definitely! Thanks @slcfv3 !

josepot (approved):
Thanks @slcfv3 !

PR #1803, opened 2021-02-22
feat: show inline market quote when placing trade from launcher (ARTP…

garyadaptive:
This adds the inline quote table to the launcher suggestion area during trade execution. The executed rate is also show in the confirmation message.

https://user-images.githubusercontent.com/43469524/108750511-00af3480-7539-11eb-9890-2eabe0056354.mp4

### DIFF
--- a/src/client/src/apps/SimpleLauncher/spotlight/components/TradeExecutionOverlay.tsx
+++ b/src/client/src/apps/SimpleLauncher/spotlight/components/TradeExecutionOverlay.tsx
@@ -74,7 +74,7 @@ export const TradeExecutionOverlay: FC<TradeExecutionProps> = ({
     return () => {
       subscription.unsubscribe()
     }
-  }, [tradeStatus])
+  }, [tradeStatus, resetAll, executeTradeRequest])
 
   useEffect(() => {
     if (currencyPairs && partialTradeRequest) {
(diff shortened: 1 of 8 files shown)


src/client/src/apps/SimpleLauncher/spotlight/components/TradeExecutionOverlay.tsx
```diff
@@ -0,0 +1,210 @@
...
+}
+
+export const TradeExecutionOverlay: FC<TradeExecutionProps> = ({
+  response,
+  handleReset,
+  handleClearSearchInput,
+}) => {
+  const [requestStatus, setRequestStatus] = useState<TradeExecutionRequestStatus>('incomplete')
+  const [tradeStatus, setTradeStatus] = useState<TradeExecutionStatus>()
+  const [loadingBarStatus, setLoadingBarStatus] = useState<IndeterminateLoadingBarStatus>()
```
iprado78, line 33: I think this is inherited from the existing code, and maybe not for now, but it seems like you could derive loadingBarStatus from the other pieces of state rather than having to explicitly set it.

src/client/src/apps/SimpleLauncher/spotlight/components/useSpotPrice.ts
```diff
@@ -0,0 +1,32 @@
...
+    setSpotPrice(undefined)
+  }, [currencyPair, setSpotPrice])
+
+  useEffect(() => {
+    if (!priceService || !currencyPair) {
+      return
+    }
+
+    const subscription = priceService
+      .getSpotPriceStream({ symbol: currencyPair })
```
iprado78, line 19: Any reason you can't just take(1)?  That would automatically cleanup the subscription each time the effect runs.  You wouldn't need worry about giving it a handle and manually unsubscribing on unmount.

iprado78 (dismissed):

src/client/src/apps/SimpleLauncher/spotlight/components/useSpotPrice.ts
garyadaptive: We need a streaming price, not just the first price.

PR #1807, opened 2021-02-23
Feature/rfq

iprado78:
@josepot - have more work to do tomorrow around the cancelling the timer on rejection and some styles, but, if you have time to take a look before standup, let me know if you see red flags in the approach so that I can refactor.


iprado78:
New PR for same functionality

PR #1810, opened 2021-02-25
Feature/rfq flow

iprado78:


josepot (dismissed):
Nice! I've some suggestions on how to improve the `rfqTileStateMap$`, but we can always make those improvements later. Thanks a lot for this!

PR #1814, opened 2021-03-02
Fix: yarn start

josepot:
Yesterday I [upgraded the dependencies of the "new-client"](https://github.com/AdaptiveConsulting/ReactiveTraderCloud/pull/1812) and I also [added a commit for leveraging one of the new utils of `@react-rxjs/utils`](https://github.com/AdaptiveConsulting/ReactiveTraderCloud/pull/1812/commits/76b098521d01c3b6cb2e6eb6f5e35c4c2ab3f8a7). After each change I made sure that the production build was working properly and I did check whether we were having improvements on the Lighthouse metrics or not, and everything was working fine.

However, a few hours later I got messages from @iprado78 and from Julie letting me know that the `yarn start` script was no longer working:

image

The error on the image doesn't make any sense, because optional chaining it's in stage-4, it's supported by the current LTS version of node and I don't see any reason why the `esm` build of a modern library shouldn't be using it these days. Also, this problem only happens on the "dev" build, not on the prod build... So, I was really puzzled about this. First I thought about changing the target of the esm build of react-rxjs to `es2019`, but that felt wrong. So, after digging into this a lot more I found that this is a long standing issue of CRA: https://github.com/facebook/create-react-app/issues/9468, https://github.com/facebook/create-react-app/issues/9749 .

A work-around is to use the same `browserslist` for both production and development, which is what [this commit does](https://github.com/AdaptiveConsulting/ReactiveTraderCloud/pull/1814/commits/d909b4f556b55c0e91257342058c3d44a2cc2d74).

### DIFF
--- a/src/new-client/package.json
+++ b/src/new-client/package.json
@@ -72,18 +72,11 @@
       "react-app/jest"
     ]
   },
-  "browserslist": {
-    "production": [
-      ">0.2%",
-      "not dead",
-      "not op_mini all"
-    ],
-    "development": [
-      "last 1 chrome version",
-      "last 1 firefox version",
-      "last 1 safari version"
-    ]
-  },
+  "browserslist": [
+    ">0.2%",
+    "not dead",
+    "not op_mini all"
+  ],
   "devDependencies": {
     "@testing-library/react-hooks": "^5.1.0",
     "@types/d3": "6.3.0",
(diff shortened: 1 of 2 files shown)


garyadaptive:
Can you add a description to this PR explaining the fix please? I'd like to understand the context.

garyadaptive (approved):
LGTM

PR #1817, opened 2021-03-05
test(new-client): tests and fixes for services/currencyPairs

josepot:


src/new-client/src/services/currencyPairs/currencyPairs.test.tsx
```diff
@@ -0,0 +1,229 @@
...
+  [MockSymbols.GBPJPY]: {
+    symbol: "GBPJPY",
+    ratePrecision: 1,
+    pipsPosition: 1,
+    base: "GBP",
+    terms: "JPY",
+  },
+}
+
+const useCurrencyPairWithUpdates = (symbol: MockSymbols) => {
```
iprado78, line 58: Do we want to use this as a pattern in all our tests and create a utility that accepts a hook and returns the hook with an update-count closure?

src/new-client/src/services/currencyPairs/types.ts
```diff
@@ -10,7 +10,7 @@ export interface RawCurrencyPairUpdate extends CollectionUpdate {
   CurrencyPair: CurrencyRaw
 }
 
-export interface RawCurrencyPairUpdates extends CollectionUpdates {
```
iprado78, line 13: We do this for Trades too.  Was there a prior world in which the API also sent diffs/updates instead of state-of-the-world? @garyadaptive?

josepot: I think that this has to do with the way that the old API was working. The thing is that -at least for the currencyPairs- we are not taking those fields into account, and the code of the current client doesn't take them into account either, so I thought that I would just remove it... That makes the mocks easier to build and I don't see the point in keeping stuff in the interface that we ignore.

src/new-client/src/services/currencyPairs/currencyPairs.test.tsx
josepot: maybe? There are lots of things that could be "abstracted" and yes, this could be one of those... If and when we start using it in other tests we can create that abstraction, for sure.

src/new-client/src/services/currencyPairs/currencyPairs.ts
```diff
@@ -41,13 +41,14 @@ export const currencyPairUpdates$ = getStream$<RawCurrencyPairUpdates>(
 export const [useCurrencyPairs, currencyPairs$] = bind(
   currencyPairUpdates$.pipe(
     scan((acc, { updateType, currencyPair }) => {
+      const result = { ...acc }
```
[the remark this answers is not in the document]
josepot: It's not causing a real problem right now because we never unsubscribe/resubscribe to this endpoint. However, the problem was coming up with the tests, because since `scan` was mutating the `seed` object, then [this test](https://github.com/AdaptiveConsulting/ReactiveTraderCloud/pull/1817/files#diff-9f2a03f10fee0cef0dc7ab30b59f019ce9f1c2a64d475f782840425b5954cedbR135) was not passing b/c the previous test was setting that value and it was still around. So, yeah, thanks to the tests we've detected a malfunctioning service.

PR #1824, opened 2021-03-09
feat: update caching and compression on gateway proxy

garyadaptive:
This enables gzip compression and content caching on the ingress gateway.
Lighthouse report: https://lighthouse-dot-webdotdevsite.appspot.com//lh/html?url=https%3A%2F%2Fpr1824.beta.lb.adaptivecluster.com%2F


josepot (approved):
Awesome! Would it be possible to also enable brotli for the browsers that support it :pray: ?

PR #1828, opened 2021-03-10
fix: add strikethrough in table to rejected trades

iowayankee:


garyadaptive (changes requested):
Looks good, but let's lift the styles into a styled component. We could extend `TableBodyRow` to take a `status` prop and apply a conditional style based on that. The styles as-is are not scoped to a component and could have side effects elsewhere.

garyadaptive (approved):
Looks good, thanks! :heavy_check_mark:

PR #1829, opened 2021-03-11
chore(new-client): preload key requests

josepot:
Currently when we run the Lighthouse performance test we can appreciate the following suggestion: "Preload key requests", and then the report shows a list with 6-7 assets that should have been pre-loaded with a `<link rel="preload />` tag.

Lighthouse is correct, those assets are being loaded immediately after the parser evaluates the "shell" bundle, so it makes sense to have them as "preload" links in the `index.html` file. I thought that Vite would be able to do that automatically if instead of declaring our lazy components like this:

```ts
const LiveRatesCore = lazy(() => import("./LiveRatesCore"))
```

we declared them like this:

```ts
const mod = import("./LiveRatesCore")
const LiveRatesCore = lazy(() => mod)
```

The reason being that with the first declaration the bundler can't know when that component would be used (mounted). However, with the second declaration the bundler should IMO be able to figure out that the dynamic import is being requested during bootstrap time... So, I think that ideally Vite should handle those cases, that's why I've opened this issuse: https://github.com/vitejs/vite/issues/2460 .

In the meanwhile, though, I've found [a rollup plugin](https://github.com/bennypowers/rollup-plugin-modulepreload) that we can use with Vite for now if we tweak it a bit... So, that's what this PR is about.

### DIFF
--- a/src/new-client/package-lock.json
+++ b/src/new-client/package-lock.json
@@ -60,6 +60,7 @@
         "lint-staged": "^10.5.4",
         "prettier": "^2.2.1",
         "react-test-renderer": "^17.0.1",
+        "rollup-plugin-modulepreload": "^1.2.3",
         "sinon": "^9.2.4",
         "ts-jest": "^26.5.3",
         "typescript": "^4.1.2",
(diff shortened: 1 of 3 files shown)


josepot:
Nice! I love the preview :star_struck:! So, yep, this is indeed making a difference in terms of performance.

Just to be clear, though: this is a workaround while we wait to have some resolution on the issue that I've opened on Vite's repo. If the resolution is that this is the expected behavior, then what I will do is to create a custom plugin for our needs based on [this code](https://github.com/bennypowers/rollup-plugin-modulepreload/blob/master/index.js). This is one of the reasons why I love Vite, because we can easily customize it :slightly_smiling_face:. If on the other hand Vite makes a fix/improvement to support this behavior OOTB, then I will get rid of this when we upgrade Vite.

garyadaptive (approved):

PR #800, opened 2018-08-21
[ARTP-193] Delayed font loading

keshaldesai:
Added html tags which will "preload" fonts required to display bold and italicized/bold fonts.

PR #1831, opened 2021-03-11
fix(new-client): Fix RFQ Timer on FF (#800)

josepot:
I thought that I could get away with having just one update... It turns out that I was wrong :sweat_smile: Also, I initially wanted to make the component resilient to prop-changes, but we don't really have that need ATM, and even if we had it, we could just use the key property to make it work, so I got rid of that logic too.


garyadaptive (approved):
Looks good :tada:

PR #1832, opened 2021-03-11
test(new-client): trades core (792)

iprado78:
This is just meant to lay the framework for adding tests to the trade components.  Because a lot of the functionality crosses different components (e.g., a filter affects the header, footer, and different parts of the grid), I think it's best if all the tests are grouped together in a` __tests__` folder and individual test files and can reach across the whole Trades component as necessary.   Could theoretically all be in 1 file, but it would eventually be many k lines long.

I also

1.  Added a regex for Jest to enforce a naming convention for test files (also lets us group test utilities in a __tests__ folder without telling Jest to bypass them if we want)
2. Added aria to the loading bar


josepot:
Looking good! But could you please remove the `yarn.lock` file? This project is no longer using `yarn` :slightly_smiling_face:

src/new-client/src/App/Analytics/Analytics.tsx
```diff
@@ -17,7 +17,9 @@ const AnalyticsWrapper = styled.div`
 
 analytics$.subscribe()
 
-const loader = <Loader minWidth="22rem" minHeight="22rem" />
+const loader = (
+  <Loader ariaLabel="analytics-loader" minWidth="22rem" minHeight="22rem" />
```
josepot, line 21: The `aria-label` attribute is not meant to be used like that. IMO this makes the App less accessible.

> The aria-label attribute is used to define a string that labels the current element. Use it in cases where a text label is not visible on the screen.

This is text that an assistant reader would read to the user. However, that text doesn't really mean anything to anyone but to our tests. Ideally, the loader should have some sort alt text like "Loading..." and then we could use that text to target the Loader that's specific to a particular test. Otherwise, let's use the `getByTestId` query.

Same thing for all the other miss usages.

josepot (changes requested):

src/new-client/src/App/Analytics/Analytics.tsx
garyadaptive: I agree, let's ensure that aria-* attributes are used for accessibility. Getting accessibility right is a big story that we've not really covered before, so I don't want to conflate that with this test work. Let's use a test-id here instead.

src/new-client/jest.config.js
```diff
@@ -1,6 +1,7 @@
 module.exports = {
   preset: "ts-jest",
   testEnvironment: "jsdom",
+  testRegex: "\.test.[t]sx?$",
```
josepot, line 4: I'm pretty sure that the default `testRegex` already matches this, right? If so, I would rather to stick with the default value.

josepot: > This narrows actually narrows the match to force the convention we're using

I rather use the standard convention than to create our own. Sorry, but I couldn't resist using this meme :see_no_evil: 
image

> it also lets us include files in a tests folder that are not tests, like shared mock data, without manually exempting them from the runner

I would prefer not to have files on the tests folder that are not tests. Mock data files belong into their respective `__mocks__` folders. For instance, if we exposed the mocked data that's on the `src/App/Trades/__tests__/mockTrades.ts` through the `src/services/trades/__mocks__/` then we will be able to re-use those mocks for other tests that are mocking the `services/trades`

josepot (approved):

PR #1833, opened 2021-03-11
fix(new-client): add/remove position bubbles (807)

josepot:
I just enjoy deleting code... :sweat_smile:

### DIFF
--- a/src/new-client/src/App/Analytics/Positions/Positions.tsx
+++ b/src/new-client/src/App/Analytics/Positions/Positions.tsx
@@ -7,22 +7,14 @@ import {
   forceX,
   forceY,
   select,
-  Selection,
-  Simulation,
 } from "d3"
-import { distinctUntilChanged, skip } from "rxjs/operators"
-import {
-  BubbleChartNode,
-  redrawCircles,
-  getNodes$,
-  useData,
-  data$,
-} from "./data"
+import { BubbleChartNode, nodes$, useData, data$ } from "./data"
 import { BubbleChart, Title } from "../styled"
 
 // extra pixel amount that nodes in the chart repel each other within
 // for collision detection purposes, a nodes radius is r + COLLIDE_BORDER_WIDTH pixels
 const COLLIDE_BORDER_WIDTH = 2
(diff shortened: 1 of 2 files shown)


iprado78:
> I just enjoy deleting code... 😅

Possible to convert the drag event into a takeUntil for even more free bits?

iprado78 (approved):

garyadaptive (changes requested):
Let's hold off merging this; I think there's still something not quite right. I need to dig into it a bit, but the bubbles are not updating correctly to show positions. Open the old and new client side-by-side and place some trades, you should see the new client gets stuck and only syncs the bubbles after a reload.

josepot:
> Let's hold off merging this; I think there's still something not quite right. I need to dig into it a bit, but the bubbles are not updating correctly to show positions. Open the old and new client side-by-side and place some trades, you should see the new client gets stuck and only syncs the bubbles after a reload.

You are correct! Thanks for catching that @garyadaptive! I was so focus on making the addition/deletion work that I didn't notice that the updates are not working correctly... I think that I know how to fix it, though. Thanks again!

josepot:
I think that it should work fine now... I will add more snapshot tests if it's confirmed that this is the expected behavior.

garyadaptive (approved):
All looks good to me :)

josepot:
> Possible to convert the drag event into a takeUntil for even more free bits?

Sorry, but I'm not following... What do you mean?

iprado78:
> > Possible to convert the drag event into a takeUntil for even more free bits?
> 
> Sorry, but I'm not following... What do you mean?

I haven't looked carefully at the full context of the drag event, and I know we are mostly not using Rx for data flow in Analytics, but it seems like you could easily model the drag event and the chart side effects as a set of streams (take the 

> > I just enjoy deleting code... 😅
> 
> Possible to convert the drag event into a takeUntil for even more free bits?

Concat drag event with chart side effects until drag event ends.  You are essentially doing this already just not in Rx.  Better your way with the Analytics code we have.  Just imagining what a more Rx version of the Analytics code would look like.

PR #1835, opened 2021-03-15
Fix Analytics panel responsive issue with FF

josepot:
As I was addressing this issue, I noticed that the `Resizer` component needed some love, which is something that @bhavesh-desai-scratch had already pointed out to me in the past, so I decided to fix that too in a separate commit.

### DIFF
--- a/src/new-client/src/components/Resizer.tsx
+++ b/src/new-client/src/components/Resizer.tsx
@@ -2,6 +2,7 @@ import { useState, useRef } from "react"
 import styled from "styled-components"
 
 const ResizerStyle = styled.div`
+  width: 100%;
   height: 100%;
 `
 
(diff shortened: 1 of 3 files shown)


src/new-client/src/components/Resizer.tsx
```diff
@@ -42,68 +43,53 @@ interface Props {
...
-    if (height > 90) height = 90
-
-    // Block heights that would hide bottom div
-    if (height < 10) height = 10
-
-    setHeight(height)
+  const startDragging = useRef<() => void>()
+  if (!startDragging.current) {
+    startDragging.current = () => {
+      console.log("hello there")
```
garyadaptive, line 55: Delete this

garyadaptive (commented):
Looks good. Just one console statement to remove.

src/new-client/src/components/Resizer.tsx
josepot: fixed! thanks!

PR #1842, opened 2021-03-17
chore(new-client): fix Husky (822)

josepot:


iprado78 (approved):

garyadaptive (changes requested):
Husky v5 is currently installed in the root package.json.

1. Do we need it in botth?
2. If so, let's keep the versions in sync.

garyadaptive:
Having Husky in both places has caused problems for me in the past, depending on the order that I install dependencies. I only ever check the project out from the root. Let's remove this from the root package.json, since it's not adding much value there anyway.

PR #1848, opened 2021-03-18
fix(new-client): improve buble-chart

josepot:
I noticed some odd behaviors with the current implementation: mostly issues when there are no trades and therefore no positions and then new trades of different currencies add bubbles into the chart. This PR addresses that, while also makes an improvement so that the movement of the bubbles don't trigger long ending updates on the DOM.

### DIFF
--- a/src/new-client/src/App/Analytics/Positions/Positions.test.tsx
+++ b/src/new-client/src/App/Analytics/Positions/Positions.test.tsx
@@ -1,5 +1,5 @@
 import { Subscribe } from "@react-rxjs/core"
-import { render, screen, act, fireEvent } from "@testing-library/react"
+import { render, screen, act, fireEvent, waitFor } from "@testing-library/react"
 import { BehaviorSubject } from "rxjs"
 import { TestThemeProvider } from "@/utils/testUtils"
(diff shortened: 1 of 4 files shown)


garyadaptive (changes requested):
A couple of new bugs now:

1. The CAD tooltip appeared over the EUR bubble
2. The label is missing from the CAD bubble

(On Firefox, if this makes a difference)

Screenshot_20210319_103501

garyadaptive:
It happens when you trade a currency pair that doesn't currently have a position (i.e. adds a new position bubble). The bubbles for both currencies in that pair break.

josepot:
Thanks @garyadaptive ! I think that I've fixed it now... You would never believe what it was :roll_eyes:  Anyways, please ping me when you try it again, just in case that I didn't understand how to reproduce the issue. Thanks!

josepot:
I'm 99.99% sure that the issue's that @garyadaptive pointed out in his prev comments has now been addressed. Also, this version behaves much better than what we currently have deployed. So, since @garyadaptive is not in today and I would like to have a better bubble-chart ASAP, I'm going to merge this. However, if you find further issues, please let me know it :pray:. Thanks!

PR #1852, opened 2021-03-22
fix(new-client): pending trades with a different background color

juliezhudev:


src/new-client/src/App/Trades/TradesGrid/TradesGrid.tsx
```diff
@@ -24,14 +24,15 @@ const TableHeadRow = styled.tr`
...
-const TableBodyRow = styled.tr`
+const TableBodyRow = styled.tr<{ pending?: boolean }>`
   &:nth-child(even) {
     background-color: ${({ theme }) => theme.core.darkBackground};
   }
   &:hover {
     background-color: ${({ theme }) => theme.core.alternateBackground};
   }
   height: 2rem;
+  ${({ pending }) => pending && "background-color: #5e626d"}
```
josepot, line 35: It looks good on dark-mode, but it doesn't look good on light-mode. The best way to ensure that it looks good on both modes is to use one of the colors defined under `theme.core`.

josepot (changes requested):

josepot (approved):

PR #1855, opened 2021-03-24
new client: update dependencies

josepot:
There are 3 different atomic commits in this PR:
- Update outdated dependencies
- Leverage `createKeyedSignal` from `@react-rxjs/utils` which greatly simplifies working with "factory Observables" (no more `split`/`collect` madness)
- Rename `createListener` to `createSignal`, since the former has been flagged as deprecated.


src/new-client/src/App/LiveRates/Tile/Rfq/Rfq.state.ts
```diff
@@ -79,16 +61,12 @@ const [, _getRfqState$] = symbolBind((symbol) =>
...
-                    mergeMap(() =>
-                      concat(
-                        of({
-                          state: QuoteState.Rejected as const,
-                          payload,
-                        }),
-                        timer(REJECT_TIMEOUT).pipe(mapTo(INIT)),
-                      ),
-                    ),
+                  merge(getRejection$(symbol), timer(payload.timeout)).pipe(
```
iprado78, line 64: why prefer merge and take(1) to race?

iprado78 (approved):

src/new-client/src/App/LiveRates/Tile/Rfq/Rfq.state.ts
josepot: No reason in particular. The thing is that now `getRejection$(symbol)` doesn't complete after it emits, so I have to put a `take(1)` somewhere. I decided to put it after the combination of the 2 observables and that made `race` look redundant/silly, because that `take(1)` already ensures that when one of the 2 emits, then we unsubscribe from the other observable, so I changed from `race` to `merge`. I could have done this instead, though:

```ts
race([getRejection$(symbol).pipe(take(1)), timer(payload.timeout)]).pipe(
  mapTo({
    state: QuoteState.Rejected as const,
    payload
  })
);
```
It's the same.

josepot: Actually, on a second thought, I've changed it to `race`... I realized that I can move the `take(1)` further down, which simplifies the code quite a bit and IMO it makes it easier to understand the flow of the state.

iprado78 (approved):

PR #1859, opened 2021-03-26
chore(new-client): improve RFQ state management

josepot:
I'm preparing a presentation on react-rxjs and I'm planing to sue the rfq state as an example... Because of that, I decided to make a few improvements to make the code more readable and easier to understand, I also decided to be a bit more explicit about the types of the `useRfqState`.

One thing that I did notice, though, is that despite the fact that the behavior hasn't changed at all I've had to readjust the tests of `Rfq.test.ts`. I really think that those tests are kinda testing the internals and that we should be testing the rfq workflow through in the `Tile.test.tsx`, so let's create a task for that in our backlog.

### DIFF
--- a/src/new-client/src/App/LiveRates/Tile/Notional/Notional.tsx
+++ b/src/new-client/src/App/LiveRates/Tile/Notional/Notional.tsx
@@ -1,6 +1,6 @@
 import styled from "styled-components"
 import { FaRedo } from "react-icons/fa"
-import { useRfqState, QuoteState } from "../Rfq"
+import { useRfqState, QuoteStateStage } from "../Rfq"
 import { symbolBind, useTileCurrencyPair } from "../Tile.context"
 import {
   InputWrapper,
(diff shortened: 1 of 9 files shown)


iprado78:
> I'm preparing a presentation on react-rxjs and I'm planing to sue the rfq state as an example... Because of that, I decided to make a few improvements to make the code more readable and easier to understand, I also decided to be a bit more explicit about the types of the `useRfqState`.
> 
> One thing that I did notice, though, is that despite the fact that the behavior hasn't changed at all I've had to readjust the tests of `Rfq.test.ts`. I really think that those tests are kinda testing the internals and that we should be testing the rfq workflow through in the `Tile.test.tsx`, so let's create a task for that in our backlog.

I think as you move to more behavioral, integration-like tests, it will be natural for them to live at a higher level in the component hierarchy.  This is definitely true of Trades.  If the test files get too large to navigate, you can still split into flows and re-use setup: `Tile.rfq.test.tsx`, `Tile.execution.test.tsx`.   This is obviously a matter of balance/taste.

iprado78 (approved):

josepot:
>  you can still split into flows and re-use setup: Tile.rfq.test.tsx, Tile.execution.test.tsx. This is obviously a matter of balance/taste

This is exactly what I suggested to @juliezhudev in a call that I had with her :slightly_smiling_face: . I'm happy to see that I'm not the only one having these kind of crazy ideas :sweat_smile:

PR #1873, opened 2021-04-08
fix(new-client): replace tile css grid with flexbox

juliezhudev:


josepot (approved):
Nicely done! Thanks @julienhamilton !

PR #1887, opened 2021-04-23
Exclude storybook and styleguide from service worker routing

garyadaptive:
Once registered, the service worker of the new client intercepts requests to /storybook and /styleguide and redirects to the Reactive Trader UI. This PR adds exceptions for those two path prefixes.

For comparison, visit: https://www.reactivetrader.com/storybook and https://pr1887.lb.adaptivecluster.com/storybook. In the former, it is not possible to open storybook once the service worker is registered (without force-reloading). The latter always works.

### DIFF
--- a/src/new-client/src/sw.js
+++ b/src/new-client/src/sw.js
@@ -34,6 +34,14 @@ registerRoute(
       return false
     }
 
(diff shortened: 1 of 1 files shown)


josepot (approved):
Makes sense! Thanks!

PR #1892, opened 2021-04-28
Storybook tradesgrid

juliezhudev:


garyadaptive (changes requested):
Looks good, nice work!

One minor change: can we rename this story to "Blotter" please? We publish storybook as part of our marketing materials and this component is referred to as a blotter everywhere else.

josepot (approved):

PR #1900, opened 2021-06-02
feat(new-client-hydra): handle reconnections

josepot:


src/new-client/src/services/connection.ts
```diff
@@ -25,3 +26,25 @@ export const [useConnectionStatus, connectionStatus$] = bind(
...
+  )
+
+  const connectionDate$ = connectionStatus$.pipe(
+    filter((status) => status !== ConnectionStatus.CONNECTING),
+    map((status) => ({ status, date: Date.now() })),
+  )
+
+  return combineLatest([sourceDate$, connectionDate$]).pipe(
+    map(([sourceDate, connection]) => {
+      if (connection.status !== ConnectionStatus.CONNECTED) return false
```
garyadaptive, line 45: If we only care about the CONNECTED status, can this be lifted into the filter above?

josepot: Good catch! Thanks!

garyadaptive (approved):

PR #1910, opened 2021-07-19
Changing titles on tiles

fsarda:
Web view 
image

Openfin view
image


josepot (approved):
LGTM!

An interesting alternative would be creating a `DocTitle` component like:

```tsx
const DocTitle: React.FC<{ title: string }> = ({ title, children }) => {
  useLayoutEffect(() => {
    document.title = title
  }, [title])

  return (
    <>{children}</>
  )
}
```

And then do:
```tsx
<Route path="/blotter" render={() => <DocTitle title="trades"><Trades /></DocTitle>} />
```

Or just leave it as it is, of course!

PR #1917, opened 2021-07-29
fix tiles pop out positioning

fsarda:
**UPDATE: this bad positioning seems to be occurring only on my PC, tested on others envs is working fine.**

This PR in meant to fix tiles position when popping out from tiles grid.  Is working only for left column of grid (I have no explanation for this) **Hopefully @garyadaptive knows better what is happening.**

I'm attaching some gifs and screenshots about this issue

1) Opening tiles on left and right column, only works for left sided tiles
opening-tiles

2) Opening tiles from one column format, works properly
opening-tiles-one-column

I've made sure position is properly calculated for all tiles, also set option `includeInSnapshots: false`. Also I've erased openfin cache files and tested.


src/new-client/src/utils/useWindowPositon.ts
```diff
@@ -0,0 +1,14 @@
+import { RefObject } from "react"
+
+export interface Offset {
+  y: number
+  x: number
+}
+
+export const calculate = (ref: RefObject<HTMLDivElement | null>): Offset => {
```
ivanfalanga, line 8: could probably rename this file as it is not a hook any more

src/new-client/src/OpenFin/DetachableTile/DetachableTile.tsx
```diff
@@ -27,16 +28,18 @@ interface Props {
...
-  isTornOut,
-}) => {
+export const DetachableTile: FC<Props> = ({ symbol, view, isTornOut }) => {
   const currencyPair = useCurrencyPair(symbol)
+  const ref = useRef<HTMLDivElement>(null)
 
   return (
-    <TearOutContext.Provider value={isTornOut}>
-      <Wrapper>
+    <TearOutContext.Provider
```
ivanfalanga, line 36: would be good to pass down the ref via prop so that this context does not cause children re-render all the times due to it being an object now instead of a primitive boolean, not too fuss about it if it does not cause issues though

ivanfalanga (approved):

ivanfalanga:
Tested this out on mac and it seems to work, the issue showed above seems to be a setup issue on @fsarda 's machine

fsarda:
Have refactored since I was passing ref unnecessarily on a context

ivanfalanga (approved):

src/new-client/src/OpenFin/DetachableTile/tornOutTiles.ts
```diff
@@ -10,17 +10,36 @@ import {
...
 
 const [tearOutEntry$, tearOut] = createSignal<TearOutEntry>()
 
-export const onTearOut = (symbol: string) => {
+const calculateOpeningWindowCoords = async (coords: Offset) => {
+  if (fin) {
+    const view = await fin.View.getCurrent()
+    const window = await view.getCurrentWindow()
+    const viewBounds = await view.getBounds()
+    const windowBounds = await window.getBounds()
```
josepot, line 24: A common problem when using `async`/`await` is that we can end up making sequential requests that instead could be running in parallel. I think that this code is better in that regard:

```suggestion
    const view = await fin.View.getCurrent()

    const [viewBounds, windowBounds] = await Promise.all([
      view.getBounds(),
      view.getCurrentWindow().then(window => window.getBounds()),
    ])
```

josepot (approved):
LGTM! Although, please consider my comment before merging.

PR #1923, opened 2021-08-06
1185 finsemble target, basic app with router to support rtc-finsemble

algreasley:
Simplest entry point app for Finsemble, and script entries.
Tested out with `VITE_MOCKS=true`  running with https://github.com/AdaptiveConsulting/ReactiveTraderCloud-Finsemble configured to address `http://localhost:1917` as root.

Update: now tested against the deployed target: 
`		"componentRoot": "https://finsemble.env.reactivetrader.com/pull/1923",`

### DIFF
--- a/src/new-client/src/Finsemble/index.tsx
+++ b/src/new-client/src/Finsemble/index.tsx
@@ -1,0 +1,44 @@
+import { StrictMode } from "react"
+import ReactDOM from "react-dom"
+import { GA_TRACKING_ID } from "@/constants"
+import GlobalStyle from "@/theme/globals"
+import { GlobalScrollbarStyle, ThemeProvider } from "@/theme"
+import { FinsembleApp } from "./FinsembleApp"
+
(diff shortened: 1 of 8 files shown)


src/new-client/src/Finsemble/index.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+const { ga } = window
+
+ga("create", {
+  trackingId: GA_TRACKING_ID,
+  transport: "beacon",
+})
+
+ga("set", {
+  dimension1: "finsemble",
```
steinalex, line 38: Do we have the different platform names in an `Enum` somewhere? Might be worth adding if not

steinalex (approved):

steinalex (approved):

src/new-client/src/Finsemble/index.tsx
algreasley: I don't know .. @garyadaptive ?

PR #1924, opened 2021-08-09
fix: build and copy OF manifest during CI

garyadaptive:


algreasley (approved):
I had a quick look at rollup-plugin-copy, btw .. think we might be able to use this in future

PR #1930, opened 2021-08-11
fix: rfq pricing colours (BUG-1189)

steinalex:
Match the previous behaviour when quoting an RFQ

**Before:**
RFQColoursAfter

**After:**
RFQColoursAfter2


algreasley (approved):
LGTM

PR #1937, opened 2021-08-20
Hydra url as /ws (proxied in nginx)

algreasley:
See corresponding deployments PR: https://github.com/AdaptiveConsulting/reactive-deployments/pull/3/files

### DIFF
--- a/src/new-client/src/Finsemble/index.tsx
+++ b/src/new-client/src/Finsemble/index.tsx
@@ -10,7 +10,7 @@ import { noop } from "rxjs"
 
 if (!import.meta.env.VITE_MOCKS) {
   connectToGateway({
-    url: import.meta.env.VITE_HYDRA_URL as string,
+    url: `${window.location.origin}/ws`,
     interceptor: noop,
     useJson: true,
   })
--- a/src/new-client/src/OpenFin/index.tsx
+++ b/src/new-client/src/OpenFin/index.tsx
@@ -10,7 +10,7 @@ import { noop } from "rxjs"
 
 if (!import.meta.env.VITE_MOCKS) {
   connectToGateway({
-    url: import.meta.env.VITE_HYDRA_URL as string,
+    url: `${window.location.origin}/ws`,
     interceptor: noop,
     useJson: true,
   })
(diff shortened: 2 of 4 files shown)


src/new-client/src/Finsemble/index.tsx
```diff
@@ -10,7 +10,7 @@ import { noop } from "rxjs"
 
 if (!import.meta.env.VITE_MOCKS) {
   connectToGateway({
-    url: import.meta.env.VITE_HYDRA_URL as string,
+    url: "/ws",
```
algreasley, line 13: _not complete_

src/new-client/src/OpenFin/index.tsx
```diff
@@ -10,7 +10,7 @@ import { noop } from "rxjs"
 
 if (!import.meta.env.VITE_MOCKS) {
   connectToGateway({
-    url: import.meta.env.VITE_HYDRA_URL as string,
+    url: "/ws",
```
algreasley, line 13: _not complete_

src/new-client/src/Web/index.tsx
```diff
@@ -15,7 +15,7 @@ import { noop } from "rxjs"
 
 if (!import.meta.env.VITE_MOCKS) {
   connectToGateway({
-    url: import.meta.env.VITE_HYDRA_URL as string,
+    url: `${window.location.href}ws`,
```
algreasley, line 18: _not complete_  just added to test quickly

src/new-client/vite.config.ts
```diff
@@ -124,6 +124,12 @@ const setConfig = ({ mode }) => {
     },
     server: {
       port: 1917,
+      proxy: !process.env.VITE_MOCKS && {
+        "/ws": {
+          target: "http://localhost:55000",
```
algreasley, line 129: we need this locally, as we have to run through the nginx running in Docker with its associated config - going direct to wss:<demo WS url> does not work as I think we need those special proxy args - we may be able to apply them to the local vite (http-proxy) config - any thoughts, @garyadaptive ?

ivanfalanga, line 129: Where does this url come from exactly? port 55000 sounds oddly specific, was this something you mapped to locally to connect to the locally running container of the gateway?

algreasley: Docker - just updating this now so the default (if no VITE_MOCKS) will route through to our one and only Hydra endpoint (demo)

ivanfalanga (approved):

PR #1941, opened 2021-09-08
fix: remove dynamic imports from index.tsx

ivanfalanga:


src/new-client/src/Web/serviceWorkerRegistration.ts
```diff
@@ -28,6 +28,9 @@ type Config = {
 export function register(config?: Config) {
   if ("serviceWorker" in navigator) {
     window.addEventListener("load", () => {
+      // This url currently breaks when running against the env gateway as it misses the correct pathname
+      // e.g. web.env.reactivetrader.com/sw.js instead of web.env.reactivetrader.com/branch/hydra-integration/sw.js
```
algreasley, line 32: ah, this must be something to do with the url path modifer in nginx?

src/new-client/src/index.tsx
```diff
@@ -1,16 +1,5 @@
 import { GA_TRACKING_ID } from "./constants"
-
-async function main() {
-  if (__TARGET__ === "web") {
-    await import("@/Web")
-  } else if (__TARGET__ === "openfin") {
-    await import("@/OpenFin")
-  } else if (__TARGET__ === "finsemble") {
```
algreasley, line 8: need to add a main for finsemble I think

ivanfalanga: oops, added

src/new-client/src/Web/serviceWorkerRegistration.ts
ivanfalanga: it does indeed, the correct url would be the origin + the full pathname for the sw.js to be downloaded correctly, however we are still getting issues with that. I want to address this as a separate issue as it only affects the env environments and the solution is not trivial

algreasley: but I think the _correct_ URL is just `origin/sw.js` and the lb/nginx should be sorting out your URL rewriting ..

algreasley: I can see no evidence of this being applied in nginx or elsewhere - my comment was in relation to a conversation with Gary a couple of weeks back, so perhaps that was reverted or maybe never properly applied .. oh well.

algreasley:
just wondering why package-lock.json was changed (no change to package.json)?

src/new-client/public/manifest.json
```diff
@@ -4,48 +4,48 @@
   "description": "Real-time FX trading showcase by Adaptive.",
   "icons": [
     {
-      "src": "/static/media/reactive-trader.ico",
```
algreasley, line 7: as discussed offline, this is not the desired end state .. the deployment was supposed to be set up to strip out the ongoing branch/blah path for the 'env' deployments .. but that can be looked at in the meantime, as other comments, I cannot find this config anywhere, although I am sure Gary pointed it out somewhere ...

src/new-client/src/Web/serviceWorkerRegistration.ts
```diff
@@ -28,7 +28,10 @@ type Config = {
 export function register(config?: Config) {
   if ("serviceWorker" in navigator) {
     window.addEventListener("load", () => {
-      const swUrl = `${window.location.origin}/sw.js`
+      // This url currently breaks when running against the env gateway as it misses the correct pathname
```
algreasley, line 31: this comment out of date now, I guess - this env deployment PWA works:
https://web.env.reactivetrader.com/branch/fix_manifest-file-for-pwa

ivanfalanga: changed thanks

PR #1949, opened 2021-09-29
feat(client): Added Version to Footer

taylor-adaptive:
Added current build version to footer. 
Ported over from `./client` with some changes to make compatible with vite.

image

### DIFF
--- a/.github/workflows/branch.yml
+++ b/.github/workflows/branch.yml
@@ -13,7 +13,7 @@ env:
   BUCKET_WEB: reactive-trader-web-builds
   BUCKET_OPENFIN: reactive-trader-openfin-builds
   BUCKET_FINSEMBLE: reactive-trader-finsemble-builds
-  VITE_HYDRA_URL: wss://trading-web-gateway-rt.demo.hydra.weareadaptive.com
+  VITE_HYDRA_URL: wss://trading-web-gateway-rt-dev.demo.hydra.weareadaptive.com
 
 jobs:
   build:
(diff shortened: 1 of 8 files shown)


algreasley (commented):
All looks ok, but the version does not appear in the deployed branch build - that may be expected for a branch build (as it is looking for a tag) but I am puzzled as to how that VITE_BUILD_VERSION is going to get set in the release build - @ivanfalanga, any ideas?
(only mention I can find is in a Dockerfile under new-client that I don't think is used)

taylor-adaptive:
@algreasley I realised this yesterday evening after I had a session with David. The application isn't served through the Dockerfile when deployed to an environment, so the `VITE_BUILD_VERSION` isn't set. This was a misunderstanding on my part. I will explore the possibilities of injecting the build version through GitHub's workflow and update the PR accordingly.

.github/workflows/branch.yml
```diff
@@ -13,7 +13,7 @@ env:
   BUCKET_WEB: reactive-trader-web-builds
   BUCKET_OPENFIN: reactive-trader-openfin-builds
   BUCKET_FINSEMBLE: reactive-trader-finsemble-builds
-  VITE_HYDRA_URL: wss://trading-web-gateway-rt.demo.hydra.weareadaptive.com
+  VITE_HYDRA_URL: wss://trading-web-gateway-rt-dev.demo.hydra.weareadaptive.com
```
algreasley, line 16: ah, interesting .. so we have both now, I guess

algreasley (approved):
looks good ... nice work on the workflow YAML bits

.github/workflows/branch.yml
taylor-adaptive: Yep, David sorted our reactive-deployments bits yesterday too.
I don't think we actually need/use this variable here in the application, but updated and leaving in for now until I'm more confident on what's happening.

PR #1950, opened 2021-09-29
fix(client): Mobile > Trade confirmation cuts off

taylor-adaptive:
Position `ExecutionMessage` & `Pending` absolutely to fix trade confirmation UI on safari (mobile and desktop

**Before**
Screenshot 2021-09-29 at 13 20 52

**After**
Screenshot 2021-09-29 at 13 21 15

### DIFF
--- a/src/new-client/src/services/trades/index.ts
+++ b/src/new-client/src/services/trades/index.ts
@@ -1,4 +1,5 @@
+import * as tradesTestData from "./__mocks__/mockTrades"
 export * from "./trades"
-export * as tradesTestData from "./__mocks__/mockTrades"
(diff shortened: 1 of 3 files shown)


src/new-client/src/services/trades/index.ts
```diff
@@ -1,4 +1,5 @@
+import * as tradesTestData from "./__mocks__/mockTrades"
 export * from "./trades"
-export * as tradesTestData from "./__mocks__/mockTrades"
```
taylor-adaptive, line 2: I had to change this export as safari doesn't support it
https://caniuse.com/mdn-javascript_statements_export_namespace

rikoe (approved):

PR #1952, opened 2021-10-01
feat(client): Add PWA banner

taylor-adaptive:
Ports over the from `client` to `new-client` with some minor refactoring/reshuffling:
- PWA banner
- Icons
- Utils
- Modal component

Introduces a simple `PlatformContext` so child components know what platform they are running in. This will undoubtedly change/grow when we move into the Openfin and Finsemble implementations.

Updates GitHub workflow actions to set `ENVIRONMENT` variable. This is used to replace `{{environment_suffix}}` in `manifest.json`


src/new-client/vite.config.ts
```diff
@@ -107,25 +107,32 @@ const copyOpenfinPlugin = (dev: boolean) => ({
...
+        {
+          src: "./public/manifest.json",
+          dest: "./dist",
+          transform: (contents) =>
+            contents
+              .toString()
+              // We don't want to show PROD in the PWA name
+              .replace(
+                /{{environment_suffix}}/g,
+                envSuffix === "RROD" ? "" : envSuffix,
```
podefr, line 124: did you mean `PROD` here?

taylor-adaptive:
@podefr thanks for taking a look at this. I was only checking this on mobile, so I had missed the issues you describe on desktop.

I have updated the vite configuration to correctly replace the `BASE_URL` and added `html-transform` to point to the correct manifest file route in our env/branch environments. Now you should be able to see the option to install application on desktop.

In regards to localhost, I was unable to successfully apply the fix you mentioned with the service worker so I can not replicate the style issue in your screenshot. Let's catch up around this when you are back, although we may address this separately as to not block this being deployed (The UI and functionality appear to be working in deployed environments as required)

PR #1955, opened 2021-10-04
fix(new-client): Add autoReconnect option to connectToGateway in all locations

ncucciniello:
Enabled auto reconnect feature provided by hydra. Adds 2 second delay before reconnecting.

As shown in screenshots below, the subscription will re connect after a 2 second delay each hour. 

Screen Shot 2021-10-04 at 3 56 08 PM

Screen Shot 2021-10-04 at 3 50 37 PM

### DIFF
--- a/src/new-client/src/Finsemble/index.tsx
+++ b/src/new-client/src/Finsemble/index.tsx
@@ -14,6 +14,7 @@ export default function main() {
       url: `${window.location.origin}/ws`,
       interceptor: noop,
       useJson: true,
+      autoReconnect: true,
     })
   }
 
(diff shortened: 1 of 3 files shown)


podefr:
we can see that the 2s delay comes from hydra's autreconnect feature: 

https://github.com/AdaptiveConsulting/hydra/blob/ada499506694439b22330d085adfac1306ca0bdc/web-proj/packages/hydra-platform/src/connectToGateway.ts#L64

podefr:
note, for some reason even with autoReconnect not activated, my client seems to automatically reconnect anyway. still trying to reproduce the original issue

PR #1970, opened 2021-10-25
Styleguide newclient

hafidk:


src/new-client/src/App/LiveRates/Tile/Header/TileHeader.tsx
```diff
@@ -76,3 +83,25 @@ export const Header: React.FC = () => {
     </HeaderWrapper>
   )
 }
+
+export const HeaderComponent: React.FC<{
```
taylor-adaptive, line 87: The render of this is the same as `Header` above. 
To reduce duplication we should have a 'dumb' component that simply renders based on props, and then a 'container' component that passes props to the dumb component.

So for this example I would:
- Copy the original `Header` component, maybe call it `HeaderBase` or `HeaderInner` or `HeaderComponent`.
- Remove anything from the component that is a dependency and define those as the component's props. In this case it is data coming from the `useTileContext` hook, the `useDate` hook and the `tearOut` function
- Refactor the original `Header` component to use the new `HeaderInner` component, passing `currencyPair`, `supportsTearOut`, `date` and `onClick` as props.
- Now we could use the new `HeaderInner` component elsewhere outside the context of the application.

src/new-client/src/App/LiveRates/Tile/HistoricalGraph/HistoricalGraph.tsx
```diff
@@ -97,3 +97,59 @@ export const HistoricalGraph: React.FC<HistoricalGraphProps> = ({
     </LineChartWrapper>
   )
 }
+
+export const HistoricalGraphComponent: React.FC<{
```
taylor-adaptive, line 101: Similar to above comment on the Tile Header..

`HistoricalGraph` should be able to use the `HistoricalGraphComponent` to reduce the duplication.

src/new-client/src/App/LiveRates/Tile/HistoricalGraph/HistoricalGraph.tsx
```diff
@@ -97,3 +97,59 @@ export const HistoricalGraph: React.FC<HistoricalGraphProps> = ({
...
+    <LineChartWrapper showTimer={showTimer} ref={ref}>
+      <Svg>
+        <Path
+          stroke={active ? "#5f94f5" : "#737987"}
+          strokeOpacity={0.9}
+          strokeWidth={1}
+          fill="none"
+          d={history}
+        />
+        {!isAtom && (
```
taylor-adaptive, line 118: I'm not sure what `isAtom` is used for. I can only see it being passed to this component in one place, hardcoded as `true`.
I would have expected this component's render to be identical to the render of the original component, passing props where necessary.

src/new-client/src/App/LiveRates/Tile/Tile.tsx
```diff
@@ -132,4 +157,171 @@ const TileContext: React.FC<{
...
 
+const history = generateHistoricPrices(30)
+const dataPoints = getDataPoints<HistoryPrice>((price, idx) => [
+  new Date(idx),
+  price.mid,
+])(history)
+const scales = withScales([0, 125], [0, 75])(dataPoints)
+const HistoryMockSvgPath = toSvgPath(curveBasis)(scales)
+
+export const TileMockComponent: React.FC<MockProps> = ({
```
taylor-adaptive, line 168: This was always going to be the difficult one.
I think creating your own mock component, composing the existing components that make up the Tile, is a good idea.
However, I think it would be good to keep this isolated from the actual Tile component.
Can we move it (and anything else supporting it) to it's own file?

src/new-client/src/App/LiveRates/Tile/Header/TileHeader.tsx
hafidk: [fixed here](https://github.com/AdaptiveConsulting/ReactiveTraderCloud/pull/1970/commits/bebfc79ca887b2478f776dfaba36f2ce1f214d5c)

src/new-client/src/App/LiveRates/Tile/HistoricalGraph/HistoricalGraph.tsx
hafidk: [fixed here](https://github.com/AdaptiveConsulting/ReactiveTraderCloud/pull/1970/commits/bebfc79ca887b2478f776dfaba36f2ce1f214d5c)

src/new-client/src/App/LiveRates/Tile/HistoricalGraph/HistoricalGraph.tsx
hafidk: [Fixed here](https://github.com/AdaptiveConsulting/ReactiveTraderCloud/pull/1970/commits/bebfc79ca887b2478f776dfaba36f2ce1f214d5c), for now I followed the structure we talked, but in this specific case it was different since the styleguide adds a dotted baseline svg, the `isAtom` prop was necesary in order to determine the length of this line (although that could be optimized too), for now I left the old version commented in case we need to get back there and we render the same as the original.


Screenshot 2021-11-04 at 08 42 00

src/new-client/src/App/LiveRates/Tile/Tile.tsx
hafidk: [Fixed here](https://github.com/AdaptiveConsulting/ReactiveTraderCloud/pull/1970/commits/bebfc79ca887b2478f776dfaba36f2ce1f214d5c), I added this data in the `SpotTilesMockData` file so it doesn't polute the `Tile.tsx`

src/new-client/src/App/LiveRates/Tile/Header/TileHeader.tsx
```diff
@@ -16,11 +16,11 @@ export const DeliveryDate = styled.div`
   margin-left: auto;
   transition: margin-right 0.2s;
 `
-const HeaderWrapper = styled.div<{ supportsTearOut?: boolean }>`
+const HeaderWrapper = styled.div<{ tearOutMock?: boolean }>`
   display: flex;
   align-items: center;
   position: relative;
-  margin-right: ${({ supportsTearOut }) => (supportsTearOut ? "1.3rem" : "")};
+  margin-right: ${({ tearOutMock }) => (tearOutMock ? "1.3rem" : "")};
```
taylor-adaptive, line 23: Can this name stay as `supportsTearOut`? It's more descriptive.

src/new-client/src/App/LiveRates/Tile/Header/TileHeader.tsx
```diff
@@ -64,40 +64,50 @@ export const Header: React.FC = () => {
...
       </TileSymbol>
       <DeliveryDate data-qa="tile-header__delivery-date">{date}</DeliveryDate>
-      {canTearOut && (
-        <HeaderAction
-          onClick={() => {
-            tearOut(symbol, !isTornOut)
-          }}
-        >
+      {canTearOut && !tearOutMock && (
+        <HeaderAction onClick={onClick ? onClick : () => {}}>
```
taylor-adaptive, line 106: Minor - Better practice is to pass undefined in cases of no callback to `onX` handlers.
`onClick={onClick || undefined}` or simply `onClick={onClick}` since `onClick` is an optional prop (it will default to `undefined`)

src/new-client/src/App/LiveRates/Tile/Header/TileHeader.tsx
```diff
@@ -64,40 +64,50 @@ export const Header: React.FC = () => {
...
-  return (
-    <HeaderWrapper supportsTearOut={supportsTearOut}>
-      <TileSymbol data-qa="tile-header__tile-symbol">
-        {currencyPair.base}/{currencyPair.terms}
-      </TileSymbol>
-      <DeliveryDate data-qa="tile-header__delivery-date">
-        {dateVal}
-      </DeliveryDate>
-      {supportsTearOut && (
+      {tearOutMock && (
```
taylor-adaptive, line 110: Can we remove this duplication so we only have 1 set of icons in the render function? 
You should be able to achieve the same styling if you pass `shouldMoveDate={true}` when `hover` is true in the `TileMockComponent`. That will add the styles to the `DeliveryDate` meaning the absolute positioning should be correct for the pop out icon

src/new-client/src/App/LiveRates/Tile/HistoricalGraph/HistoricalGraph.tsx
```diff
@@ -83,73 +83,82 @@ export const HistoricalGraph: React.FC<HistoricalGraphProps> = ({
...
-              y2="40"
-            ></line>
-          </g>
-        )}
       </Svg>
     </LineChartWrapper>
   )
 }
+
+// export const HistoricalGraphComponent: React.FC<{
```
taylor-adaptive, line 110: Good job in here.
As discussed on slack, there is an issue in this component when running the app in analytics mode.
Once fixed, could you remove this commented block?

src/new-client/src/App/LiveRates/Tile/HistoricalGraph/HistoricalGraph.tsx
taylor-adaptive: Ahh, understood. 
Maybe add the line back in but give the prop a more descriptive name so it is not tied to the context of styleguide.
Something like `showCenterLine` or similar.

src/new-client/src/styleguide/components/ChartingGrid.tsx
```diff
@@ -79,7 +79,7 @@ const ChartsVariants: React.FC<{ active?: boolean }> = ({ active = false }) => (
             history={HistoryMockSvgPath}
             showTimer={true}
             active={active}
-            isAtom={true}
+            // isAtom={true}
```
taylor-adaptive, line 82: As above, lets add back in but with a different name

src/new-client/src/App/LiveRates/Tile/Header/TileHeader.tsx
```diff
@@ -86,32 +85,27 @@ export const HeaderComponent: React.FC<{
...
   return (
-    <HeaderWrapper tearOutMock={tearOutMock}>
+    <HeaderWrapper supportsTearOut={supportsTearOut}>
       <TileSymbol data-qa="tile-header__tile-symbol">
         {base}/{terms}
       </TileSymbol>
       <DeliveryDate data-qa="tile-header__delivery-date">{date}</DeliveryDate>
-      {canTearOut && !tearOutMock && (
-        <HeaderAction onClick={onClick ? onClick : () => {}}>
+      {canTearOut && (
```
taylor-adaptive, line 97: This isn't visible in the styleguide for the hover state unless we hover over the tile. Sorry if I lead you down the wrong path in earlier comments.
I suggest adding an optional `hover` prop that sets the opacity so the styleguide can show it without manually hovering.

taylor-adaptive:
Closing in favour of updated PR https://github.com/AdaptiveConsulting/ReactiveTraderCloud/pull/2020

PR #1980, opened 2021-11-04
feat(new-client): NLP Cloud function

taylor-adaptive:
Created and deployed a cloud function to replace the current broker/messaging solution for the nlp functionality in `new-client`.


rikoe (approved):
This is great!

PR #1992, opened 2021-11-11
feat (new-client): torn out sections (1596)

hafidk:
+ Added torn out to each section
+ Once two sections are torn out, you can't torn the third one (like classic)
+ Torn out sections keep a similar aspect ratio to their originals

Pending:
- Code reviews
- Opening the torn out window in the same coordinates as the section


src/new-client/src/Web/tearoutSections.ts
```diff
@@ -0,0 +1,47 @@
+import { bind } from "@react-rxjs/core"
+import { createSignal } from "@react-rxjs/utils"
+import { map } from "rxjs/operators"
+
+type TearOutEntry = [boolean, HTMLDivElement?]
+
+export const [tearOutEntryTiles$, tearOutTiles] = createSignal(
```
taylor-adaptive, line 7: Perhaps a better way to do this would be to have 1 signal for all the tear out windows and use a `type` to be able to distinguish what window it is. That way we can reduce the duplicated signals.
You could still bind this to create different hooks (useTearOutTileState, useTearOutAnalyticsState etc..) or just bind to one hook, whichever feels best.

src/new-client/src/App/Trades/TradesHeader/TradesHeader.tsx
```diff
@@ -30,17 +47,49 @@ const HeaderToolbar = styled("div")`
...
+          <AppliedFilters />
+          <QuickFilter />
+        </HeaderToolbar>
+        {!isTornOut && !(tearOutTileState && tearOutAnalyticsState) && (
+          <HeaderAction
+            onClick={() => {
+              tearOutTrades(true)
+            }}
+          >
+            <PopOutIcon />
```
taylor-adaptive, line 88: Maybe still on your list to do, but in case you overlooked - Need to change this to `PopInIcon` when it is popped out

src/new-client/src/App/Trades/TradesHeader/TradesHeader.tsx
```diff
@@ -30,17 +47,49 @@ const HeaderToolbar = styled("div")`
...
-)
+var isTornOut = false
+export const TradesHeader: React.FC = () => {
+  const tearOutTileState = useTearOutTileState("Tiles")
+  const tearOutAnalyticsState = useTearOutAnalyticsState("Analytics")
+  useObservableSubscription(
+    tearOutEntryTrades$.subscribe(async ([tornOut]) => {
+      if (tornOut && !isTornOut) {
+        isTornOut = true
+        openWindow(
```
taylor-adaptive, line 58: In common App components we shouldn't use any platform specific stuff (web, openfin ec..)
Here you should use `openWindow` from `utils/window`.

src/new-client/src/App/Trades/TradesHeader/TradesHeader.tsx
```diff
@@ -30,17 +47,49 @@ const HeaderToolbar = styled("div")`
...
+  return (
+    <TradesHeaderStyle>
+      <HeaderLeftGroup>Trades</HeaderLeftGroup>
+      <HeaderRightGroup>
+        <ExcelButton />
+        <HeaderToolbar>
+          <AppliedFilters />
+          <QuickFilter />
+        </HeaderToolbar>
+        {!isTornOut && !(tearOutTileState && tearOutAnalyticsState) && (
```
taylor-adaptive, line 82: We need to be able to not show this when in OpenFin (as it handles popout layout automatically).
Options include - 
1. Adding an optional prop to the parent `Trades` component where web can say 'I support tear outs' and openfin says 'I don't support tear outs'
2. Create a file that determines if the component supports tear out and let the vite build handle the import replacement

Number 2 would look something like
```
// /App/Trades/supportsTearOut.ts 
export const supportsTearOut = true
```

```
// /App/Trades/supportsTearOut.openfin.ts 
export const supportsTearOut = false
```

```
import { supportsTearout } from './supportsTearOut'
// supportsTearout will be true when compiled for web, false when compiled for openfin
```

src/new-client/src/App/Trades/TradesHeader/TradesHeader.tsx
```diff
@@ -30,17 +47,49 @@ const HeaderToolbar = styled("div")`
...
-      </HeaderToolbar>
-      <Fill />
-    </HeaderRightGroup>
-  </TradesHeaderStyle>
-)
+var isTornOut = false
+export const TradesHeader: React.FC = () => {
+  const tearOutTileState = useTearOutTileState("Tiles")
+  const tearOutAnalyticsState = useTearOutAnalyticsState("Analytics")
+  useObservableSubscription(
```
taylor-adaptive, line 54: I think you may be working off an old version of master, this util function no longer exists.

src/new-client/src/App/Trades/TradesHeader/TradesHeader.tsx
```diff
@@ -30,17 +47,49 @@ const HeaderToolbar = styled("div")`
...
+    tearOutEntryTrades$.subscribe(async ([tornOut]) => {
```
taylor-adaptive, line 55: Rather than subscribing here you can use the hook created by `bind`. It will give you access to the latest value for that stream.

src/new-client/src/App/LiveRates/styled.tsx
```diff
@@ -153,3 +153,11 @@ export const DropdownOption = styled.div<{ selected: boolean }>`
     padding-left: 20px;
   }
 `
+
+export const HeaderAction = styled.button`
```
taylor-adaptive, line 157: This is duplicated across components, a good candidate to move out into a generic component

src/new-client/src/App/Analytics/AnalyticsCore.tsx
```diff
@@ -7,20 +7,71 @@ import {
...
+  useTearOutTradeState,
+} from "@/Web/tearoutSections"
+
 const analytics$ = merge(pnL$, profitAndLoss$, positions$)
 
 const SuspenseOnStaleData = createSuspenseOnStale(isAnalyticsDataStale$)
 
+var isTornOut = false
 const Analytics: React.FC = ({ children }) => {
+  const tearOutTradeState = useTearOutTradeState("Trade")
```
taylor-adaptive, line 34: Similar logic is used in the other components. Perhaps a util function somewhere would be better.
eg:
const canBeTornOut = (windowToTearOut) => yes or no based on the state of tear outs

taylor-adaptive (commented):
Good start.
The comments I left for Trades are true for the LiveRates and Analytics too.

src/new-client/src/App/Analytics/AnalyticsCore.tsx
```diff
@@ -7,10 +7,12 @@ import {
   AnalyticsStyle,
   AnalyticsHeader,
   AnalyticsInnerWrapper,
+  RightNav,
 } from "./styled"
 import { createSuspenseOnStale } from "@/utils/createSuspenseOnStale"
 import { isAnalyticsDataStale$ } from "@/services/analytics"
-
+import { supportsTearOut } from "@/Web/TearOutSection/supportsTearOut"
```
taylor-adaptive, line 14: This shouldn't live in `@Web` as it has different platform implementations.

src/new-client/src/Web/TearOutSection/handleTearOutSection.openfin.ts
```diff
@@ -0,0 +1,39 @@
+// import { calculateWindowCoordinates, Offset } from "@/utils"
```
taylor-adaptive, line 1: Don't think this file is needed?

src/new-client/src/Web/TearOutSection/handleTearOutSection.web.ts
```diff
@@ -0,0 +1,29 @@
+import { constructUrl } from "@/utils/url"
```
taylor-adaptive, line 1: This can go in `handleTearOutSection.ts` as we don't need any other platform implementations at the moment.

src/new-client/src/Web/TearOutSection/handleTearOutSection.web.ts
```diff
@@ -0,0 +1,29 @@
...
+import { openWindow } from "@/utils/window/openWindow"
+import { tearOutSection } from "./state"
+
+export function handleTearOutSection(section: string) {
```
taylor-adaptive, line 5: We should create a type for section throughout this functionality so we remove the ability to use a string with a typo.

src/new-client/src/Web/TearOutSection/handleTearOutSection.web.ts
```diff
@@ -0,0 +1,29 @@
...
+      url: constructUrl(`/${section}`),
+      name: section,
+      width: width,
+      height: height,
+    },
+    () => tearOutSection(false, section),
+  )
+}
+
+export const supportsTearOut = true
```
taylor-adaptive, line 29: Is this used?

src/new-client/src/Web/TearOutSection/handleTearOutSection.web.ts
```diff
@@ -0,0 +1,29 @@
+import { constructUrl } from "@/utils/url"
+import { openWindow } from "@/utils/window/openWindow"
+import { tearOutSection } from "./state"
+
+export function handleTearOutSection(section: string) {
+  let width = 380
+  let height = 380
+  if (section == "liverates") {
```
taylor-adaptive, line 8: An alternative would be add this config to an object and use section as the key, to remove if/else/else etc..

Example:
```
const sectionConfig = {
  liverates: {
    width: 1600,
    height: 800
  }
  ...
}

const { width, height } = sectionConfig[section]
```
and this can be enhanced with types once you add a `section` type

src/new-client/src/Web/WebApp.tsx
```diff
@@ -29,6 +37,36 @@ export const WebApp: React.FC = () => (
           )
         }}
       />
+      <Route
+        path="/liverates"
+        render={() => {
+          return (
+            <TornOutTileContext.Provider value={{ isTornOut: true }}>
+              <LiveRates />{" "}
```
taylor-adaptive, line 45: Minor - Whitespace

src/new-client/src/components/styled.tsx
```diff
@@ -0,0 +1,9 @@
+import styled from "styled-components"
+
+export const HeaderAction = styled.button`
```
taylor-adaptive, line 3: This should have a more descriptive filename

src/new-client/src/Web/context.tsx
```diff
@@ -0,0 +1,4 @@
+import { createContext } from "react"
+export const TornOutTileContext = createContext({ isTornOut: false })
```
taylor-adaptive, line 2: This context doesn't belong in the Web folder as it is used outside of the web scope.

taylor-adaptive, line 2: Perhaps a more generic context could be used so we remove the duplication - `SectionTornOutContext` or similar

src/new-client/src/Web/WebApp.tsx
```diff
@@ -29,6 +37,36 @@ export const WebApp: React.FC = () => (
...
+            </TornOutTileContext.Provider>
+          )
+        }}
+      />
+      <Route
+        path="/trades"
+        render={() => {
+          return (
+            <TornOutTradeContext.Provider value={{ isTornOut: true }}>
```
taylor-adaptive, line 54: See comment about single context. If you use a single context then you could wrap all these routes as children of the context, reducing the duplicaton.

src/new-client/src/Web/TearOutSection/TearOutComponent.tsx
```diff
@@ -0,0 +1,35 @@
...
+import {
+  TornOutTileContext,
+  TornOutTradeContext,
+  TornOutAnalyticsContext,
+} from "@/Web/context"
+
+export const TearOutComponent: React.FC<{ section: string }> = ({
+  section,
+}) => {
+  var context = TornOutTileContext
```
taylor-adaptive, line 14: Minor - Avoid the usage of `var`.
`const` for everything unless you're going to reassign the value - `let`

src/new-client/src/Web/TearOutSection/TearOutComponent.tsx
```diff
@@ -0,0 +1,35 @@
...
+  if (section === "trades") {
```
taylor-adaptive, line 15: See comment earlier about single context

src/new-client/src/Web/TearOutSection/TearOutComponent.tsx
```diff
@@ -0,0 +1,35 @@
...
+    context = TornOutTradeContext
+  } else {
+    context = TornOutAnalyticsContext
+  }
+  return (
+    <context.Consumer>
```
taylor-adaptive, line 21: You can use the `useContext` hook, it's much nicer to work with than this api.
https://reactjs.org/docs/hooks-reference.html#usecontext

src/new-client/src/Web/MainRoute.tsx
```diff
@@ -36,19 +44,35 @@ const MainWrapper = styled.div`
...
-      </MainWrapper>
-      <Footer />
-    </AppLayoutRoot>
-  </Wrapper>
-)
+export const MainRoute: React.FC = () => {
+  const tearOutTileState = useTearOutSectionState$("liverates")
+  const tearOutAnalyticsState = useTearOutSectionState$("analytics")
+  const tearOutTradeState = useTearOutSectionState$("trades")
+  const tearOutEntry = useTearOutSectionEntry()
```
taylor-adaptive, line 51: This hook gives us access to all the data w need.
Rather than creating 4 subscriptions in total, we could use this and maintain the tearOut state inside this component.
In other words - listen to the `tearOutEntry` and add/remove the section from some local state, and use that to conditionally render the sections

src/new-client/src/App/Trades/Trades.tsx
```diff
@@ -14,7 +14,7 @@ const TradesWrapper = styled.article`
 `
 
 trades$.subscribe()
-export const Trades: React.FC = () => (
+export const Trades: React.FC = (tornOut) => (
```
taylor-adaptive, line 17: ??

hafidk: Fixed.

src/new-client/src/Web/MainRoute.tsx
hafidk: Deleted the other three hooks, now we have a single state, I don't know if we prefer 3 different states for each section or a single state with all 3 keys (the latter makes updating the state simplier at first glance):

`
...
const [tornOutSectionState, setTornOutSectionState] = useState({
    liverates: false,
    trades: false,
    analytics: false,
  })

  useEffect(() => {
    if (tearOutEntry) {
      const [tornOut, section] = tearOutEntry
      setTornOutSectionState({ ...tornOutSectionState, [section]: tornOut })
      ...
`

src/new-client/src/Web/TearOutSection/TearOutComponent.tsx
hafidk: Done

src/new-client/src/Web/TearOutSection/TearOutComponent.tsx
hafidk: Fixed

src/new-client/src/Web/TearOutSection/TearOutComponent.tsx
hafidk: Done.

src/new-client/src/Web/WebApp.tsx
hafidk: Done.

src/new-client/src/Web/context.tsx
hafidk: Fixed, you can find the implementation at `@/components/tearOutContext`

hafidk: Done, I didn't know where to put it so for now it's at `@/components/tearOutContext`  (I also changed the file name since it was misleading)

src/new-client/src/components/styled.tsx
hafidk: Fixed, it's `headerActionStyles` now

src/new-client/src/Web/WebApp.tsx
hafidk: Fixed

src/new-client/src/Web/TearOutSection/handleTearOutSection.web.ts
hafidk: Fixed, did it as you said.

src/new-client/src/Web/TearOutSection/handleTearOutSection.web.ts
hafidk: Fixed, it was replicated with the other `supportsTearOut` within `@/App/supportsTearOut`

src/new-client/src/Web/TearOutSection/handleTearOutSection.web.ts
hafidk: Done, for now it's `type Section = "liverates" | "trades" | "analytics"`

src/new-client/src/Web/TearOutSection/handleTearOutSection.web.ts
hafidk: I'm not sure I follow, `handleTearOutSection.web.ts` uses `constructUrl` 

`
export function handleTearOut(symbol: string) {
  openWindow(
    {
      url: constructUrl(`/spot/${symbol}`),
`

If I move this to the generic handleTearOutSection.ts wouldn't I get a warning?

src/new-client/src/Web/TearOutSection/handleTearOutSection.openfin.ts
hafidk: Fixed, the file is no more.

src/new-client/src/App/Analytics/AnalyticsCore.tsx
hafidk: Fixed, I didn't know where to put it so for now you can find it  at `@/App/supportsTearOut`

src/new-client/src/Web/TearOutSection/TearOutComponent.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+export const TearOutComponent: React.FC<{ section: string }> = ({
+  section,
+}) => {
+  const tearOutContext = useContext(TearOutContext)
+  return (
+    <HeaderTearOutAction
+      onClick={() =>
+        tearOutContext.isTornOut
+          ? window.close()
```
taylor-adaptive, line 16: `utils/window/closeWindow`

src/new-client/src/Web/TearOutSection/handleTearOutSection.web.ts
taylor-adaptive: What I mean to say is - we don't have any other implementations for handleTearOutSection so we don't need to have a `.web.ts` version, this logic can be moved to `handleTearOutSection.ts`

src/new-client/src/Web/TearOutSection/handleTearOutSection.web.ts
```diff
@@ -0,0 +1,33 @@
+import { constructUrl } from "@/utils/url"
+import { openWindow } from "@/utils/window/openWindow"
+import { tearOutSection } from "./state"
+
+type Section = "liverates" | "trades" | "analytics"
```
taylor-adaptive, line 5: Suggest moving this to `TearOutSection/state.ts` and using it everywhere in the app that references the sections (rather than just a string)

src/new-client/src/Web/TearOutSection/state.ts
```diff
@@ -0,0 +1,14 @@
+import { bind } from "@react-rxjs/core"
+import { createSignal } from "@react-rxjs/utils"
+import { map, filter } from "rxjs/operators"
+
+type TearOutSectionEntry = [boolean, string]
```
taylor-adaptive, line 5: As above, this string should use the Section type

src/new-client/src/components/tearOutContext.tsx
```diff
@@ -0,0 +1,15 @@
...
+  },
+  analytics: {
+    isTornOut: false,
+  },
+  trades: {
+    isTornOut: false,
+  },
+}
+
+export const TearOutContext = createContext(sections.liverates)
```
taylor-adaptive, line 15: Shouldn't this just be `sections`?

src/new-client/src/components/tearOutContext.tsx
```diff
@@ -0,0 +1,15 @@
+import { createContext } from "react"
```
taylor-adaptive, line 1: components probably isn't the place for this context file, I would put it alongside the tear out state file

src/new-client/src/Web/TearOutSection/TearOutComponent.tsx
hafidk: Done

src/new-client/src/Web/TearOutSection/handleTearOutSection.web.ts
hafidk: Done

src/new-client/src/Web/TearOutSection/state.ts
hafidk: Done

src/new-client/src/components/tearOutContext.tsx
hafidk: I simplified it a bit now

src/new-client/src/components/tearOutContext.tsx
hafidk: I ended up putting it in` @/App/TearoutSection `because the state file for the tear out sections was within `@/Web`

src/new-client/src/App/TearOutSection/tearOutContext.tsx
```diff
@@ -0,0 +1,7 @@
+import { createContext } from "react"
+
+export const sectionsTornOut = {
+  isTornOut: false,
+}
+
+export const TearOutContext = createContext(sectionsTornOut)
```
taylor-adaptive, line 7: Can you add an explicit type here, it's a good practice to get into.

src/new-client/src/Web/MainRoute.tsx
```diff
@@ -36,19 +41,38 @@ const MainWrapper = styled.div`
...
-        </Resizer>
-        <Analytics />
-      </MainWrapper>
-      <Footer />
-    </AppLayoutRoot>
-  </Wrapper>
-)
+export const MainRoute: React.FC = () => {
+  const tearOutEntry = useTearOutSectionEntry()
+  const [tornOutSectionState, setTornOutSectionState] = useState({
```
taylor-adaptive, line 46: Same here, good practice to supply a type

src/new-client/src/Web/TearOutSection/state.ts
```diff
@@ -0,0 +1,31 @@
...
+export const [tearOutSectionEntry$, tearOutSection] = createSignal(
+  (bool: boolean, section: Section): TearOutSectionEntry => [bool, section],
+)
+
+export const [useTearOutSectionEntry] = bind<TearOutSectionEntry | null>(
+  tearOutSectionEntry$,
+  null,
+)
+
+export const sectionConfig = {
```
taylor-adaptive, line 18: Explicit type here too

src/new-client/src/App/Trades/TradesHeader/TradesHeader.tsx
```diff
@@ -2,6 +2,8 @@ import styled from "styled-components"
 import { ExcelButton } from "./ExcelButton"
 import { AppliedFilters } from "./AppliedFilters"
 import { QuickFilter } from "./QuickFilter"
+import { supportsTearOut } from "@/App/supportsTearOut"
+import { TearOutComponent } from "@/Web/TearOutSection/TearOutComponent"
```
taylor-adaptive, line 6: Can we move all of the `TearOutSection` stuff from web into `App/TearOutSection` please, that way no imports to web specific stuff is happening from `@/App` and will allow us to extend if our other platforms will support tear out (finsemble, glue42 etc..)

src/new-client/src/Web/TearOutSection/state.ts
hafidk: Done, defined these types and certain consts have explicitly these
```
type tornOutSize = { width: number, height: number}

export type Section = "liverates" | "trades" | "analytics"

export type SectionConfig = { [section in Section]: tornOutSize }

export type TornOutSection = { [section in Section]: boolean }

File: src/new-client/src/Web/MainRoute.tsx (hunk 30 above)
Comment (hafidk) [reply src/new-client/src/Web/MainRoute.tsx]: Done

File: src/new-client/src/App/TearOutSection/tearOutContext.tsx (hunk 29 above)
Comment (hafidk) [reply src/new-client/src/App/TearOutSection/tearOutContext.tsx:11]: Done

File: src/new-client/src/App/Trades/TradesHeader/TradesHeader.tsx (hunk 32 above)
Comment (hafidk) [reply src/new-client/src/App/Trades/TradesHeader/TradesHeader.tsx]: Moved

File: src/new-client/src/App/TearOutSection/tearOutContext.tsx [hunk 33]
```diff
@@ -0,0 +1,11 @@
+import { createContext } from "react"
+
+type sectionTornOut = {
```
taylor-adaptive, line 3: Minor - Suggest a better name - `TearOutContextType` or similar

src/new-client/src/App/TearOutSection/tearOutContext.tsx
```diff
@@ -0,0 +1,11 @@
...
+  isTornOut: boolean
+}
+
+export const sectionsTornOut: sectionTornOut = {
+  isTornOut: false,
+}
+
+export const TearOutContext = createContext(sectionsTornOut)
```
taylor-adaptive, line 11: Minor - Something more descriptive and explicit -
`createContext<TearOutContextType>(initialTearOutContext)`

src/new-client/src/App/supportsTearOut.openfin.ts
```diff
@@ -0,0 +1 @@
+export const supportsTearOut = false
```
taylor-adaptive, line 1: Should be in same location as supportsTearOut.ts

src/new-client/src/components/headerActionStyles.tsx
```diff
@@ -0,0 +1,9 @@
+import styled from "styled-components"
+
+export const HeaderTearOutAction = styled.button`
```
taylor-adaptive, line 3: If this is only used by the `TearOutComponent` then a better place would be in `TearOutComponent.styles.ts` in the same folder

taylor-adaptive (approved):
Great job

PR #1999, opened 2021-11-19
feat(new-client): draggable component

hafidk:


src/new-client/src/components/DraggableComponent/DraggableComponent.tsx
```diff
@@ -0,0 +1,87 @@
...
+interface DraggableComponentProps {
+  children: JSX.Element[] | JSX.Element
+  section: Section
+  isTile?: boolean
+  ref?: RefObject<HTMLDivElement>
+  isTornOut?: boolean
+  symbol?: string
+}
+
+export const DraggableComponent: React.FC<DraggableComponentProps> = (
```
taylor-adaptive, line 17: Minor - `DraggableTearOut` or something similar would be more descriptive.

src/new-client/src/components/DraggableComponent/DraggableComponent.tsx
```diff
@@ -0,0 +1,87 @@
...
+  props: DraggableComponentProps,
+) => {
+  //stopPropagation() is not enough to avoid the parent (in case we are dragging a Tile) also tearing out
+  //this state makes sure that the drag event started and finished succesfully
+  const [finishedDrag, setFinishedDrag] = useState(false)
+
+  const tearOutContext = useContext(TearOutContext)
+  const symbol = props.symbol ? props.symbol : ""
+  const ref = props.ref ? props.ref : { current: null }
```
taylor-adaptive, line 26: Minor - You should be able to use `props.ref!` and `props.symbol!` below without having to conditionally assign here.

src/new-client/src/components/DraggableComponent/DraggableComponent.tsx
```diff
@@ -0,0 +1,87 @@
...
+import { tearOutSection, Section } from "@/App/TearOutSection/state"
+import { canDrag } from "@/components/DraggableComponent/canDrag"
+import { tearOut } from "@/App/LiveRates/Tile/TearOut/state"
+
+interface DraggableComponentProps {
+  children: JSX.Element[] | JSX.Element
+  section: Section
+  isTile?: boolean
+  ref?: RefObject<HTMLDivElement>
+  isTornOut?: boolean
```
taylor-adaptive, line 13: I'm not sure we need this. Classic doesn't let you drag things to pop them back into the application, I think this is intentional.
Perhaps we should have a 'disabled' prop here (which we would use `isTornOut` from the parent component to determine if it's disabled or not). 
EG:
`draggable={canDrag && !disabled}`

src/new-client/src/components/DraggableComponent/DraggableComponent.tsx
```diff
@@ -0,0 +1,87 @@
...
+  const [finishedDrag, setFinishedDrag] = useState(false)
+
+  const tearOutContext = useContext(TearOutContext)
+  const symbol = props.symbol ? props.symbol : ""
+  const ref = props.ref ? props.ref : { current: null }
+
+  const eventHandler = (event: React.DragEvent) => {
+    props.isTile
+      ? tearOut(symbol, !props.isTornOut, ref.current!)
+      : tearOutSection(!tearOutContext.isTornOut, props.section)
```
taylor-adaptive, line 31: As above, I think this could just be `tearOutSection(true, props.section)` without the need for dragging back in (same goes for `tearOut` above)

src/new-client/src/components/DraggableComponent/DraggableComponent.tsx
hafidk: Partially fixed, the only thing I'm not quite happy with the disabled prop, when we use the `draggableTearOut` for the sections, we don't know if it's torn out or not there (hence we would have to import the context function, generate the context and pass the prop), since we would have to do that x3 times, I don't know if it's preferable to contain that logic within  the `DraggableComponent`.

So, in this case we determine if its a Tile, in that case we can pass the prop without issues since the parent has that value, with the sections, in order to pass the prop we would have to drill it from the grandparent (with the components being lazyloaded in between) so we just generate the contest in `DraggableTearOut` and use it from there.

`const disabled = props.isTile ? props.disabled! : tearOutContext.isTornOut`

src/new-client/src/components/DraggableComponent/DraggableComponent.tsx
hafidk: Fixed

src/new-client/src/components/DraggableTearOut/DraggableTearOut.tsx
```diff
@@ -0,0 +1,84 @@
...
+interface DraggableTearOutProps {
+  children: JSX.Element[] | JSX.Element
+  section: Section
+  isTile?: boolean
+  ref?: RefObject<HTMLDivElement>
+  symbol?: string
+  disabled?: boolean
+}
+
+export const DraggableTearOut: React.FC<DraggableTearOutProps> = (
```
taylor-adaptive, line 20: Some further thoughts - 
I'm thinking it might be nice to have 2 different components - `DraggableSectionTearOut` and `DraggableTileTearOut`  because at the moment there is nothing stopping you saying `isTile: true` without passing a `symbol`, and `section` is mandatory when it is not needed when used by tiles.
They can both use the same `DraggableTearOut` component but have different Props, and `DraggableTearOut` can be simplified.

EG:
```
type DraggableSectionTearOutProps = React.PropsWithChildren<{
  section: Section;
  disabled?: boolean;
}>

type DraggableTileTearOutProps = React.PropsWithChildren<{
  symbol: string;
  tileRef: RefObject<HTMLDivElement>;
  disabled? boolean;
}>

type DraggableTearOutProps = React.PropsWithChildren<{
  dragHandler: () => void;
  // could pass disabled down here, but I think it might be better to just return the children in the parent component instead of rendering DraggableTearOut at all
}>
```

src/new-client/src/App/Analytics/AnalyticsCore.tsx
```diff
@@ -21,23 +23,25 @@ const Analytics: React.FC = ({ children }) => {
...
-          role="region"
-          aria-label="Trade and position analytics"
-          data-qa="analytics__analytics-content"
-        >
-          <ProfitAndLoss />
-          <Positions />
-          <PnL />
-        </AnalyticsStyle>
-      </AnalyticsInnerWrapper>
+      <DraggableTearOut section="analytics">
```
taylor-adaptive, line 26: Another thought - Since these sections are only tearout'able from the web context, maybe we should wrap them in `DraggableSectionTearOut` in `web/MainRoute.tsx`.
That would keep the section components free from any draggable parent component which isn't needed when in openfin (or when rendering standalone in styleguide/storybook).

src/new-client/src/components/DraggableTearOut/DraggableTearOut.tsx
hafidk: Fixed at `DraggableTearout.tsx`

src/new-client/src/App/Analytics/AnalyticsCore.tsx
hafidk: fixed at `MainRoute.tsx`

src/new-client/src/Web/MainRoute.tsx
```diff
@@ -70,10 +70,31 @@ export const MainRoute: React.FC = () => {
         <Header />
         <MainWrapper>
           <Resizer defaultHeight={30}>
-            {!tornOutSectionState.liverates && <LiveRates />}
-            {!tornOutSectionState.trades && <Trades />}
+            {!tornOutSectionState.liverates && (
+              <DraggableSectionTearOut
+                section="liverates"
+                disabled={tornOutSectionState.liverates}
```
taylor-adaptive, line 76: The disabled prop aren't required here. 
If `tornOutSectionState.liverates` is true, it wouldn't be rendered in the first place.

src/new-client/src/components/DraggableTearOut/DraggableTearOut.tsx
```diff
@@ -0,0 +1,120 @@
...
+  tileRef: RefObject<HTMLDivElement>
+  disabled?: boolean
+}>
+
+type DraggableTearOutGenericProps = React.PropsWithChildren<{
+  dragHandler: () => void
+}>
+
+export const DraggableSectionTearOut: React.FC<DraggableSectionTearOutProps> = (
+  props: DraggableSectionTearOutProps,
```
taylor-adaptive, line 27: Just a minor style point - 
If you define a component's props with `React.FC<Props>` then you do not need to type the props argument `(props: Props)`.

src/new-client/src/components/DraggableTearOut/DraggableTearOut.tsx
```diff
@@ -0,0 +1,120 @@
...
+) => {
+  const draggable = canDrag && !props.disabled
+  const dragHandler = () => {
+    tearOutSection(true, props.section)
+  }
+  return (
+    <>
```
taylor-adaptive, line 34: Another really minor style point - If the top level of a render function is javascript then you don't need to provide the empty `<></>` tags

src/new-client/src/components/DraggableTearOut/DraggableTearOut.tsx
```diff
@@ -0,0 +1,120 @@
...
+    </>
+  )
+}
+
+export const DraggableTileTearOut: React.FC<DraggableTileTearOutProps> = (
+  props: DraggableTileTearOutProps,
+) => {
+  const draggable = canDrag && !props.disabled
+  const dragHandler = () => {
+    tearOut(props.symbol!, true, props.tileRef?.current!)
```
taylor-adaptive, line 51: minor - symbol and `tileRef` are mandatory props, so no need for `!` and `?`

src/new-client/src/Web/MainRoute.tsx
hafidk: Fixed, removed the `disabled` prop

src/new-client/src/components/DraggableTearOut/DraggableTearOut.tsx
hafidk: Fixed, now they only receive (props) without the mandatory type

src/new-client/src/components/DraggableTearOut/DraggableTearOut.tsx
hafidk: I'm not sure if I follow, whenever I try to remove the encapsulation of the props.children, a typescript error appears:

```
Type '(props: PropsWithChildren<DraggableSectionTearOutProps>) => ReactNode' is not assignable to type 'FC<DraggableSectionTearOutProps>'.
  Type 'ReactNode' is not assignable to type 'ReactElement<any, any> | null'.
    Type 'undefined' is not assignable to type 'ReactElement<any, any> | null'.ts(2322)
```

for now I moved the encapsulation of empty tags be with the props.children

```
return (
    draggable ? (
        <DraggableTearOut dragHandler={dragHandler}>
          {props.children}
        </DraggableTearOut>
      ) : <>{props.children}</>
  )
```

src/new-client/src/components/DraggableTearOut/DraggableTearOut.tsx
hafidk: Fixed

src/new-client/src/components/DraggableTearOut/DraggableTearOut.tsx
```diff
@@ -0,0 +1,112 @@
...
+  const draggable = canDrag && !props.disabled
+  const dragHandler = () => {
+    tearOutSection(true, props.section)
+  }
+  return draggable ? (
+    <DraggableTearOut dragHandler={dragHandler}>
+      {props.children}
+    </DraggableTearOut>
+  ) : (
+    <>props.children</>
```
taylor-adaptive, line 38: Missing `{}` here, it's just a string at the moment.
Also, can just be `: ( props.children ) `

hafidk: For some reason TS doesn't like this notation, for now I added a `@ts-ignore`, will investigate further.

taylor-adaptive: I couldn't work out why so I checked out locally.
`React.PropsWithChildren` adds children as an optional prop, so TS doesn't like `: ( props.children )` because it could be undefined, which would not satisfy the `React.FC` type.
Options are to turn (trick) it into thinking it's a component with the `<>{props.children}</>` like you had, or to change the props to make children mandatory. 
`<>{props.children}</>` is probably the simplest thing to do.

hafidk: Done!

taylor-adaptive (approved):
Great job, good to merge

PR #2000, opened 2021-11-19
[DNM] feat(symphony): Refactored Symphony bot to connect to Hydra

taylor-adaptive:
**DO NOT MERGE** 

- Replaces old websocket connection with `hydra-platform`
- Removes hardcoded currency pairs in favour of getting them from Hydra
- Removes types and uses generated types instead
- Refactors the application services injection to match the approach in RT (import streams directly in files)
- Minor small refactors/improvements to handlers/messages

image


src/server/node/bot/src/services/rt-services.ts
```diff
@@ -0,0 +1,22 @@
+import { connectToGateway } from '@adaptive/hydra-platform'
+import { noop } from 'rxjs'
+import logger from '../logger'
+
+const { Crypto } = require('@peculiar/webcrypto')
+
+// Global packages used by hydra-platform that aren't available in node by default
+Object.assign(global, {
+  WebSocket: require('ws'),
+  crypto: new Crypto()
```
taylor-adaptive, line 10: Polyfill is needed as `crypto` is used in `hydra-platform` but isn't defined in node. 
Node's version of `crypto` doesn't have the same api as the browser, so using this polyfill instead.
An improvement could be to inject the `crypto` provider into hydra through `connectToGateway`, rather than assigning to global.

algreasley:
this is old now .. could resurrect if Symphony becomes a required feature again, but for now it is dormant

PR #2006, opened 2021-11-26
feat(new-client): virtualized table 1/2

hafidk:
# Description

Added the first part of the virtualization for the trades list in order to reduce load times. For extremely big lists of trades this should do wonders. The implementation uses the library react-virtuoso

Fixes # 

Changes the current trades table to a virtualized table

## Type of change

Please delete options that are not relevant.
- [X] New feature (non-breaking change which adds functionality)
- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## List of changes

One of the main issues with this implementation is the virtuoso itself, it's not an out of the box solution since it encapsulates the list of virtualized divs we give it (in this case, the list of trades). This conflicts with the current table structure and forces plenty of changes to at least get it working, the main ones are as it follows:

TradesGrid changes it's structure quite a bit
Since the table headers and the table rows are two separate entities, it's difficult to match styles, specifically widths, in the colConfigs.js now we work with reative vw instead of fixed px

# Checklist:

- [X] Addapt the react-virtuoso list to the current table
- [X] Adjust styles
- [x] Handle responsive heights
- [x] Handle overflow-x in the table
- [x] Adjust container when there are no trades
- [x] Fix footer component (width doesn't adjust table as it is now)


src/new-client/package.json
```diff
@@ -95,6 +96,8 @@
     "@types/styled-components": "^5.1.9",
     "@typescript-eslint/eslint-plugin": "^4.20.0",
     "@typescript-eslint/parser": "^4.20.0",
+    "@types/react-window": "1.8.5",
```
taylor-adaptive, line 99: I don't see these packages as a dependency. Were they removed, or are these needed as a peer dependency of virtuoso?

src/new-client/src/App/Trades/TradesCore.tsx
```diff
@@ -7,28 +7,48 @@ import { TradesHeader } from "./TradesHeader"
...
   width: 100%;
   color: ${({ theme }) => theme.core.textColor};
   font-size: 0.8125rem;
+  position: relative;
 `
 
 const SuspenseOnStaleData = createSuspenseOnStale(isBlotterDataStale$)
+const Trades: React.FC = () => {
+  const [height, setHeight] = useState(0)
+  const [width, setWidth] = useState(0)
```
taylor-adaptive, line 23: Minor - Add explicit types

src/new-client/src/App/Trades/TradesCore.tsx
```diff
@@ -7,28 +7,48 @@ import { TradesHeader } from "./TradesHeader"
...
+  const ref = useCallback((node) => {
+    if (node) {
+      const resizeObserver = new ResizeObserver((entry) => {
+        const { height, width } = entry[0].contentRect
+        setHeight(height)
+        setWidth(width)
+      })
+      resizeObserver.observe(node)
+    }
```
taylor-adaptive, line 32: Need to add a disconnect/unobserve here

src/new-client/src/App/Trades/TradesGrid/TradesGrid.tsx
```diff
@@ -40,20 +41,21 @@ const highlightBackgroundColor = css`
...
+const TableBodyCell = styled.td<{
+  numeric?: boolean
+  rejected?: boolean
+  relativeWidth: number
+}>`
   text-align: ${({ numeric }) => (numeric ? "right" : "left")};
   padding-right: ${({ numeric }) => (numeric ? "1.6rem;" : "0.1rem;")};
   position: relative;
+  width: ${({ relativeWidth }) => `${relativeWidth}vw`};
+  min-width: ${({ relativeWidth }) => `${relativeWidth * 11.5}px`};
```
taylor-adaptive, line 57: Where does 11.5 come from?

src/new-client/src/App/Trades/TradesGrid/TradesGrid.tsx
```diff
@@ -82,10 +84,32 @@ const StatusIndicatorSpacer = styled.th`
...
+const Spacer = styled.th`
+  min-width: 0.275rem;
+`
+
+const SpacerHeader = styled.th`
+  min-width: 0.275rem;
+  border-bottom: 0.25rem solid ${({ theme }) => theme.core.darkBackground};
+`
+
+export const TradeGridRow: React.FC<any> = ({
```
taylor-adaptive, line 108: Can we add props typing here?

src/new-client/src/App/Trades/TradesGrid/TradesGrid.tsx
```diff
@@ -95,56 +119,83 @@ export const TradesGrid: React.FC = () => {
...
-                highlight={trade.tradeId === highlightedRow}
-                onClick={() => tryBroadcastContext(trade.symbol)}
-              >
-                <StatusIndicator
-                  status={trade.status}
-                  aria-label={trade.status}
+        <VirtuosoTable role="grid" width={width}>
+          <TableHead>
+            <TableHeadRow>
+              {/* <StatusIndicatorSpacer scope="col" aria-label="Trade Status" /> */}
```
taylor-adaptive, line 160: Remove?

src/new-client/src/App/Trades/TradesGrid/TradesGrid.tsx
```diff
@@ -95,56 +119,83 @@ export const TradesGrid: React.FC = () => {
...
+              {colFields.map((field) => (
+                <TableHeadCellContainer key={field} field={field} />
+              ))}
+            </TableHeadRow>
+          </TableHead>
+          {trades.length !== 0 && (
+            //Pending to handle the height in order to
+            //@ts-ignore
+            <Virtuoso
+              ref={ref}
```
taylor-adaptive, line 172: I'm not familiar with virtuoso, but these seems like it might be redundant?

src/new-client/src/App/Trades/TradesGrid/TradesGrid.tsx
```diff
@@ -95,56 +119,83 @@ export const TradesGrid: React.FC = () => {
...
+              //@ts-ignore
```
taylor-adaptive, line 173: Why is this needed?

taylor-adaptive (commented):
Looks good so far. 
In case you didn't notice - there are some console warnings around dom elements.
Let me know when these and the comments are addressed, and the styling matches the previous table and I will take another look.

src/new-client/package.json
hafidk: Forgot to delete the types, they were for the previous iterations while I was trying to get those libraries to work here. Fixed. As for the comments above, fixed warnings and adjusted styles

src/new-client/src/App/Trades/TradesCore.tsx
hafidk: Fixed

src/new-client/src/App/Trades/TradesCore.tsx
hafidk: How can I unobserve a node when this already dissapeared from the dom?

```
  const ref = useCallback((node) => {
    let resizeObserver = new ResizeObserver(() => {})
    if (node) {
      resizeObserver = new ResizeObserver((entry) => {
        const { height, width } = entry[0].contentRect
        setHeight(height)
        setWidth(width)
      })
      resizeObserver.observe(node)
    }else{
      //Can't unobserve the node since it's undefined
      // resizeObserver.unobserve(node)
      //Don't know if the disconnect would do the trick
      // resizeObserver.disconnect()
    }
  }, [])
```

src/new-client/src/App/Trades/TradesGrid/TradesGrid.tsx
hafidk: **Not necesary anymore**, fixed. 

The value originally was 12.5 instead of 11.5, and I took that from the proportional value I had in the colConfig.js, there we had the width values in pixels so I converted the proportion of each value in vw (which maybe wasn't the best approach since the vw will take the entire window and not the parent div), for example, if all columns occupied 1000px and the fist one had a witdth of 100px, taking a 10% of the total, the vw relation would be of 10vw and so on, then we would do the convertion again from vw to px in the case of min-width **luckly this is not necesary anymore**. Also, it went from 11.5 to 12.5 by adjusting the values "By eye" (i.e by trial and error). 

As for the width, I kept this value since it's the only way I found that would allow me to keep similar widths between the header of the table and the cells within the rows.

src/new-client/src/App/Trades/TradesGrid/TradesGrid.tsx
hafidk: done

src/new-client/src/App/Trades/TradesGrid/TradesGrid.tsx
hafidk: fixed

src/new-client/src/App/Trades/TradesGrid/TradesGrid.tsx
hafidk: It is as for now, fixed.

src/new-client/src/App/Trades/TradesGrid/TradesGrid.tsx
hafidk: It was during development to make it work since the props were different, not necesary now, fixed.

src/new-client/src/App/Trades/TradesCore.tsx
taylor-adaptive: disconnect should work
https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver/disconnect

src/new-client/src/App/Trades/TradesFooter.tsx
```diff
@@ -14,6 +14,11 @@ const TradesFooterStyled = styled("div")`
   color: ${({ theme }) => theme.core.textColor};
   background-color: ${({ theme }) => theme.core.lightBackground};
   border-radius: 0 0 0.25rem 0.25rem;
+  margin-top: 5px;
```
taylor-adaptive, line 17: Really minor - think it makes sense to give top/bottom/left/right values for position absolute elements, rather than margins.

src/new-client/src/App/Trades/TradesGrid/TradesGrid.tsx
taylor-adaptive: I'm confused - the ref is needed?

src/new-client/src/App/Trades/TradesGrid/TradesGrid.tsx
```diff
@@ -96,56 +43,86 @@ export const TradesGrid: React.FC = () => {
...
-              </TableBodyCell>
-            </TableBodyRow>
+              </TableHeadRow>
+            </TableHead>
+          </Table>
+          {trades.length !== 0 && (
+            <Virtuoso
+              ref={ref}
+              style={{
+                height: height - 120 + "px",
```
taylor-adaptive, line 96: Where does the 120 come from? Can it be calculated? Or can we add a comment to explain?

src/new-client/src/App/Trades/TradesGrid/TradesGrid.tsx
hafidk: Sorry, thought you meant the //@ts-ignore, removed the ref, wasn't necesary after all

src/new-client/src/App/Trades/TradesCore.tsx
```diff
@@ -7,28 +7,51 @@ import { TradesHeader } from "./TradesHeader"
...
   font-size: 0.8125rem;
+  position: relative;
 `
 
 const SuspenseOnStaleData = createSuspenseOnStale(isBlotterDataStale$)
+const Trades: React.FC = () => {
+  const [height, setHeight] = useState<number>(0)
+  const [width, setWidth] = useState<number>(0)
+  const ref = useCallback((node) => {
+    let resizeObserver = new ResizeObserver(() => {})
```
taylor-adaptive, line 25: Rather than instantiating a `ResizeObserver` twice, you can do it once here and observe/disconnect in the if/else.

hafidk: fixed

src/new-client/src/App/Trades/TradesGrid/TradesGrid.tsx
hafidk: Added the comment, it's the sum of the rest of the elements within the Trades component

src/new-client/src/App/Trades/TradesFooter.tsx
hafidk: Fixed

hafidk:
Now I would have to adress the elephant on the room: The horitzontal scroll, if we test the current implementation with Virtuoso and we shrink the  scren until we get horintzontal scroll with the table, we will see that the vertical scroll it's within the child and gives this sensation that we can't scroll vertically (at least not until you position yourself to the end of the X axis)
Screenshot 2021-11-30 at 09 26 04

I've been trying different solutions but to no avail, the main problem here is that vertical scroll is from the Virtuoso table we define whereas the rest of the horitzontal scroll is from the Parent, we can't have the vertical scroll on the parent since it's the child the one who needs it (specially for the virtualization) and we can't have the horitzontal scroll for the child since not only he overflows, but also the header.

src/new-client/src/App/Trades/TradesGrid/TradesGrid.tsx
```diff
@@ -96,56 +43,89 @@ export const TradesGrid: React.FC = () => {
...
+        </TableBodyCell>
+      ))}
+    </TableBodyRow>
+  )
+}
+
+export const TradesGrid: React.FC<any> = ({ height = 200, width = 200 }) => {
+  const trades = useTableTrades()
+  const highlightedRow = useTradeRowHighlight()
+  const ref = useRef(null)
```
taylor-adaptive, line 72: This should be removed now it is unused

taylor-adaptive:
Overall happy with the code.

If we aren't able to come up with an elegant solution for the vertical scroll at smaller widths then I think it is down to product/stakeholders to get their opinion. Would they rather:

- Have the potential benefits of a virtualized table when we have a large set of trades, with the slight regression in UX with the scrollbar not being immediately visible at smaller screen widths
- Keep the current UX and forgo potential benefits of a virtualized table

Can you chat with Andres and explain him the UX regression and see what his thoughts are? 

I don't have a strong opinion either way. Having a virtualized table is nice to have, but we probably wont see much benefit from it. Equally, I don't think the regression in the UX is a huge issue, but it is a regression.

hafidk:
Removed unused ref and notified Andres. Task on hold until we decide what to do.

hafidk:
Parking task, we remain with current implementation.

PR #2009, opened 2021-11-30
fix(new-client): Draggable component depends on tearout

hafidk:
While I was working with the virtualized table, I disabled the tearout to test something but saw that I could still drag, since the draggable components are just an extension of the already implemented tearout, I don't know if its an expected behaviour to be able to drag a component but not tear it out 

Besides, there's also the issue that if we drag and a new window appears, we wouldn't be able to revert it unless we close it, since the tearIn would also be disabled.

I just added to the canDrag constant the need for supportsTearOut to be enabled, I don't know which would be the preferred behaviour for openfin so I left that as it is.

### DIFF
--- a/src/new-client/src/components/DraggableTearOut/canDrag.ts
+++ b/src/new-client/src/components/DraggableTearOut/canDrag.ts
@@ -1,2 +1,2 @@
 import { supportsTearOut } from "@/App/TearOutSection/supportsTearOut"
-export const canDrag = false && supportsTearOut
+export const canDrag = true && supportsTearOut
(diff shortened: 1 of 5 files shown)


src/new-client/src/components/DraggableTearOut/canDrag.ts
```diff
@@ -1 +1,2 @@
-export const canDrag = true
+import { supportsTearOut } from "@/App/TearOutSection/supportsTearOut"
+export const canDrag = true && supportsTearOut
```
taylor-adaptive, line 2: Minor - Can we add a comment here, it's a bit confusing. Something like
`If tear out is supported, allow drag`.

taylor-adaptive (commented):
It might be a good idea to align the terminology and be a bit more explicit. This took me a moment to get my head around.

`supportsTearOut` - The section can be torn out (and back in)
`supportsDragToTearOut` - The section can be torn out by dragging

It's fine to leave canDrag.openfin.ts as it is (`false`)

taylor-adaptive:
Also, please rename the branch without `&`, it is breaking the build

PR #2011, opened 2021-12-01
feat(new-client): implementet horitzontal resizer

hafidk:
# Description

Added horitzontal resizer to the web client, works similary to the previous Resize component, now we work with HResize and VResizer.

https://user-images.githubusercontent.com/11291319/144224224-154832d7-1f78-4c0b-8d83-e17159e73d19.mov

## Type of change

- [X] New feature (non-breaking change which adds functionality)

# Tasks:

- [X] Implement horitzontal resize bar
- [ ] Investigate how feasible is to increase width of charts as well and not just the parent wrapper
- [ ] Add canHoritzontalResize so it can be enabled/disabled
- [X] Fix horitzontal bar on top when tiles grid tears out (pre-commit, bar appears even though it doesn't resize since there's no sibling)

### DIFF
--- a/src/new-client/src/App/Analytics/Analytics.tsx
+++ b/src/new-client/src/App/Analytics/Analytics.tsx
@@ -9,7 +9,7 @@ const AnalyticsCore = lazy(() => AnalyticsCoreDeferred)
 
 const AnalyticsWrapper = styled.div<{ hideIfMatches?: string | null }>`
   height: 100%;
-  flex: 0 0 371px;
+  flex: 0 0 100%;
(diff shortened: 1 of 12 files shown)


taylor-adaptive:
I think we need to answer the question about making the analytics responsive. If this is possible (and desired) then we need to do that all together. At the moment the resizer is redundant.
Please chat to Andres and see if there is any appetite for this. We can park it for now, good job so far.

taylor-adaptive:
Closing, we can pick this up again if it ever becomes a requirement

PR #2022, opened 2022-01-10
fix(client): grid number filter (RT-1657-1658)

tprince232:
*Note - this is my first contributions - please let me know if anything is not following the conventions!*
*Also note - raised MR from main repo b/c PRs from a private fork don't work with the builds at the moment*

### Overview
- Convert string to number for tradeId comparisons
- Fixes === and !== operations

### Details
- For the tradeId column, the `numberFilterTrueOfTrade` is being called on filter change, and the value in the tradeId column is coming as a string and is typed as string|number.  Less than/greater than operator were evaluating the comparison between the number inputted and string value correctly but === and !== were always returning false
- The fix converts the string value found on the field to a number with parseFloat to allow the exact comparisons to compare correctly when the field value is not a number type
- parseFloat was used because it can also work in scenarios where a number column has floats formatted as strings

##### Equals
image image

##### Not equal
image
Note that 2659 is present in the search results.

##### Equals
image image

##### Not equal 
image
Note that 2403 is omitted from the search results.

### DIFF
--- a/src/new-client/src/App/Trades/TradesState/tableTrades.ts
+++ b/src/new-client/src/App/Trades/TradesState/tableTrades.ts
@@ -1,13 +1,6 @@
 import { startOfDay } from "date-fns"
 import { combineLatest, merge } from "rxjs"
-import {
-  delay,
-  filter,
-  map,
-  mergeMap,
-  scan,
-  startWith,
-} from "rxjs/operators"
+import { delay, filter, map, mergeMap, scan, startWith } from "rxjs/operators"
 import { bind } from "@react-rxjs/core"
 import { Trade, trades$ } from "@/services/trades"
 import type { ColField } from "./colConfig"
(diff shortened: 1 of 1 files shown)


src/new-client/src/App/Trades/TradesState/tableTrades.ts
```diff
@@ -162,7 +155,9 @@ const numFiltersTrueOfTrade = (
       return true
     }
 
-    const tradeNumber = trade[field as NumColField]
+    const tradeValue = trade[field as NumColField]
+    const tradeNumber =
+      typeof tradeValue === "number" ? tradeValue : parseFloat(tradeValue)
```
taylor-adaptive, line 160: Minor - I would have opted for `parseInt` to be explicit that we expect a whole number, but the outcome is the same so no need to change if you don't agree.

taylor-adaptive (approved):

PR #2023, opened 2022-01-13
fix(client) Connection Status Indicator (RT-1652)

tprince232:
*Please review carefully - new to working with Reactive Trader and I have limited experience with RXJS.   Open to ideas like moving the internet connection stuff to another file or combining the offline/online connection observables into one, etc.*

# Overview

- Created connection status observable from window offline/online events
- Update status to OFFLINE_DISCONNECTED when going offline if Hyrda status is connected
- Update status to latest Hydra status when user connects (if not connected, keep not connected)
- Update indicator to handle OFFLINE_DISCONNECTED status
- Update message to accurately describe disconnection reason

# Details
- When the user would loose internet connection, the `$ConnectionStatus` imported from Hydra was not updating that the user disconnected.  Instead,  the websocket remained "connected" until internet connection returned and values were updated without the user getting an indication that something was wrong
- I created an observable from the navigator.onLine boolean and the window offline and online events to indicate internet connection updates, "true" for online and "false" for offline.
- I created two additional observables that pipe the `online$` observable.   
    - `offlineDisconnect$` filters for online as false and gets the latest status/disposable to only take action if the user had a connection, and it updates the status to OFFLINE_DISCONNECTED
    - `onlineConnect$`filters for online as true and re-sends the latest Hydra connection status
- I modified the message for an OFFLINE_DISCONNECTED status in the disconnected status modal.  This is because in this scenario, Hydra is not "Trying to re-connect to the server", and what actually is happening is that the user is offline and their connection will resume once they return online. 

### Testing
- I tested against a variety of scenarios such as when the user is IDLE_DISCONNECTED and connection status changes and the experience appears to be what would be expected
- I tested by turning my WiFi on and off on my actual modem, and the experience was what I would expect - the Hydra connection did launch a "reconnecting" state which I hadn't seen when I just disabled/enabled wifi, as the wifi was unstable as my connection returned, but its not too clear to me why it did not just return to connected given the code I wrote.

### Alternate Approaches
- I was originally attempting to dispose of the connection when the user goes offline, the same way that the idle timeout disposes of the connection.  On reconnect, I would re-call initializeConnection.  However, there were some problems, the main pitfall being that the dispose was not completing from what appear to me to be to the lack of internet connection.   The `closeObserver` was not updating the status to disconnected upon dispose (again, I think because the dispose was not working without internet conncetion).  Then it would timeout after a minute.   If the user reconnected during this time (do a new WS connection), the user would connect, but then the OLD websocket closeObserver would send a disconnected message.

# Screen Capture

## Before

https://user-images.githubusercontent.com/10551665/149420862-6824d277-9611-471f-ba3c-ced4214da9aa.mov

## After

https://user-images.githubusercontent.com/10551665/149420880-3f8eab23-d8e0-4931-b8e2-308ed754d817.mov

## Side By Side
- Below is a side-by-side screen capture of the PR build and Dev env
- In this scenario I turn off my Modem/Router and then turn back on
- Upon turning off, there is slight delay before the device is detected to be offline
- Upon turning on, it appears the device is online but without a stable internet connection - I am curious about how what happens next happens as it appears that here React Trader can tell Hydra is disconnected and watches as Hydra tries to reconnect as the internet connection becomes more stable, eventually connecting 

https://user-images.githubusercontent.com/10551665/149553637-02d66d14-daba-4429-8aa4-c4f3d0334086.mov

### DIFF
--- a/src/new-client/src/services/connection.ts
+++ b/src/new-client/src/services/connection.ts
@@ -92,14 +92,17 @@ const idleDisconnect$: Observable<ConnectionStatus> = combineLatest([
   }),
 )
 
+/**
+ * Observable that fires when user's network connection status changes
+ * "true" for user is connected to the network, "false" for not connected to the network
+ * Note - Network connection does not guarantee internet connection
+ * The user's network might lose connection to the internet even though the device is connected to the network
+ * In this scenario $online will not fire that the user went offline (false positive for internet connection status)
+ */
 const online$: Observable<Boolean> = merge(
   of(navigator.onLine),
   fromEvent(window, "online").pipe(mapTo(true)),
   fromEvent(window, "offline").pipe(mapTo(false)),
-).pipe(
-  map((status) => {
-    return status
-  }),
 )
 
 // Update connection status when use goes offline
--- a/src/new-client/src/services/connection.ts
+++ b/src/new-client/src/services/connection.ts
@@ -18,6 +18,7 @@ import {
   distinctUntilChanged,
   filter,
   map,
+  mapTo,
   scan,
   startWith,
   withLatestFrom,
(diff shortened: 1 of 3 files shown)


src/new-client/src/services/connection.ts
```diff
@@ -86,12 +92,53 @@ const idleDisconnect$: Observable<ConnectionStatus> = combineLatest([
   }),
 )
 
+const online$: Observable<Boolean> = merge(
+  of(navigator.onLine),
+  fromEvent(window, "online").pipe(mapTo(true)),
+  fromEvent(window, "offline").pipe(mapTo(false)),
+).pipe(
```
taylor-adaptive, line 99: Minor - This pipe isn't necessary

taylor-adaptive (commented):
Code looks good.
However when I turn my wifi off (Windows 10) the ononline/onoffline events don't trigger. If I turn off in chrome developer tools the events do trigger.
This appears to be the correct way of detecting a connection so I'm not sure why they aren't triggering, but the functionality looked good when turning on/off through developer tools.

src/new-client/src/services/connection.ts
tprince232: Removed the pipe.

Also added comments above $online to be clear on Network connection vs Internet connection

taylor-adaptive (approved):

PR #2024, opened 2022-01-14
fix(client): Filter Text (RT-1651)

tprince232:
# Overview
- On Windows, select options were rending with white background and text was white on white
- Fix - apply theme background color directly to option child element on styled.select
- *Note - did not consult with a UI/PO, assumed theme light background would be acceptable*
- *Note - I would like to fix the padding around the text in the option tags but a quick check on adding margin/padding was not working and from what I found on the internet margin/padding on option tags are generally not supported across browsers*
- *Concern - this was working with Windows 11 but I'd wonder if older Windows/browser versions might not receive the options background color, is there a way for me to test this (e.g. browserstack)?*

# Screenshots
### Dark - Before and After
image image

### Light - Before and After (no change)
image

## Mac
- No impacts to MacOS select for light or dark mode
image


taylor-adaptive (approved):

taylor-adaptive:
I'm not sure if we have an Adaptive account with Browserstack but I tested on some older versions of Windows and chome/edge using a free account and it was working as expected. Good to merge!

PR #2042, opened 2022-01-26
fix(new-client): miscellaneous buy/sell numbers

tprince232:
## Trailing 0 for "bigNumber" portion of Buy/Sell number < 1
- Problem: For AUD/USD and NZD/USD conversions, `formatTo3Digits` was appending a trailing 0 on the first part of the buy/sell numbers, a rate of .82671 would render as .820 67 1, since .82 only has 2 significant digits.
- Fix: Added a check for `bigFigureNumber < 1 && pipsPosition === 4`, which accurately finds scenarios where the `bigFigureNumber` only has 2 significant digits and use `formatTo2Digits` in this scenario

### Before
Screen Shot 2022-01-26 at 11 21 26 AM

### After
Screen Shot 2022-01-26 at 11 18 41 AM

## Internationalized decimal separator usage for buy/sell when decimal separator is manually added
- Problem: When the pips number is 2 (JPY currencies), the decimal separator is manually added between `bigFigureNumber` and the pips because the `bigFigureNumber` is a whole number and the separator is not added by the formatter.  It was hard coded to `.`
- Fix: Use calculated internationalized `DECIMAL_SEPARATOR`

### Before
Screen Shot 2022-01-26 at 11 21 59 AM

### After
Screen Shot 2022-01-26 at 11 19 56 AM

## Tenths (final) digit was never passed through a `Intl` formatter, causing an inconsistency (Arabic)
- Problem: The Tenths (final) digit was rendering as a raw number while the rest of the buy/sell number was getting formatted at some point
- Fix: Use `formatSimple` to pass the `tenth` value through an `Intl` formatter

### Before
Screen Shot 2022-01-26 at 11 22 38 AM

### After
Screen Shot 2022-01-26 at 11 20 37 AM

### DIFF
--- a/src/new-client/src/App/LiveRates/Tile/PriceButton/PriceButton.tsx
+++ b/src/new-client/src/App/LiveRates/Tile/PriceButton/PriceButton.tsx
@@ -3,6 +3,7 @@ import { Direction } from "@/services/trades"
 import {
   customNumberFormatter,
   significantDigitsNumberFormatter,
+  DECIMAL_SEPARATOR,
 } from "@/utils/formatNumber"
 import { map, switchMap } from "rxjs/operators"
 import { bind } from "@react-rxjs/core"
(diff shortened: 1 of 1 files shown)


taylor-adaptive (approved):
Good spot with these

PR #2047, opened 2022-02-15
Feat/cypress

VenkatAragala:
- Installed Cypress
- Added Smoke test


.github/workflows/tests.yml
```diff
@@ -0,0 +1,29 @@
...
+  cypress-run:
+    name: Cypress run
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: app start
+        continue-on-error: true
+        run:  |
+              cd src/client
```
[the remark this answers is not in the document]
VenkatAragala: done the changes in latest commit

.github/workflows/tests.yml
```diff
@@ -0,0 +1,29 @@
...
+              npm install
```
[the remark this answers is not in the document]
VenkatAragala: done the changes in latest commit.  using npm ci now

.github/workflows/tests.yml
```diff
@@ -0,0 +1,29 @@
...
+              npm run tests  
+      # - name: wait for localhost
```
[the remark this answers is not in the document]
VenkatAragala: removed commented linesin latest commit

.gitignore
```diff
@@ -25,3 +25,5 @@ deploy/**/docker-compose.yml
 *.sln
 lerna-debug.log
 .DS_Store
+/src/cypress/videos/
+/src/cypress/screenshots/
```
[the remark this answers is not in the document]
VenkatAragala: removed it from .gitignore

src/client/package.json
```diff
@@ -23,6 +23,8 @@
     "test": "jest",
     "lint": "eslint",
     "dev": "cross-env TARGET=web vite",
+    "tests": "cross-env TARGET=web vite& cd ../cypress && npm install && sleep 60 && npx cypress run --spec 'cypress/integration/demo.feature' --browser=chrome",
```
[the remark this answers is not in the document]
VenkatAragala: removed sleep and npm install from the script line

src/cypress/package.json
```diff
@@ -0,0 +1,15 @@
+{
+  "scripts": {
+    "tests": "npx cypress run --spec 'cypress/integration/demo.feature' --browser=chrome"
```
taylor-adaptive, line 3: Let's remove the `--spec` and expand these scripts.
For example -
`test:cmd` - run cypress run for all tests
`test:interactive` - open cypress interactive window

.github/workflows/tests.yml
```diff
@@ -0,0 +1,39 @@
+# This workflow will do a clean installation of node dependencies, cache/restore them, build the source code and run tests across different versions of node
+# For more information see: https://help.github.com/actions/language-and-framework-guides/using-nodejs-with-github-actions
+
+name: Cypress Tests
+on: [push]
```
taylor-adaptive, line 5: Let's run these on 
- pull requests
- Merges to master (this is the dev deployment) or env/uat (this is the uat deployment)
- Tags starting with `v` (this is the prod deployment)
Check out the other files in `.github/workflow`

src/client/package.json
```diff
@@ -23,6 +23,7 @@
     "test": "jest",
     "lint": "eslint",
     "dev": "cross-env TARGET=web vite",
+    "starts": "cross-env TARGET=web vite& cd ../cypress",
```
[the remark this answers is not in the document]
VenkatAragala: with existing start, its not going to next step as appication is not running in background mode..

src/cypress/cypress/README.md
```diff
@@ -0,0 +1,38 @@
+[![GitHub Workflow Status](https://img.shields.io/github/workflow/status/AdaptiveConsulting/ReactiveTraderCloud/CI)](https://github.com/AdaptiveConsulting/ReactiveTraderCloud/actions?query=workflow%3ACI)
+[![GitHub release (latest by date)](https://img.shields.io/github/v/release/AdaptiveConsulting/ReactiveTraderCloud)](https://github.com/AdaptiveConsulting/ReactiveTraderCloud/releases/latest)
+[![GitHub](https://img.shields.io/github/license/AdaptiveConsulting/ReactiveTraderCloud)](https://opensource.org/licenses/Apache-2.0)
+
+[![image](https://raw.githubusercontent.com/AdaptiveConsulting/ReactiveTrader/master/images/adaptive-logo.png)](http://weareadaptive.com/)
+
+# Reactive Trader®
```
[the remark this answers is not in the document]
VenkatAragala: removed in latest commit

src/cypress/cypress/plugins/index.js
```diff
@@ -0,0 +1,30 @@
...
+// ***********************************************************
+
+// This function is called when a project is opened or re-opened (e.g. due to
+// the project's config changing)
+
+/**
+ * @type {Cypress.PluginConfig}
+ */
+ const cucumber = require('cypress-cucumber-preprocessor').default
+ let tradedetails
```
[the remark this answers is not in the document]
VenkatAragala: For cucumber thats the syntax, we require the lib from index.js. is there any other way?

src/cypress/package.json
VenkatAragala: fixed in latest commit

.github/workflows/tests.yml
VenkatAragala: fixed in latest commit

src/cypress/cypress/integration/demo/trade.js
```diff
@@ -0,0 +1,27 @@
...
+
+})
+And('{string} price tile is displayed', (symbol) => {
+    cy.VerifyCurrencyOnPage(symbol)
+
+})
+
+When('user performs {string} trade for {string} of {string}', (direction, notional, symbol) => {
+  
+    cy.VerifySuccessTradeExecution(symbol, notional, direction)
```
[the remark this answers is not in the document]
VenkatAragala: fixed the function names in latest commit

src/cypress/cypress/README.md
```diff
@@ -0,0 +1,38 @@
+[![GitHub Workflow Status](https://img.shields.io/github/workflow/status/AdaptiveConsulting/ReactiveTraderCloud/CI)](https://github.com/AdaptiveConsulting/ReactiveTraderCloud/actions?query=workflow%3ACI)
```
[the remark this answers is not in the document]
VenkatAragala: remvoed the section in latest commit

src/cypress/cypress/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+6. Now hit the command in different terminal `npx cypress open` to open the cypress interactive window.
+
+7. Now, click on the feature file present on the Testrunner window to trigger the test.
+
+8. If you want to run the test in terminal, please hit `npx cypress run --spec <path to feature file> --browser chrome`
+
+9. To run test using command line, please use `npm run tests`
+
+9. Cucumber Report Integration is in progress currently.
```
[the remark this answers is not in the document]
VenkatAragala: updated in latest commit

src/cypress/cypress/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+1. Fork and clone the ReactiveTraderCloud repo
+
+2. Start the application on localhost by hitting `npm run start`
+
+4. Open a browser and navigate to http://localhost:1917 to see the application running. 
+
+5. Navigate to `/cypress/integration` to see the list of feature files
+
+6. Now hit the command in different terminal `npx cypress open` to open the cypress interactive window.
```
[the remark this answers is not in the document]
VenkatAragala: I have mentioned the npm script too,  this command is to run in interactive mode if someone wants to open cypress runner

src/client/package.json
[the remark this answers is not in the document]
taylor-adaptive: I don't understand what you mean. I'm sure this isn't needed, please explain

src/cypress/cypress/README.md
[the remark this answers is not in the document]
taylor-adaptive: What I mean to say is, let's have an npm script that does this too

src/cypress/cypress/plugins/index.js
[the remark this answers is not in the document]
taylor-adaptive: `let tradedetails` is redundant

.github/workflows/tests.yml
```diff
@@ -0,0 +1,48 @@
...
+
+name: Cypress Tests
+on:
+  pull_request:
+    types:
+      - opened
+      - synchronize
+  push:
+    branches:
+      - '**'
```
taylor-adaptive, line 12: I think on all branches might be a bit of a waste of resources. Merges to master and env/uat branches are the important ones

src/package.json
```diff
@@ -0,0 +1,16 @@
...
+  },
+  "cypress-cucumber-preprocessor": {
+    "nonGlobalStepDefinitions": true
+  },
+  "devDependencies": {
+    "cypress": "^8.7.0",
+    "cypress-cucumber-preprocessor": "^4.3.1"
+  },
+  "dependencies": {
+    "cypress-configuration-builder": "^1.2.4"
```
taylor-adaptive, line 14: Move to devDeps

src/cypress/plugins/index.js
```diff
@@ -0,0 +1,30 @@
...
+// ***********************************************************
+
+// This function is called when a project is opened or re-opened (e.g. due to
+// the project's config changing)
+
+/**
+ * @type {Cypress.PluginConfig}
+ */
+ const cucumber = require('cypress-cucumber-preprocessor').default
+ let tradedetails
```
taylor-adaptive, line 20: This is redundant

taylor-adaptive (commented):
If you fancy a challenge, we should have lint-staged running on precommit hook to format these files, this may not be working because it's outside of the /client folder.
If not, can you go through the .js files and format them (prettier) before merging (and we can setup the precommit hook later)? You should be able to do this in your IDE or command line.
Lastly, a video has crept into the branch.

src/cypress/plugins/index.js
VenkatAragala: I removed that line but failed to commit, will do that in next commit

src/package.json
VenkatAragala: Fixed it in the latest commit, whenever i save the file the prettier will format the file, removed the video

src/package-lock.json
```diff
@@ -0,0 +1,5865 @@
+{
```
taylor-adaptive, line 1: There shouldn't be a package-lock in `/src`, it would have been generated when running `npm install` from `src`, it should be run where a matching `package.json` file is.

taylor-adaptive (commented):
The introduction of cypress into the codebase and github workflow looks good.

Having another look at the actual testing code I would still say this feels too brittle.

For example - 
```'
div[data-testid="tile-' +
        currency +
        '"] > div > div > div:nth-child(2) div:nth-child(2) input'
```
If the dom structure changed, the test would fail. Targetting the input directly would be safer. In this particular example, `div[data-testid="tile-${currency}"] input` might actually be enough.

Also, sharing state between Cypress commands feels like something we should be avoiding.
`verifyTradeSuccess` is relying on `performTrade` to have been run first to do it's logic and store the values in some global variables. What happens if these get executed in a wrong order, or with mismatched arguments, or if we run tests in parallel?

Once we have a nice way of mocking the server, the solution will be much simpler, but I'm hesitant to use this approach until then. Perhaps there is a way of returning values from Cypress commands, so if we run `performTrade` which returns us the value of current price, we can then assert on that in the test

src/package-lock.json
VenkatAragala: removed package-lock.json from /src in latest commit

VenkatAragala:
fixed the element locator as per the suggestion.

[Taylor]-
 verifyTradeSuccess is relying on performTrade to have been run first to do it's logic and store the values in some global variables. What happens if these get executed in a wrong order, or with mismatched arguments, or if we run tests in parallel?

[Venkat] - 
Here, even if we return the value, then also it will fail if we execute in wrong order or in parallel.  prerequisite for verifyTradeSuccess is to have performTrade function executed before. 

Also, I tried returning the values but due to async but it did not return the latest change and was only returning initial value which was assigned while declaration. 
Cucumber is using gherkin language, so no two steps in scenario will be executed in parallel.

src/cypress/support/commands.js
```diff
@@ -0,0 +1,133 @@
...
+      finalList.push(textList[8])
+      finalList.push(textList[14] + '-' + textList[15].replace('.', '') + '-' + currentYear)
+      finalList.push('JPW')
+    })
+  })
+})
+
+Cypress.Commands.add('verifyTradeSuccess', () => {
+  cy.log('').then(() => {
+    expect(Number(currentPrice)).to.eq(Number(finalList[7]))
```
taylor-adaptive, line 116: I mentioned before, but the concept of extracting bits into some `finalList` state is confusing and hard to reason about.
To verify a trade success I think we can be a bit simpler by just asserting that the tile contains a string that we expect.
Take a look at my early examples I originally showed you - https://github.com/taylor-adaptive/ReactiveTraderCloud/blob/feat/cyrpress/src/new-client/cypress/integration/trade-execution.ts#L28
I find this a lot easier to reason about, and therefor maintain. Interested to hear your thoughts.

Finally, for the bits that we can't really do nicely (like asserting on the current price) I think it's acceptable to leave it as a TODO until we have a mocking solution (similar to what I did in my earlier examples)

src/cypress/support/commands.js
```diff
@@ -0,0 +1,133 @@
...
+const dayjs = require('dayjs')
+var currentYear = dayjs().get('year')
+let currency
+
+// Cypress.Commands.add('dismiss', { prevSubject: 'optional'}, (subject, options) => { ... })
+
+Cypress.Commands.add('VerifyCurrencyOnPage', symbol => {
+  let currency = symbol.replace('/', '')
+
+  cy.log('').then(() => {
```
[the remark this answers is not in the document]
VenkatAragala: We are using .then method here, which yields the current state of the variables.
so before assertions we are making sure of getting the latest value. This is one of the way mentioned in cypress documentation.

but in this case, I have removed the extra variable and .then()
cy.log is just a method to get .then(), we can use any cy method.

src/cypress/support/commands.js
VenkatAragala: I am using finalList to make it scalable, Lets if in future we need to integrate blotter tests, where after trade we need to check details in blotter row. then finaList will help us in fast assertions. 

we can give any other name to it, if it sounds confusing.

taylor-adaptive: I would say the opposite, I think using an array to hold random values is not scalable.

For example - 
There is no way of knowing what `finalList[6]` holds unless you step through the whole code and count through the pushes to `finalList`, and then that contains a `textList[x]` which we then have to look at what `textList` is and try and figure out what value that holds. As someone who may need to read and update these tests, it will be difficult.

I'm not an expert with testing so I am open to solutions, but this doesn't feel like the best way to go about things.

@lidia-adaptive can you take a look through this code and the early examples I mention in the above comment and offer your opinion on approaches?

lidia-adaptive: Absolutely agree with you @taylor-adaptive, this is a mystery-guest or magic number anti-pattern right there. 
Looking at your example.

src/cypress/support/commands.js
```diff
@@ -0,0 +1,173 @@
...
+
+    cy.get(
+      'div[data-testid="tile-' +
+        currency +
+        '"] > div > div > div:nth-child(2)  > div:nth-child(2) > div > button[direction=' +
+        direction +
+        ']'
+    ).click({ force: true })
+    if (currency == 'EURJPY') {
+      cy.wait(2000)
```
lidia-adaptive, line 79: Why are we waiting here?

src/cypress/support/commands.js
```diff
@@ -0,0 +1,173 @@
...
+        finalList.push(textList[8])
+        finalList.push(textList[14] + '-' + textList[15].replace('.', '') + '-' + currentYear)
+        finalList.push('JPW')
+      }
+    })
+  })
+})
+
+Cypress.Commands.add('verifyTradeRejected', () => {
+  cy.log('').then(() => {
```
taylor-adaptive, line 137: An example of how to assert this without extracting bits into an ambiguous array in a previous command - 
```
Cypress.Commands.add('verifyTradeRejected', (symbol) => {
  cy.get(`[data-testid="tile-${symbol}"]`).contains('rejected')
}
```

src/cypress/integration/demo/trade.js
```diff
@@ -0,0 +1,99 @@
...
+And('{string} price tile is displayed', symbol => {
+  cy.VerifyCurrencyOnPage(symbol)
+})
+
+When('user performs {string} trade for {string} of {string}', (direction, notional, symbol) => {
+  cy.performTrade(symbol, notional, direction)
+})
+
+Then('trade is success for {string} and {string}', (notional, direction) => {
+  cy.verifyTradeSuccess(notional, direction)
```
taylor-adaptive, line 15: I still don't like that there is some state held in the commands, when you read this file you have no idea that is happening behind the scenes and the command file is hard to read and follow code.

At a minimum I think the state should be held here, to keep each command isolated and to give the control the tests. 

Example - 

```
let symbol;

When('user performs {string} trade for {string} of {string}', (direction, notional, symbol) => {
  symbol = symbol
  cy.performTrade(symbol, notional, direction)
})

Then('trade is success for {string} and {string}', (notional, direction) => {
  cy.verifyTradeSuccess(notional, direction, symbol)
})
```

Alternatively (and perhaps more appropriate) you could look into using sharing context - https://github.com/TheBrainFamily/cypress-cucumber-preprocessor#sharing-context

src/cypress/integration/demo/trade.js
```diff
@@ -0,0 +1,99 @@
...
+})
+
+And('User clicks on Requote button {string}', symbol => {
+  cy.clickOnRequoteButton(symbol)
+})
+
+Then(
+  'After 60 secs Prices will be disapper and requote button appears again for {string}',
+  symbol => {
+    cy.wait(10000) // there is a bug in flow
```
taylor-adaptive, line 60: Can you elaborate on what the problem here is

src/cypress/integration/demo/trade.js
```diff
@@ -0,0 +1,99 @@
...
+When('User clicks on price view', () => {
+  cy.clickPriceView()
+})
+
+Then('User is not able to see price graph for any Currencies', () => {
+  cy.verifyPriceView()
+})
+
+And('User clicks on graph view', () => {
+  cy.wait(2000)
```
taylor-adaptive, line 85: Why is this needed?

src/cypress/integration/demo/trade.js
```diff
@@ -0,0 +1,99 @@
...
+
+Then('User is able to see price graph for all currencies', () => {
+  cy.verifyGraphView()
+})
+
+And('User clicks on {string} tab', currency => {
+  cy.clickCurrencyTab(currency)
+})
+
+Then('User sees curreiny combination only for {string}', currency => {
```
taylor-adaptive, line 97: Minor - Type in curreiny

src/cypress/integration/pages/TradePage.js
```diff
@@ -0,0 +1,79 @@
+class TradePage {
+  getSymbol(symbol) {
+    return cy.get('div[data-testid="tile-' + symbol.replace('/', '') + '"]')
```
taylor-adaptive, line 3: `symbol.replace('/', '')` is used a lot throughout, consider creating a util function for it to reduce duplication.

taylor-adaptive: Also, would `getTile` be a more appropriate name?

src/cypress/integration/pages/TradePage.js
```diff
@@ -0,0 +1,79 @@
...
+  }
+  enterNotional(symbol) {
```
taylor-adaptive, line 5: `enterNotional` is misleading. This function is returning the notional input of a tile.

src/cypress/integration/pages/TradePage.js
```diff
@@ -0,0 +1,79 @@
...
+        symbol +
+        '"] > div > div > div:nth-child(2)  > div:nth-child(2) > div > button[direction=' +
+        direction +
+        '] > div > div'
+    )
+  }
+
+  getDirectionButton(symbol, direction) {
+    return cy.get(
+      'div[data-testid="tile-' +
```
taylor-adaptive, line 21: Again, these selectors are too brittle.

`div[data-testid="tile-${symbol}"] button[direction="${direction}"]` would be sufficient.

taylor-adaptive: `div[data-testid="tile-' + symbol + '"]` is also used a lot throughout. Extracting this to a utli function would be a good idea.

src/cypress/integration/pages/TradePage.js
```diff
@@ -0,0 +1,79 @@
...
+  getRFQButton(symbol) {
+    return cy.get(
+      'div[data-testid="tile-' +
+        symbol +
+        '"] > div > div > div:nth-child(2)  > div:nth-child(2) > div > div:nth-child(4) button'
+    )
+  }
+
+  getRejectButton(symbol) {
+    return cy.get('div[data-testid="tile-' + symbol + '"] button[data-testid=rfqReject]')
```
taylor-adaptive, line 38: This is an example of a good selector

src/cypress/integration/pages/TradePage.js
```diff
@@ -0,0 +1,79 @@
...
+  }
+  getRequoteButton(symbol) {
+    return cy.get(
+      'div[data-testid="tile-' +
+        symbol +
+        '"] > div > div > div:nth-child(2)  > div:nth-child(2) > div > div:nth-child(4) button'
```
taylor-adaptive, line 44: This is an example of a bad selector. See `getRejectButton`, we should do something similar. If the attribute doesn't exist on the element, you can add it yourself. It will be useful for you to get familiar with how to do this, I can help where necessary.

src/cypress/integration/pages/TradePage.js
```diff
@@ -0,0 +1,79 @@
...
+    )
+  }
+
+  getWaitingTextMessage(symbol) {
+    return cy.get('div[data-testid=tile-' + symbol + '] div[role=dialog] > div[role=alert]')
+  }
+
+  getConfirmationTextMessage(symbol) {
+    return cy.get('div[data-testid=tile-' + symbol + '] div[role=dialog] > div[role=alert]')
```
taylor-adaptive, line 53: `getWaitingTextMessage` and `getConfirmationTextMessage `return the same thing. Consider a single function with generic naming - perhaps something like `getTileResponseMessage` or similar

src/cypress/integration/pages/TradePage.js
```diff
@@ -0,0 +1,79 @@
...
+  getPriceViewButton() {
+    return cy.get('li[data-qa-id=workspace-view-Analytics] > svg > g')
+  }
+
+  getGraphViewButton() {
+    return cy.get('li[data-qa-id=workspace-view-Normal] > svg > g')
+  }
+
+  getPriceView() {
+    return cy.get('svg')
```
taylor-adaptive, line 69: This isn't specific enough

src/cypress/integration/pages/TradePage.js
```diff
@@ -0,0 +1,79 @@
+class TradePage {
```
taylor-adaptive, line 1: I like the `TradePage` concept. General comment about the selectors, we need to make some less brittle and some more specific. I've made a few comments, but please apply those comments to all the selectors.

src/cypress/support/commands.js
```diff
@@ -0,0 +1,220 @@
...
+    .enterNotional(symbol.replace('/', ''))
+    .clear()
+    .type(notional)
+})
+
+Cypress.Commands.add('clickOnRejectButton', symbol => {
+  tradePage.getRejectButton(symbol.replace('/', '')).click()
+})
+
+Cypress.Commands.add('clickOnRejectButton', symbol => {
```
taylor-adaptive, line 49: This is duplicated above

src/cypress/support/commands.js
```diff
@@ -0,0 +1,220 @@
...
+  tradePage.getRejectButton(symbol.replace('/', '')).click()
+})
+
+Cypress.Commands.add('clickOnRequoteButton', symbol => {
+  tradePage.getRequoteButton(symbol.replace('/', '')).click()
+})
+
+Cypress.Commands.add('clickOnRequoteButton', symbol => {
```
taylor-adaptive, line 57: This is duplicated above

src/cypress/support/commands.js
```diff
@@ -0,0 +1,220 @@
...
+    tradePage.getConfirmationTextMessage(currency).then(el => {
+      cy.log('Confirmation text ' + el.text())
+      confirmationText = el.text()
+    })
+  })
+})
+
+Cypress.Commands.add('verifyTradeRejected', () => {
+  cy.log('').then(() => {
+    expect(confirmationText).to.eq('Your trade has been rejected')
```
taylor-adaptive, line 116: This assertion is better, but see earlier comment about using state (confirmationText) inside the commands file.
Also, with an assertion like this, we shouldn't need to do it inside a `cy.log()`?

src/cypress/support/commands.js
```diff
@@ -0,0 +1,220 @@
...
+})
+
+Cypress.Commands.add('verifyPriceView', () => {
+  tradePage
+    .getPriceView()
+    .find('path[stroke]')
+    .should('have.length', 1)
+})
+Cypress.Commands.add('verifyGraphView', () => {
+  cy.wait(3000)
```
taylor-adaptive, line 127: Why the wait?

src/cypress/integration/demo/trade.js
```diff
@@ -0,0 +1,99 @@
...
+  symbol => {
+    cy.wait(10000) // there is a bug in flow
+    cy.verifyRequoteButton(symbol)
+  }
+)
+
+Then(
+  'User sees the {string} value on {string} field for {string}',
+  (expected, notional, symbol) => {
+    if (notional == '1m' || notional == '1k') {
```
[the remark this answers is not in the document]
VenkatAragala: fixed in latest commit

src/cypress/integration/demo/trade.js
VenkatAragala: fixed in latest commit

src/cypress/integration/pages/TradePage.js
VenkatAragala: fixed in latest commit

src/cypress/integration/pages/TradePage.js
VenkatAragala: Both are fixed in latest commit

src/cypress/integration/pages/TradePage.js
VenkatAragala: we need to add attribute,  but in this case, I have used .find('path[stroke]') also, in command.js

src/cypress/support/commands.js
```diff
@@ -0,0 +1,220 @@
...
+let tradePage = new TradePage()
+Cypress.Commands.add('VerifyCurrencyOnPage', symbol => {
+  tradePage
+    .getSymbol(symbol)
+    .scrollIntoView()
+    .should('be.visible')
+})
+
+Cypress.Commands.add('verifyDisabledDirections', symbol => {
+  tradePage.getDirectionButton(symbol.replace('/', ''), 'Buy').should('be.disabled')
```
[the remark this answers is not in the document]
VenkatAragala: fixed in latest commit

src/cypress/integration/pages/TradePage.js
VenkatAragala: fixed in latest commit.

src/cypress/support/commands.js
VenkatAragala: fixed in latest commit

src/cypress/support/commands.js
VenkatAragala: ficed in latest commit

src/cypress/support/commands.js
```diff
@@ -0,0 +1,220 @@
...
+Cypress.Commands.add('clickOnRequoteButton', symbol => {
+  tradePage.getRequoteButton(symbol.replace('/', '')).click()
+})
+
+Cypress.Commands.add('verifyNotionalValue', (expected, symbol) => {
+  tradePage.getNotionalValue(symbol.replace('/', '')).should('contain', expected)
+})
+
+Cypress.Commands.add('verifyErrorMessage', (expected, symbol) => {
+  cy.contains(expected).should('be.visible')
```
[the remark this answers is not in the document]
VenkatAragala: fixed in latest commit

PR #2062, opened 2022-05-03
chore(client): upgrade OpenFin to v25

bruceharris:
The latest FDC3 API does not support sending messages across different platforms.
We want to send messages to ReactiveTraderCloud. Since these two applications
have distinct UUIDs in their OpenFin manifest (app.json), they cannot communicate
via FDC3 Interop API. We don't want to change to a shared UUID because the
Launcher app currently references ReactiveAnalytics and ReactiveTrader by their
UUIDs, so they need to be distinct.

In order to upgrade OpenFin runtime to latest version without breaking interop
between ReactiveTrader and ReactiveAnalytics, we are moving from using the
FDC3 Interop API to instead use the OpenFin InterApplicationBus
https://developer.openfin.co/docs/javascript/25.100.69.8/InterApplicationBus.Channel.html
which supports communicating across OpenFin platforms.

## Test Plan

1. pull upgrade-openfin branches for RT and RA repos
2. ensure no OpenFin processes running
3. clear OpenFin cache
4. run Launcher, RT, and RA
    1. `ReactiveTraderCloud\src\client> npm run openfin:dev`
    1. `ReactiveTraderCloud\src\client> npm run launcher:start`
    1. `ReactiveAnalytics> yarn start`
    1. launch RA and RT from launcher
5. in OpenFin
    1. execute a trade in RT
        1. verify it appears in blotter
        1. verify trade notifications appear
    1. select a trade in the RT blotter
        1. verify that selected currency pair appears in RA
        1. to test reconnecting to InterApplicationBus, close and relaunch RA
        1. verify that selecting trade in RT blotter still populates selected currency pair in RA
    1. save a snapshot, modify view, save another snapshot
    1. restore initial snapshot, verify it appears as expected
6. Chrome
    1. `ReactiveTraderCloud\src\client> npm run openfin:dev`
    1. execute a trade in RT, verify it appears in blotter
    1. select a trade in blotter, verify no messages in console

PR #2063, opened 2022-05-23
feat(client): Add navigation to credit dashboard

bruceharris:
- currently only supported in web browser, not in OpenFin or other platforms
- `DropdownMenu.tsx` is a generalized implementation of `CurrencyOptions.tsx`, which as an aside, is
  - tightly coupled to the currency use case
  - [unused at present, as far as I can tell](https://github.com/AdaptiveConsulting/ReactiveTraderCloud/blob/master/src/client/src/App/LiveRates/MainHeader/CurrencyOptions.tsx#L13)
- This PR currently targets the `upgrade-openfin` branch, but #2062 should be merged first, and once merged, we should change the base of this PR to `master`
- set `VITE_IS_CREDIT_ENABLED=yes` in `.env.local` to enable the feature

nav-browser


taylor-adaptive:
The work around routes and feature toggling all looks fine.
The `CurrencyOptions` is used - It's hidden but shown at smaller screen sizes/mobile.
I'd suggest keeping the `DropdownMenu` component you've created and refactor `CurrencyOptions` to use it so we're not duplicating it. 
After that, this is good to go.

bruceharris:
Thanks @taylor-adaptive - I removed `CurrencyOptions` in favor of `DropdownMenu`, and also fixed a bug where the state of the currency selection state was previously not synchronized between the 2 currency selection components.

## Before:
currency options bug

## After fix:
currency options bug fixed

src/client/src/theme/ThemeContext.tsx
```diff
@@ -73,7 +73,7 @@ export const ThemeProvider: React.FC<Props> = ({
   )
 }
 
-export const useTheme = () => {
+export const useThemeName = () => {
```
taylor-adaptive, line 76: What was your reasoning behind this name change? The hook returns more than just the theme name.

taylor-adaptive:
> Thanks @taylor-adaptive - I removed `CurrencyOptions` in favor of `DropdownMenu`, and also fixed a bug where the state of the currency selection state was previously not synchronized between the 2 currency selection components.
> 
> ## Before:
> currency options bug [ ](https://user-images.githubusercontent.com/1081979/170561868-87ffe0e1-4a37-40bf-b463-bdaee4f628be.gif) currency options bug [ ](https://user-images.githubusercontent.com/1081979/170561868-87ffe0e1-4a37-40bf-b463-bdaee4f628be.gif)
> 
> ## After fix:
> currency options bug fixed [ ](https://user-images.githubusercontent.com/1081979/170561859-d4ba6c11-4e24-494f-a8f8-d78ba4f3dd84.gif) currency options bug fixed [ ](https://user-images.githubusercontent.com/1081979/170561859-d4ba6c11-4e24-494f-a8f8-d78ba4f3dd84.gif)

Looks good!

src/client/src/theme/ThemeContext.tsx
bruceharris: Noted in commit message of - https://github.com/AdaptiveConsulting/ReactiveTraderCloud/pull/2063/commits/9baeda8c1ff384456174c5cbfbee23275a159e01 - to avoid confusion with
https://styled-components.com/docs/api#usetheme which returns a theme object and is used elsewhere in the app, e.g. https://github.com/AdaptiveConsulting/ReactiveTraderCloud/blob/master/src/client/src/OpenFin/Window/LayoutLock.tsx#L3

Open to suggestions for a better name, or we can just drop this commit if the original is preferred.

taylor-adaptive: It doesn't bother me too much, I think we probably have other instances of similar names with dependencies.
If you feel strongly about it then maybe something like `useApplicationTheme` or `useReactiveTraderTheme`, or drop the change.

bruceharris:
@TobyLi96  @taylor-adaptive removed the `useTheme` refactor commit and retargeted the PR to master

taylor-adaptive (approved):

PR #2067, opened 2022-06-07
fix(client): Fix PWA not triggering

TobyLi96:


ivanfalanga-adaptive (approved):

src/client/vite.config.ts
```diff
@@ -203,6 +203,8 @@ const copyWebManifestPlugin = (dev: boolean) => {
             contents
               .toString()
               .replace(/<BASE_URL>/g, getBaseUrl(dev))
+              // We want the PWA banner to show on reactivetrader.com
+              .replace(/web\.prod\./g, "")
```
ivanfalanga-adaptive, line 207: This would work for now, maybe we could do some work in the future to have the reverse proxy do these transformations before responding back. That way independently of whether we hit Reactive Trader from web.prod or reactivetrader or any other origin that is allowed we can always map it back to the correct URL

PR #2090, opened 2022-07-29
fix(client): filter on blotter - ios (2547)

patelpaavan55:
- [x] fixed no keyboard input on blotter filter on iOS devices

Before: No action on keyboard input

https://user-images.githubusercontent.com/25262082/181528239-b9634207-7240-45d5-a351-218e51ebd074.mp4

After:

https://user-images.githubusercontent.com/25262082/181528648-21cf645e-c457-41cb-90e8-062a5c7f2df3.mp4


ivanfalanga-adaptive:
I had to remove the package.lock changes as the were introduced probably as a result of running npm install in an older node version

ivanfalanga-adaptive (approved):

PR #2093, opened 2022-08-05
fix(positions): bubbles div size fix

umitao:
d3Effect function where the svg for the positions bubbles are created, is running before the encompassing div gets correct width and height leading to an error in the detection of the size of this div(it is 0 when the function runs on load/reload of the page). 
For demo purposes it is conditioned to get a fixed width and height(288px) temporarily as the dimensions of this element do not change with resizing.

### DIFF
--- a/src/client/src/App/Analytics/Positions/Positions.tsx
+++ b/src/client/src/App/Analytics/Positions/Positions.tsx
@@ -24,9 +24,11 @@ const d3Effect = (chartDiv: HTMLDivElement) => {
     .attr("data-testid", "tooltip")
 
   // const svg: Selection<SVGElement> = select(chartDiv).select("svg")
-  const { width, height } = chartDiv
-    ? chartDiv.getBoundingClientRect()
-    : { width: 0, height: 0 }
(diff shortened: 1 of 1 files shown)


src/client/src/App/Analytics/Positions/Positions.tsx
```diff
@@ -24,9 +24,11 @@ const d3Effect = (chartDiv: HTMLDivElement) => {
...
 
   // const svg: Selection<SVGElement> = select(chartDiv).select("svg")
-  const { width, height } = chartDiv
-    ? chartDiv.getBoundingClientRect()
-    : { width: 0, height: 0 }
+  //TODO Refactor - Make sure getBoundingClientRect runs after the chartDiv has width
+  const { width, height } =
+    chartDiv.getBoundingClientRect().width !== 0
+      ? chartDiv.getBoundingClientRect()
+      : { width: 288, height: 288 }
```
TobyLi96, line 31: I think the width is more like 274.4 when loaded correctly.

TobyLi96 (approved):

PR #2098, opened 2022-08-12
feat(client): incoming quote notification

TobyLi96:


src/client/src/Web/MainCreditRoute.tsx
```diff
@@ -3,13 +3,25 @@ import { getTornOutSections } from "@/App/TearOutSection/state"
...
+import { useEffect } from "react"
 import MainLayout from "./MainLayout"
 
 const CREDIT_TEAR_OUT_SECTIONS = ["newRfq", "creditBlotter"] as const
 const useTornOutSections = getTornOutSections(CREDIT_TEAR_OUT_SECTIONS)
 
 const MainCreditRoute: React.FC = () => {
   const tornOutSections = useTornOutSections()
+
+  useEffect(() => {
```
ivanfalanga-adaptive, line 19: Does this need to be part of an effect or can we host it up in the file instead? Do we need the notifications to work outside the MainCreditRoute? (e.g. when the sections are torn out)

ivanfalanga-adaptive (approved):

PR #2101, opened 2022-08-18
feat(client): Add highlights to credit RFQ, quotes and blotter

TobyLi96:


src/client/src/App/Trades/TradesState/tableTrades.ts
```diff
@@ -378,7 +378,7 @@ export const [useTradeRowHighlight] = bind(
     ),
     newTradeId$,
     newTradeId$.pipe(
-      delay(1000),
+      delay(4000),
```
ivanfalanga-adaptive, line 381: This is changing the delay for both credit and FX for row highlights, can we confirm that this is what we want?

ivanfalanga-adaptive (approved):

PR #2105, opened 2022-08-22
feat(client): Make credit tiles more responsive

TobyLi96:


ivanfalanga-adaptive (approved):

src/client/src/App/Credit/CreditRfqs/CreditRfqCards/styled.tsx
```diff
@@ -4,11 +4,12 @@ import styled, { css, keyframes } from "styled-components"
...
-export const CreditRfqCardsWrapper = styled.div`
-  display: flex;
-  flex-wrap: wrap;
+export const CreditRfqCardsWrapper = styled.div<{ empty: boolean }>`
+  display: ${({ empty }) => (empty ? "flex" : "grid")};
+  grid-template-columns: repeat(auto-fill, minmax(300px, 1fr));
+  grid-gap: 8px;
   align-content: start;
-  gap: 8px;
+  margin-right: 8px;
```
ivanfalanga-adaptive, line 12: we might be able to get rid of this margin here if we put the proper padding on the parent container but let's merge this and can look into removing it later so that it is even from both left and right

PR #2112, opened 2022-08-25
feat(client): quote row highlight improvements

ivanfalanga-adaptive:


TobyLi96 (approved):
The dot and price for a sell RFQ are the same blue as a buy RFQ. Should it be red for a sell RFQ or is the blue intended?

PR #2118, opened 2022-09-16
feat(client): disable credit route on mobile

ivanfalanga-adaptive:


src/client/src/Web/index.tsx
```diff
@@ -33,18 +35,20 @@ export const WebApp: React.FC = () => {
...
-          render={() => (
-            <>
-              <DisconnectionOverlay />
-              <Suspense fallback={<Loader />}>
-                <MainCreditRoute />
-              </Suspense>
-            </>
-          )}
-        />
+        {canDisplayCredit && (
```
algreasley, line 38: suggested a default path to 'catch' the user if a route is unavailable, like here

algreasley (approved):
default route idea is not blocking .. approved anyway

PR #2128, opened 2022-11-01
chore(client): update openfin scripts

algreasley:
Raised this to allow us to run openfin stuff from production builds as well as local dev server.
Particularly, this lets us fetch/inspect the manifests in the browser - with local server I was unable to do that (timed out) ..
I now know why this happens - thanks @bruceharris for the vite digging - can verify with local server using curl anyway, so not essential, but best to have the scripts there for completeness I think.


src/client/vite.config.ts
```diff
@@ -281,11 +302,24 @@ const setConfig = ({ mode }) => {
...
 
   if (TARGET === "openfin" || TARGET === "launcher") {
-    plugins.push(copyOpenfinPlugin(isDev, TARGET))
+    plugins.push(copyOpenfinPlugin(isDev, openfinBaseUrl, TARGET))
   }
 
+  let proxy
   if (process.env.VITE_MOCKS) {
     plugins.unshift(apiMockReplacerPlugin())
+  } else {
```
bruceharris, line 311: TS is complaining about the ambiguous typing of `proxy`. Perhaps we can improve this by removing the else clause and instead do something like
```
const proxy = process.env.VITE_MOCKS ? null : {
  ...
}
```

bruceharris (dismissed):
Approved with minor suggestion for improvement

src/client/vite.config.ts
algreasley: yep, that works .. as long as I use `undefined` as the fallback for when in mocks mode
(note to ourselves, mock mode is broken right now ... so this needs to be on our Tech Debt list)

PR #2129, opened 2022-11-08
fix(blotter): add CSS containment directive for blotter performance

algreasley:
minimal fix for ticket 3800 (iOS poor blotter performance with more than a few hundred rows)
* seems ok in testing, some concerns from @ivanfalanga-adaptive regarding cross-browser compatibility, which I share .. this is a HINT rather than a command, so is not a permanent fix.
* considering applying to other sections - this was the minimal change required to bring iOS browsers under control when blotter rows started exceeding the 500 mark (although we start to perceive degradation after only a few 100)
* this has been debugged a fair bit in Safari with attached device, and the layout phase is almost totally dominant when performance is bad .. hence the optimisations in the CSS .. this particular attribute chosen for cross-browser compatibility


bruceharris (approved):
👍

PR #2130, opened 2022-11-14
2850 Allow character multipliers for credit quantities

keshaldesai:


bruceharris (approved):
Very nice! Much cleaner with all that logic moved out of the React Components!

PR #2131, opened 2022-11-17
chore(client): openfin build script corrections for local mode handing

algreasley:
correcting the local-mode handling that went a bit too far in PR #2128


bruceharris (approved):
👍

PR #2133, opened 2022-11-21
chore(storybook): fix webpack vulnerabilities

algreasley:
started out as minor vuln-squash, then hours of wrestling with storybook config to get it to come up properly when running locally .. just to test
see also #2071 for historical reference of previous storybook-related vulnerability-squashing


bruceharris (dismissed):
👍

PR #2134, opened 2022-11-22
chore(test): update jest and associates

algreasley:
primary switch to Jest 27 (that is a big jump, in Jest terms, issues around fake timers etc.)

further upgrades were blocked by timing issues (specifically in the Tile, "Executing" -> too long) .. spent a while combing through jest release / upgrade material but I could find nothing to actually fake the Date handling in the RxJs schedulers underlying the timer-related behaviour of the 'too long' Tile timeout case.

Calling time on this as we are intending to upgrade to vitest, following successful testing in the RA project.


src/client/src/services/connection.ts
```diff
@@ -50,6 +50,7 @@ export const initConnection = async () => {
     url: `${window.location.origin}/ws`,
     interceptor: noop,
     autoReconnect: true,
+    useJson: true,
```
algreasley, line 53: oops - wasn't supposed to be checked in

algreasley: restored now

src/workspace/package.json
```diff
@@ -39,5 +40,10 @@
     "dialogflow": "^1.2.0",
     "openfin-notifications": "^1.11.0",
     "rxjs": "^7.5.2"
+  },
```
algreasley, line 43: this is one of my left-over bits from the last time I was working with the generated `TradingGateway.ts` which was the same across client / workspace but formatted differently so it was quite difficult to establish that they were in sync!

src/client/src/services/trades/__tests__/trades.test.ts
```diff
@@ -126,9 +126,8 @@ describe("trades", () => {
       const creditServiceMock = creditService as any
       creditServiceMock.__creditRfqsById(creditRfqs$)
     })
-    it("should be true", (done) => {
+    it("should convert RFQ and Quote details to trades stream", (done) => {
```
algreasley, line 129: fixing a goof from a few commits ago with the Credit sell-side number handling

src/client/jest.config.js
```diff
@@ -8,6 +8,22 @@ module.exports = {
   globals: {
     "ts-jest": {
       babelConfig: true,
+      diagnostics: {
+        ignoreCodes: [1343], // otherwise throws error about "import.meta" before babel can help
+      },
     },
   },
+  // for jest > 27 ..
```
algreasley, line 16: if switching to vitest works out, we can get rid of (all) this .. otherwise it is handy to know what to tweak for later Jests ..

bruceharris (approved):
👍

PR #2135, opened 2022-11-28
Chore/4146 remove husky hooks & remove more vulnerabilities

algreasley:
Idea was to remove husky entirely, but turns out that existing husky behaviour pollutes the .git/hooks folder and the hook will fail with an error until the developer removes the hooks manually.
The husky model is broken IMO .. currently a husky install at src/client level forces the generic husky-derived hooks to seek out one sub-directory only for the hook configuration .. upgrading to husky > 4 doesn't improve matters as that uses newer Git feature git.hooksPath to put hooks in <proj-root>/.husky - this stops the generic pollution as above, but the project git config is again set to look for hooks config in only one sub-directory.
Install husky at root, then, you say .. ok, but say goodbye to any workspace-specific husky config.
Removal of husky integration seems to be a manual job for _each_ developer in all cases.

The interim compromise, for now, is to remove the hooks from configuration.


bruceharris (dismissed):
👍

src/client/package.json
```diff
@@ -21,15 +21,16 @@
...
     "build": "cross-env TARGET=web vite build",
     "serve": "cross-env TARGET=web vite preview",
     "openfin:dev": "cross-env TARGET=openfin vite",
     "_openfin:run": "cross-env-shell \"wait-on -l $npm_config_manifest_url && openfin -l -c $npm_config_manifest_url\"",
-    "openfin:run:fx": "npm run _openfin:run --manifest_url=http://localhost:1917/dist/config/rt-fx.json",
-    "openfin:run:credit": "npm run _openfin:run --manifest_url=http://localhost:1917/dist/config/rt-credit.json",
-    "openfin:run:launcher": "npm run _openfin:run --manifest_url=http://localhost:1917/dist/config/launcher.json",
+    "openfin:run:fx": "npm run _openfin:run --manifest_url=http://localhost:1917/config/rt-fx.json",
+    "openfin:run:credit": "npm run _openfin:run --manifest_url=http://localhost:1917/config/rt-credit.json",
+    "openfin:run:launcher": "npm run _openfin:run --manifest_url=http://localhost:1917/config/launcher.json",
```
bruceharris, line 33: perhaps we can remove `openfin:run:preview` as it is now identical to `openfin:run:launcher`; we can update `openfin:start:preview` to use `openfin:run:launcher` instead

src/client/vite.config.ts
```diff
@@ -153,62 +156,50 @@ const customPreloadPlugin = () => {
   return result
 }
 
-const copyOpenfinPlugin = (dev: boolean) => {
+const copyPlugin = (
+  isDev: boolean,
+  buildTarget: "web" | "openfin" | "launcher",
```
bruceharris, line 161: 1. "launcher" shouldn't be a thing anymore, right?
1. type `BuildTarget` is defined on L17

bruceharris:
looks great at first pass! I will give a deeper look tomorrow morning

src/client/vite.config.ts
algreasley: dang .. well spotted

src/client/package.json
algreasley: good point .. that script does not make sense now .. the start:preview is probably still ok, but call the openfin:run:launcher instead to kick off openfin after build/serve

src/client/vite.config.ts
algreasley: sorted now, and I added `"finsemble"` as that is a genuine build target that I had not included (out of sight .. eh?)
but haven't used it for anything .. previously it seemed to just inherit the general stuff, including a bunch of web artefacts that I am sure it doesn't really need .. but I cannot test it easily right now ... so if we need to boot up Finsemble in future, we will need to verify some of this stuff, and possibly code-switching points (do we need any more or any less .finsemble.ts alternative implementations .. what about latest FDC3 etc. etc.
cc: @Clark-Adaptive

src/client/vite.config.ts
```diff
@@ -270,22 +261,18 @@ const setConfig: (env: ConfigEnv) => UserConfigExport = ({ mode }) => {
...
-  const plugins: any[] = [react()]
+  const plugins: any[] = [react()] // stays as any[] as WB injectManifest does not return PluginOption
 
   if (!isDev) {
-    plugins.push(customPreloadPlugin())
+    plugins.push(splitVendorChunkPlugin(), customPreloadPlugin())
   }
 
-  const TARGET = process.env.TARGET || "web"
+  const buildTarget: BuildTarget = (process.env.TARGET as BuildTarget) || "web"
```
bruceharris, line 271: very minor, orthogonal to the changes in this PR, and not a blocker at all, but FYI it seems the idiomatic way to apply certain plugins in build vs serve (e.g. development mode) is like this:
https://vitejs.dev/guide/api-plugin.html#conditional-application

bruceharris: actually probably not relevant, I realized it only applies to line 267, the other if statements would still be necessary

src/client/package.json
```diff
@@ -122,20 +123,20 @@
     "husky": "^4.3.8",
     "jest": "^27.0.0",
     "jest-styled-components": "^7.0.3",
-    "lint-staged": "^10.5.4",
+    "lint-staged": "^13.0.0",
```
bruceharris, line 126: does a git hook need to be configured to run lint-staged? if I understand correctly it is not being triggered now that we've removed husky.

algreasley:
> when I run `npm run openfin:run:launcher` the process crashes with
> 
> ```
> ❯ npm run openfin:run:launcher
> 
> > openfin:run:launcher
> > npm run _openfin:run --manifest_url=http://localhost:1917/config/launcher.json
> 
> 
> > _openfin:run
> > cross-env-shell "wait-on -l $npm_config_manifest_url && openfin -l -c $npm_config_manifest_url"
> 
> waiting for 1 resources: http://localhost:1917/config/launcher.json
> wait-on(99269) complete
> report usage data:  OK
> ReferenceError: config is not defined
> ```

this is the same old issue, right? Node 18 and the dns lookup ordering .... per Slack discussion

src/client/package.json
algreasley: I thought we had agreed that we are leaving hooks behind .. ??

src/client/vite.config.ts
algreasley: I see .. that is a bit buried in the plugin conf ... I prefer our existing approach, actually

src/client/vite.config.ts
```diff
@@ -153,62 +156,50 @@ const customPreloadPlugin = () => {
   return result
```
algreasley, line 156: we might be able to clean this up now (since 3.2.0):
https://github.com/vitejs/vite/blob/v3.2.5/packages/vite/CHANGELOG.md#buildmodulepreload-options

src/client/package.json
bruceharris: Yes, to clarify, my understanding is that lint-staged has no utility without git hooks, although perhaps I'm misunderstanding. Since I saw that we've upgraded lint-staged here, I'm looking to understand if/how it is being used as I don't see any hooks configuration.

bruceharris:
> > when I run `npm run openfin:run:launcher` the process crashes with
> > ```
> > ❯ npm run openfin:run:launcher
> > 
> > > openfin:run:launcher
> > > npm run _openfin:run --manifest_url=http://localhost:1917/config/launcher.json
> > 
> > 
> > > _openfin:run
> > > cross-env-shell "wait-on -l $npm_config_manifest_url && openfin -l -c $npm_config_manifest_url"
> > 
> > waiting for 1 resources: http://localhost:1917/config/launcher.json
> > wait-on(99269) complete
> > report usage data:  OK
> > ReferenceError: config is not defined
> > ```
> 
> this is the same old issue, right? Node 18 and the dns lookup ordering .... per Slack discussion
> 
> I added the Node 17+ advice in the README (I had stashed it and not unstashed)

Yes when I downgrade to Node 16 it all works

src/client/package.json
bruceharris: so if my understanding is correct, we can remove the `lint-staged` dependency and script command

algreasley: guess so, although nothing wrong with having it as a quick command to manually check before committing
but I am happy to remove it with the removal of the husky hooks if we just go with the approach of vscode + GH actions build for the branch

PR #2139, opened 2022-12-06
Fix credit snapshots

bruceharris:
This ensures that the OpenFin snapshots are segregated across Credit and FX. Without this configuration, `localStorage` is shared by both apps, thus they see each others snapshots which lead to undesirable layout state changes.

### DIFF
--- a/src/client/public-openfin/rt-fx.json
+++ b/src/client/public-openfin/rt-fx.json
@@ -13,7 +13,8 @@
     }
   },
   "runtime": {
-    "version": "17.85.53.10"
+    "version": "17.85.53.10",
+    "arguments": "--security-realm=rtfx"
   },
   "services": [
     {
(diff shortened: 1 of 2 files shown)


src/client/public-openfin/rt-fx.json
```diff
@@ -13,7 +13,8 @@
     }
   },
   "runtime": {
-    "version": "17.85.53.10"
+    "version": "17.85.53.10",
```
algreasley, line 16: how about changing the uuid/name ... I think that might be a more straightforward way of achieving distinct contexts .. security realm is usually about segregating runtimes / IABus traffic (I think .. it's been a while 😄)

bruceharris: I will test interop

bruceharris: good call! interop is in fact broken with the differing security realm. I'll work on an alternate fix.

PR #2140, opened 2022-12-07
fix: segregate OpenFin snapshots for credit vs FX...

bruceharris:
Local storage is shared by both apps, thus without this change, they see each other's snapshots, which lead to undesirable layout state changes. This commit appends the app name to the local storage keys to  ensure that the OpenFin snapshots are segregated across Credit and FX.

Interestingly, on Mac, local storage is segregated for the 2 apps, so this fix would seem to be unnecessary. On Windows, however, local storage is shared across apps that are served from the same origin, even if they have different UUIDs. Creating separate security realms for each app would create segregated local storage, but has the side effect of breaking interop with ReactiveAnalytics, so we did not pursue that approach.

### DIFF
--- a/src/client/public-openfin/rt-fx.json
+++ b/src/client/public-openfin/rt-fx.json
@@ -67,7 +67,7 @@
           "reloadIgnoringCache": true,
           "zoom": true
         },
-        "url": "<BASE_URL>/openfin-window-frame",
+        "url": "<BASE_URL>/openfin-window-frame?app=FX",
         "name": "Reactive-Trader-MAIN",
         "layout": {
           "settings": {
--- a/src/client/src/OpenFin/apps/Launcher/utils/openfin.ts
+++ b/src/client/src/OpenFin/apps/Launcher/utils/openfin.ts
@@ -1,5 +1,3 @@
-// TODO - Revisit if/when launcher is cross platform
-import { Bounds } from "openfin/_v2/shapes/shapes"
 import { useEffect } from "react"
 
 export async function getExistingOpenFinApplication(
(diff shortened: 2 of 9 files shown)


src/client/tsconfig.json
```diff
@@ -6,7 +6,7 @@
     },
     "target": "ESNext",
     "lib": ["DOM", "DOM.Iterable", "ESNext"],
-    "types": ["vite/client", "node", "openfin", "jest"],
+    "types": ["vite/client", "node", "jest"],
```
algreasley, line 9: ah, nice

src/client/src/OpenFin/apps/Launcher/utils/openfin.ts
```diff
@@ -61,7 +59,10 @@ export const getCurrentWindowBounds = async () => {
   return window.getBounds()
 }
 
-export async function animateCurrentWindowSize(bounds: Bounds, duration = 200) {
+export async function animateCurrentWindowSize(
+  bounds: OpenFin.Bounds,
```
algreasley, line 63: thanks .. ref prev (mega) PR

src/client/public-openfin/rt-fx.json
```diff
@@ -67,7 +67,7 @@
           "reloadIgnoringCache": true,
           "zoom": true
         },
-        "url": "<BASE_URL>/openfin-window-frame",
+        "url": "<BASE_URL>/openfin-window-frame?app=FX",
```
algreasley, line 70: I might be out of date, but we used to use a window.options attribute for things like this
It is not critical, but just floating it as a more "OpenFinny" way of passing context to an app/window

algreasley: for ref, this is what I was thinking of
https://developer.openfin.co/docs/javascript/stable/Window.html#~options
..
specifically the `customData` or perhaps `customContext` attributes

PR #2141, opened 2022-12-08
fix: credit sell side popup should show close icon

bruceharris:
**Before**:

Screen Shot 2022-12-08 at 12 36 53 PM

**After:**
Screen Shot 2022-12-08 at 12 31 21 PM

### DIFF
--- a/src/client/src/utils/window/openWindow.openfin.ts
+++ b/src/client/src/utils/window/openWindow.openfin.ts
@@ -45,7 +45,7 @@ export async function openWindow(
     (!hasChildWindows && !hasCoordinates) || config.center === "screen"
 
   const platform = await fin.Platform.getCurrent()
-  const options = {
+  const options: OpenFin.PlatformWindowCreationOptions = {
     autoShow: true,
     contextMenu: true,
     defaultCentered: centered,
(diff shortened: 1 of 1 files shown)


src/client/src/utils/window/openWindow.openfin.ts
```diff
@@ -71,7 +71,6 @@ export async function openWindow(
       content: [
         {
           type: "stack",
-          title: displayName || windowName,
```
bruceharris, line 74: see https://github.com/AdaptiveConsulting/ReactiveTraderCloud/blob/master/src/client/src/OpenFin/Window/ChildWindowFrame.tsx#L14-L15

not super intuitive, and not sure I understand the rationale behind the design of the API OF ChildWindowFrame, but this fix goes with the flow of how that component works today

algreasley: strange, but ok 👍🏻

algreasley: hmm - the ChildWindowFrame switching between windowName and windowTItle too .. not ideal, is it?

algreasley (approved):

PR #2142, opened 2022-12-19
chore(installer): update RT Mac installers

algreasley:
End result of this is:
- app.json deployed alongside rt-fx.json (and is identical) so existing RT installers will find their manifest
- the whole OpenFin build is deployed to launcher.xxxxx (identical bundle to what is deployed to openfin.xxxx so addresses are all in openfin bucket
(therefore, could optimise this by only uploading manifests, or arranging a redirect from launcher.xxx to openfin.xxx)


bruceharris (approved):
👍

PR #2146, opened 2023-01-13
End to end testing set up

snony14:
- [x] Install Playwright
- [x] Define the fixtures to override Playwright initial chromium lanucher in order to connect to RT openfin application
- [x] Amend devtools port in rt fx (Still researching on how to keep the port defaulted to 9090). **Obs:** There is a problem with connectivity on default devtools port of openfin chrome 9090.
- [x] add data test id to tile and buy sell button
- [x] define the test for spot tile to demonstrate how to test with openfin.
- [x] add jobs in workflow for end to end test

**Running playwright test against Chrome**
playwright_chrome

**Running playwright test against Openfin**
playwright_openfin


src/client/fixtures.ts
```diff
@@ -0,0 +1,143 @@
...
+  openfinAnalytics: Page
+}
+
+export const test = base.extend<IPlaywrightFixtures>({
+  browser: async ({}, use) => {
+    try {
+      const runtimeConnection = await chromium.connectOverCDP(RUNTIME_ADDRESS)
+      await use(runtimeConnection)
+    } catch (e) {
+      const foo = await chromium.launch()
```
snony14, line 24: rename

src/client/package.json
```diff
@@ -63,6 +65,7 @@
     "lodash": "^4.17.21",
     "openfin-adapter": "^27.71.21",
     "openfin-notifications": "^1.10.1",
+    "playwright": "^1.29.1",
```
algreasley, line 68: should be dev dep .. unless it is because we need in CI (as we do `npm ci` there before running e2e)
in which case, I would prefer that we did full npm install in CI, rather than label this a production dependency

src/client/package.json
```diff
@@ -41,6 +42,7 @@
     "openfin:start:launcher": "concurrently \"npm:openfin:dev\" \"npm:openfin:run:launcher\"",
     "openfin:build": "cross-env TARGET=openfin vite build",
     "openfin:serve": "cross-env TARGET=openfin vite preview",
+    "openfin:test":"playwright test --project=openfin",
```
algreasley, line 45: I think I would prefer `test:e2e:openfin` and then `test:e2e:web` above .. and any variations on that as we add more cases, like say `test:web:fx` and so on... we could even go with just `e2e:web` etc?

src/client/openfinGlobal.ts
```diff
@@ -0,0 +1,7 @@
+import { fin as MockFin } from 'openfin-adapter/src/mock';
```
algreasley, line 1: should not need this .. let's sync up later, I think we may need to swap notes between this and the OpenFin upgrade branch 😄

src/client/fixtures.ts
```diff
@@ -0,0 +1,102 @@
+import { test as base } from "@playwright/test"
```
algreasley, line 1: maybe this file should be in an `e2e` directory

src/client/.env
```diff
@@ -1,4 +1,5 @@
 VITE_VERSION=$npm_package_version
 VITE_NAME=$npm_package_name
 REACT_EDITOR=code
-EXTEND_ESLINT=true
\ No newline at end of file
+EXTEND_ESLINT=true
+URL_PATH=http://localhost:1917/
```
algreasley, line 5: I would not expect a localhost address to be in the general .env file (perhaps `.env.development` ?)

src/client/public-openfin/rt-fx.json
```diff
@@ -13,6 +13,7 @@
     }
   },
   "runtime": {
+    "arguments": "--remote-debugging-port=9090",
```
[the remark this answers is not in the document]
snony14: Connecting to openfin dev tools on 9090 is not working, all the applications are not served there except notifications (they are hidden), see [issue](https://bugs.chromium.org/p/chromium/issues/detail?id=1232509). I added that attribute to control the dev port in case I wanted to connect playwright with openfin application.

src/client/package.json
snony14: Sounds good

src/client/fixtures.ts
```diff
@@ -0,0 +1,102 @@
...
+import { Page, chromium } from "playwright"
+// ensures all window objects we interact with in our spec have fin tyepdefs
+export * from "./openfinGlobal"
+
+//
+type FXPage = "mainWindow" | "fx-tiles" | "fx-blotter" | "fx-analytics"
+
+const RUNTIME_ADDRESS = "http://localhost:9090"
```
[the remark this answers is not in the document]
snony14: You are absolutely right 😄 , currently openfin chrome cannot be connected over 9090 due to some security issue, but changing the devtools, that works.

.github/workflows/pull.yml
```diff
@@ -147,3 +147,23 @@ jobs:
...
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+
+      - name: Install
+        run: npm ci
+
+      - name: Test
+        env:
+          URL_PATH: https://web.env.reactivetrader.com/pull/${{ github.event.number }}
```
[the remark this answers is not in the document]
snony14: I will do this later on another Pr branched from this Pr (credit tests)

Clark-Adaptive:
I checked out the pr and tried running the e2e scripts and ran into some errors.
I am on Mac.


Screen Shot 2023-01-27 at 3 10 08 PM

Screen Shot 2023-01-27 at 3 10 08 PM

Steps I took:
1. gh pr checkout 2146
2. cd src/client
3. npm install
4. npm run e2e:web
5. npm run e2e:openfin

Let me know if I missed something.

snony14:
> 

Thanks for pointing that out @Clark-Adaptive, you should be running the application (`npm start`) before running `npm run e2e:web`. And you should be running openfin fx before running `npm run e2e:openfin`.

snony14:
@algreasley , for openfin e2e we need to change the dev tool port to 9091.

src/client/openfinGlobal.ts
algreasley: we'll come back to this when I merge in the OF upgrade

src/client/public-openfin/rt-fx.json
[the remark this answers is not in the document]
algreasley: still a bit confused about whether we should be settling on 9091 as the "standard" debug port for OpenFin, at least for the reactive ecosystem platform manifests (I realise the notifications manifest currently has 9090 burned into it)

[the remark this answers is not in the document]
algreasley: also, given we have added it to `rt-fx.json`, should it also be added as standard to all manifests (inc `rt-credit.json` and `launcher.json`)

PR #2149, opened 2023-01-31
E2e/4650 test credit

snony14:
- [x] Update data test ids where necessary
- [x] Remove unused URL Path
- [x] Increase test timeout
- [x] Change openfin dev tool port
- [x] Define pages for openfin credit
- [x] Define test for Credit
- [x] Update readme


src/client/e2e/fixtures.ts
```diff
@@ -0,0 +1,146 @@
...
+        `Unexepcted Context(s) for Notification: Expected 1, Found ${contexts.length}`,
+      )
+    }
+
+    const pages = await contexts[0].pages()
+
+    const notificationPage = pages.find(
+      (page) =>
+        page.url() ===
+        "https://cdn.openfin.co/services/openfin/notifications/0.12.10/provider.html",
```
algreasley, line 140: as we saw today, this is dependent on the nature of the manifest in question .. can we extract into env do you think?
we could do the same with the localhost:9090 OpenFin debug address above too, perhaps

src/client/e2e/spot-tile.spec.ts
```diff
@@ -0,0 +1,47 @@
...
+        currentWindow.maximize()
+        return window.fin
+      })
+      tilePage = fxOpenfinPagesRec["fx-tiles"]
+      blotterPage = fxOpenfinPagesRec["fx-blotter"]
+    } else {
+      const pages = await context.pages()
+      const mainWindow = pages.length > 0 ? pages[0] : await context.newPage()
+      await mainWindow.goto(
+        `${process.env.URL_PATH ?? "http://localhost:1917/"}`,
```
algreasley, line 29: dang it, sorry, changed my mind on this .. can we add back into `.env.development` and just use URL_PATH in here?

algreasley (approved):
Nice work, Moise

PR #2151, opened 2023-02-01
Schedule smoke test against prod every 10 min (Test)

snony14:


.github/workflows/schedule.yml
```diff
@@ -0,0 +1,28 @@
+name: Runs e2e every day at 2pm
```
algreasley, line 1: This is great .. so simple
(name does not match reality tho 😛 )
Also, I think we should canvass opinion on RT channel .. I don't want this to get in the way of demos, so I think we may want to run it out-of-hours ... 
Could even co-ordinate this with support, who have the responsibility right now of ensuring the RT/RA are running
(ooh, that has given me an idea about some further work for our expanding team 😄)

.github/workflows/schedule.yml
```diff
@@ -0,0 +1,28 @@
...
+  schedule:
+    - cron: "*/10 * * * *"
+
+defaults:
+  run:
+    working-directory: src/client
+
+jobs:
+  e2e:
+    name: End-to-end test
```
algreasley, line 12: mention RT Prod?

.github/workflows/schedule.yml
snony14: This is just a sample test  to see if it is going to work for every 10 min😅 . If that's the case, then I will update the schedule.yml and have the cron job run every day at 2pm 😃

.github/workflows/schedule.yml
```diff
@@ -0,0 +1,28 @@
+name: Runs e2e every day at 2pm against RT Prod
+on:
+  schedule:
+    - cron: "*/10 * * * *"
```
algreasley, line 4: as disc, let's run this at 9am to start with, and I think aim it at Dev system until we are happy with the tests and the timing

.github/workflows/schedule.yml
algreasley: ok, def switch to Dev instance then 😅

algreasley (approved):

PR #2152, opened 2023-02-02
E2e/4623 spot tile test

snony14:
* [x] Define test for spot tile

### DIFF
--- a/src/client/e2e/spot-tile.spec.ts
+++ b/src/client/e2e/spot-tile.spec.ts
@@ -112,6 +112,18 @@ test.describe("Spot Tile", () => {
         .inputValue()
       expect(notionalValue).toEqual("1,000,000")
     })
+
+    test("When I enter a number too large (over 1,000,000,000) then an error will appear 'Max exceeded'", async () => {
+      await tilePage.locator("input[id='notional-input-EURUSD']").clear()
+      await tilePage
+        .locator("input[id='notional-input-EURUSD']")
+        .type("1200000000")
+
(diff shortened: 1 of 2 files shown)


src/client/playwright.config.ts
```diff
@@ -17,6 +17,11 @@ const config: PlaywrightTestConfig = {
       name: "openfin",
     },
   ],
+  // use: {
```
algreasley, line 20: useful, thanks 👍🏻

src/client/e2e/spot-tile.spec.ts
```diff
@@ -2,41 +2,182 @@ import { expect, Page } from "@playwright/test"
...
       const mainWindow = pages.length > 0 ? pages[0] : await context.newPage()
+
       await mainWindow.goto(`${process.env.URL_PATH}`)
+
       tilePage = mainWindow
       blotterPage = mainWindow
     }
+  })
+
+  test.describe("Valid Purchase", () => {
```
algreasley, line 30: In time, I think we will need to find a way to split these into _smoke_ tests and _regression_ tests ..
When I worked with Selenium way back in the dark(er) ages .. there was a way of 'tagging' the tests so one could run them by tag instead of all-or-nothing, so a select few were tagged as `@smoke` and we managed the different types of test run that way ...

algreasley (approved):

src/client/e2e/spot-tile.spec.ts
snony14: Cool, I had no idea. It is good to split them by tagging. I will investigate how to do that with playwright.

PR #2160, opened 2023-03-14
Smoke test annotations, OpenFin test fixes

SHolleworth:
Added annotations to 2 tests for a minimal smoke test set.

Fixed some issues with the OpenFin tests.

Tidied up some tests and env files.

### DIFF
--- a/src/client/e2e/credit.spec.ts
+++ b/src/client/e2e/credit.spec.ts
@@ -3,26 +3,24 @@ import { test } from "./fixtures"
 
 test.describe("Credit", () => {
   test.describe("New RFQ", () => {
-    test("When I select Googl instrument and click Send RFQ button then I should see a GOOGL RFQ created on the RFQ sections and I can accept any value", async ({
+    test("When I select Googl instrument and click Send RFQ button then I should see a GOOGL RFQ created on the RFQ sections and I can accept any value @smoke", async ({
       context,
-      creditOpenfinPagesRec,
+      creditPagesRec,
     }, testInfo) => {
       test.setTimeout(120000)
       let newRfqPage: Page
       let rfqsPage: Page
       let rfqBlotterPage: Page
       if (testInfo.project.name === "openfin") {
-        const mainWindow = creditOpenfinPagesRec["mainWindow"]
+        const mainWindow = creditPagesRec["mainWindow"]
         await mainWindow.evaluate(async () => {
-          const currentWindow = window.fin.desktop.Window.getCurrent()
-          currentWindow.maximize()
-          return window.fin
+          window.fin.Window.getCurrentSync().maximize()
         })
-        newRfqPage = creditOpenfinPagesRec["credit-new-rfq"]
-        rfqsPage = creditOpenfinPagesRec["credit-rfqs"]
-        rfqBlotterPage = creditOpenfinPagesRec["credit-blotter"]
+        newRfqPage = creditPagesRec["credit-new-rfq"]
+        rfqsPage = creditPagesRec["credit-rfqs"]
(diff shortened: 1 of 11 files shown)


src/client/e2e/credit.spec.ts
```diff
@@ -3,7 +3,7 @@ import { test } from "./fixtures"
 
 test.describe("Credit", () => {
   test.describe("New RFQ", () => {
-    test("When I select Googl instrument and click Send RFQ button then I should see a GOOGL RFQ created on the RFQ sections and I can accept any value", async ({
+    test("When I select Googl instrument and click Send RFQ button then I should see a GOOGL RFQ created on the RFQ sections and I can accept any value @smoke", async ({
```
algreasley, line 6: _to discuss:_ should we put tags at the beginning or end .. end makes more sense as it doesn't mess up your reading of the purpose of the test, but I fear it is a little _invisible_ ..
it is a pity we do not have decorators yet, perhaps we could if we jumped on TS 5 ?

src/client/e2e/fixtures.ts
```diff
@@ -58,18 +59,18 @@ const urlPathToCreditPage = (path: string): CreditPage => {
 }
 
 export const test = base.extend<IPlaywrightFixtures>({
-  browser: async ({}, use) => {
-    try {
+  browser: async ({}, use, workerInfo) => {
+    if (workerInfo.project.name === "openfin") {
```
algreasley, line 63: perhaps use a little util, so we can isolate the project ID for OpenFin runs

src/client/e2e/fixtures.ts
```diff
@@ -125,25 +126,27 @@ export const test = base.extend<IPlaywrightFixtures>({
...
+      const runtimeConnection = await chromium.connectOverCDP(
+        `${process.env.NOTIFICATION_RUNTIME_ADDRESS}`,
       )
-    }
 
-    const pages = await contexts[0].pages()
+      const contexts = runtimeConnection.contexts()
+      if (contexts.length !== 1) {
+        throw Error(
+          `Unexepcted Context(s) for Notification: Expected 1, Found ${contexts.length}`,
```
algreasley, line 138: little typo (inherited) ..

src/client/e2e/test-data/blotter-data.csv
algreasley, line 1: why do we need to check these files in?

src/client/public-openfin/launcher.json
```diff
@@ -21,7 +21,7 @@
     }
   },
   "runtime": {
-    "version": "17.85.53.10",
+    "version": "22.94.67.8",
```
algreasley, line 24: don't check in please .. affects the main app, which is not tested on this version .. will get increased to V28/29 by OF upgrade very soon anyway

algreasley (commented):
nice, @SHolleworth  - good bit of `await`ing tidy up - tests much cleaner now

src/client/e2e/test-data/blotter-data.csv
SHolleworth: I did wonder if we needed to check this in, removing

src/client/e2e/credit.spec.ts
SHolleworth: What kind of issues do you think less visibility could create?

algreasley: not important ... let's leave it and I'll hope that Decorators will be the way forward, and I can annotate the tests like I used to in the past (was Java/Selenium not JS)

algreasley (approved):
nice .. looking good!

PR #2163, opened 2023-03-14
Chore/4545 openfin pop in

Clark-Adaptive:
implemented pop-in functionality on openfin.

Notes:
1. I manually upgraded Openfin to 22 for this feature, and seeing as @algreasley is upgrading Openfin at the moment and changing titles and names, there will need to be some revision on my end before this PR can be approved.
3. This solution uses pre-made layouts based on how Openfin handles pop out. Thus, if the user were to pop a section out and move the tabs around in the main window, when the section gets popped in it will restore the main layout. Personally, since there are so many ways the user can rearrange the sections, I believe the slight adjustment to the main layout is fine, but we'll see what everyone says, I can refine it if need be.
4. If the user pops out the tiles section, then pops out one of the tiles, then pops the tiles section back in to the main window BEFORE closing the popped out tile, it resets the tearOutState, and has no recollection that a tile has been popped out. This is not a huge issue, it just creates a duplicate of that tile, if you close that tile everything will go back to normal.
5. There are still lots of debugging console logs, I kept them in for now just in case someone discovers a bug, I intend to erase them after reviewers greenlight everything.


src/client/src/App/LiveRates/Tile/Header/TileHeader.tsx
```diff
@@ -93,6 +93,7 @@ export const HeaderInner = forwardRef<HTMLDivElement, HeaderProps>(
 export const Header = () => {
   const ref = useRef<HTMLDivElement>(null)
   const { currencyPair, isTornOut, supportsTearOut } = useTileContext()
+  console.log("tileContext: ", useTileContext())
```
SHolleworth, line 96: Left by mistake?

src/client/src/App/LiveRates/Tiles.tsx
```diff
@@ -21,7 +21,7 @@ export const [useFilteredCurrencyPairs, filteredCurrencyPairs$] = bind(
   combineLatest([currencyPairs$, selectedCurrency$, tearOutState$]).pipe(
     map(([currencyPairs, selectedCurrency, tearOutState]) => {
       const result = { ...currencyPairs }
-
+      console.log("tearoutState: ", tearOutState)
```
SHolleworth, line 24: Left by mistake?

src/client/src/App/LiveRates/Tiles.tsx
```diff
@@ -54,6 +54,12 @@ export const Tiles = () => {
   const tearOutEntry = useTearOutEntry()
 
   useEffect(() => {
+    console.log(
```
SHolleworth, line 57: Left by mistake

src/client/src/OpenFin/utils/tileWindowNames.ts
```diff
@@ -0,0 +1,12 @@
+const tileWindowNames = new Set()
+tileWindowNames.add("EURUSD")
```
SHolleworth, line 2: This manual method of adding the windows is super brittle and will cause some hard to figure out bugs for anyone who adds another currency pair in the future. A better method might be to generate tileWindowNames from the currencyPair stream or something else that will update automatically with the backend.

src/client/src/OpenFin/utils/poppedOutSections.ts
```diff
@@ -0,0 +1,16 @@
+export let tilesPoppedOut = false
```
SHolleworth, line 1: Is this way of managing state recommened for OpenFin? If not I think this could do with a refacor, maybe using Contexts or Rx to get the state where it needs to be

src/client/src/OpenFin/utils/layoutTemplates.ts
```diff
@@ -0,0 +1,338 @@
+import { ROUTES_CONFIG } from "@/constants"
+import { constructUrl } from "@/utils/url"
+
+export const mainLayout = {
+  settings: {
```
SHolleworth, line 5: You could store these settings in a variable and share it over all the layouts since they all seem to be the same.

src/client/src/OpenFin/utils/tileWindowNames.ts
algreasley: it actually exposes a bug in the window close logic (been there a while by the looks, and I didn't spot it either) so I'll rename these to "Reactive-Trader-FX-Spot-Tile--CCYCCY" in the upgrade branch and close these too when the main RT close is clicked

Clark-Adaptive: > we should give the tile windows a better name than this, anyway, to identify each one as a Tile, with the currency suffixed, then we should not care about enumerating the currencies ..

I do agree the names could be better, they should have "tile" in the name. Right, then I can filter by a regex and won't need to touch the currencyPair observable at all.

src/client/src/OpenFin/utils/layoutTemplates.ts
SHolleworth: `const createStack = (overrides: Partial<StackOptions> //an example type for the stack options) => ({
  //default options,
...overrides
})` Could work to deal with any variance in the objects

src/client/src/OpenFin/Window/WindowViewport.tsx
```diff
@@ -24,19 +42,81 @@ const getEmptyContent = (key: LayoutKey, useIcon = true) => {
   return key
 }
 
+const closeTileWindows = async () => {
+  const childWindows = await fin.Application.getCurrentSync().getChildWindows()
+  const childWindowNames = childWindows.map((window) => window.identity.name)
```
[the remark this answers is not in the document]
Clark-Adaptive: will chain in next commit.

src/client/src/OpenFin/Window/WindowViewport.tsx
```diff
@@ -24,19 +42,81 @@ const getEmptyContent = (key: LayoutKey, useIcon = true) => {
...
+}
+
 export const WindowViewport = ({ children }: WithChildren) => {
   //TODO: Remove this HACK once OpenFin exposes content of "empty" layout containers...
   useEffect(() => {
     // eslint-disable-next-line @typescript-eslint/ban-ts-comment
     //@ts-ignore
     if (!window.fin.me.isView) {
-      const listenerViewAttached = (e: any) => {
+      const listenerViewAttached = async (e: any) => {
```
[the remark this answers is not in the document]
Clark-Adaptive: right whoops, good catch.

src/client/public-openfin/rt-fx.json
```diff
@@ -74,9 +75,9 @@
             "blockedPopoutsThrowError": false,
             "closePopoutsOnUnload": true,
             "constrainDragToContainer": false,
-            "hasHeaders": true,
+            "hasHeaders": false,
```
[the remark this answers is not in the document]
Clark-Adaptive: yes we do, let me get on that..

src/client/src/OpenFin/utils/layoutTemplates.ts
Clark-Adaptive: That makes sense, I will be making a common createStack() function and pass in width, height, and content as parameters.

src/client/src/OpenFin/utils/poppedOutSections.ts
Clark-Adaptive: Yes, the max number of popped out sections should be 2. However, since I'm using view attached to track when sections get popped out, it gets called 3 times on initialization, and when a view gets attached I decrement numPoppedOutSections. And the only reason I am using the viewAttached listener over windowCreated is because we are using the openfin popout button, which automatically assigns "internally-generated-..." as the window name, so I won't know which window was popped out in the windowCreated listener, otherwise I would not need to set the initial value of numPoppedOutSections as 3. Since we know main window and snapshots and contacts are the only windows created on initialization.

Clark-Adaptive: So yes, if in the openfin upgrade, we are given control over naming the window created from openfin's popout, I will refactor to use windowCreated and this will all make much more sense.

algreasley (approved):

PR #2165, opened 2023-03-22
feat: virtualised trades blotter

SHolleworth:
Virtualised the trades blotter with react-window to increase performance when there are a large number of rows.


src/client/e2e/spot-tile.spec.ts
```diff
@@ -47,10 +48,10 @@ test.describe("Spot Tile", () => {
     test("When I buy USD/JPY then an Execution screen appears until transaction is performed and tile displays green confirmation screen", async () => {
       await tilePage.locator("[data-testid='Buy-USDJPY']").click()
 
-      const executingSpinner = await tilePage.getByText(/Executing/)
+      const executingSpinner = tilePage.getByText(/Executing/)
```
algreasley, line 51: good catch here .. there are so many awaits in the Playwright docs that it must be tempting to slap it on everything.
but as you have called out here - these are returning Locator objects, not Promises .. it is the method on the Locator that is typically async .. e.g. click()

src/client/jest.config.js
```diff
@@ -4,7 +4,7 @@ module.exports = {
   moduleNameMapper: {
     "^@/(.*)$": "<rootDir>/src/$1",
   },
-  testPathIgnorePatterns: ["<rootDir>/e2e/"],
+  testPathIgnorePatterns: ["<rootDir>/e2e/", "utils.ts"],
```
algreasley, line 7: losing `jest` soon, so this may be irrelevant, but shouldn't jest just be picking up spec/test files .. and stuff in __tests__, so I would argue `utils.ts` does not belong in `__tests__` as it isn't a test ?
.. bit of a minor point, I'd agree .. suggest a `testUtils` folder, if we still need it post-migration to `vitest`

src/client/src/App/Trades/TradesGrid/TradesGrid.tsx
```diff
@@ -1,60 +1,76 @@
...
 const highlightBackgroundColor = css`
   animation: ${({ theme }) => theme.flash} 1s ease-in-out 3;
 `
 
-const TableBodyRow = styled.tr<{ pending?: boolean; highlight?: boolean }>`
-  &:nth-child(even) {
-    background-color: ${({ theme }) => theme.core.darkBackground};
-  }
+const TableBodyRow = styled.div<{
+  pending?: boolean
```
algreasley, line 46: wot's pending for?

src/client/src/App/Trades/TradesGrid/TradesGrid.tsx
```diff
@@ -1,60 +1,76 @@
...
   }
-  height: 2rem;
   ${({ highlight }) => highlight && highlightBackgroundColor}
 `
 
-const TableBodyCell = styled.td<{
+const TableBodyCell = styled.div<{
   align?: "right" | "left"
   crossed?: boolean
+  pending?: boolean
```
algreasley, line 64: same - pending and highlight?

src/client/src/App/Trades/TradesGrid/TradesGrid.tsx
```diff
@@ -76,19 +93,153 @@ const StatusIndicator = styled.td<{ status?: TradeStatus | QuoteState }>`
...
-  onRowClick?: (row: Row) => void
-  isRowCrossed?: (row: Row) => boolean
-  caption: string
+const TableBackground = styled.div`
+  min-width: 60rem;
+`
+
+const BlotterContext = createContext("Caption not loaded")
+
+const innerElementType = forwardRef<
```
algreasley, line 107: minor: as these are components, I'd capitalise the names

src/client/src/App/Trades/__tests__/utils.ts
```diff
@@ -0,0 +1,34 @@
+export const setupWindow = () => {
```
algreasley, line 1: this is a mock, I guess .. can name / locate accordingly

src/client/jest.config.js
SHolleworth: Yeah, this felt weird as I was doing it, I'll move it to a different folder

src/client/src/App/Trades/TradesGrid/TradesGrid.tsx
SHolleworth: It's for keeping senior devs reviewing skills sharp

src/client/tsconfig.json
```diff
@@ -9,7 +9,7 @@
     "types": ["vite/client", "node", "jest"],
     "allowJs": false,
     "skipLibCheck": true,
-    "esModuleInterop": false,
+    "esModuleInterop": true,
```
algreasley, line 12: just need reminder for why we have switched this ... there are consequences, with mitigations, as it will create a lot of boilerplate code in our generated js

src/client/src/App/Trades/TradesGrid/utils.ts
```diff
@@ -0,0 +1,9 @@
+//Using because react-window seems to warp flex values in its items
+export const getWidthPercentage = (widths: number[], width: number) =>
+  width / (widths.reduce((acc, curr) => acc + curr) / 100)
+
+export function convertRemToPixels(rem: number) {
+  return process.env.NODE_ENV !== "test"
```
algreasley, line 6: as disc, not happy with test code in main code .. suggested move test impl to module in __mocks__

src/client/tsconfig.json
SHolleworth: jest wasn't able to recognise a default import unless this setting was switched on, I will make a note to revisit it after the conversion to vitest

algreasley (approved):

PR #2171, opened 2023-03-30
chore: schedule full set of e2e tests for weekdays

SHolleworth:


.github/workflows/e2e.yml
```diff
@@ -0,0 +1,28 @@
+name: Scheduled E2E
```
algreasley, line 1: need a different name .. this one is used to post the results of smoke below -> Slack -support channel
we should do the same for this one -> -dev channel (@Dev-Alex23 can show you)

algreasley (approved):

PR #2175, opened 2023-04-03
chore: upgrade openfin-cli to 4.0.0

SHolleworth:


algreasley (dismissed):
This works fine locally for me (Mac) .. fixes a number of vulnerabilities too, which is nice.
I have a few extra vulnerability fixes (per npm audit) lined up, so could add to this, or just raise a separate PR after this and the SB 7 one is in.
Note that OpenFin startup still fails on Node 18 (on Mac) .. will test again with this later CLI when we are on vite 4 .. also Node 20 coming v soon.

algreasley (approved):

PR #2178, opened 2023-04-12
chore(workspace): update app addresses and descriptions, add RT Credit

algreasley:
simple update of addresses to be in line with RT FX/Credit split of last year - also amended naming to allow for Credit views to be added later .. need to alter the main Credit app first so RFQ Tiles can be a standalone section.

Note that on Mac, shutting this thing down is a pain, so you need to run the OF Process Manager
```
npx openfin -l -c https://cdn.openfin.co/release/apps/openfin/processmanager/app.json
```
and kill off all the parts, starting with Notification Service, as that will really shout at you if you leave it until last (if you have started up RT)


src/workspace/config/fx-analytics.json
```diff
@@ -0,0 +1,3 @@
+{
```
algreasley, line 1: these aren't brand new, but git got confused about renaming (I only added 3 chars!)

src/workspace/src/apps.ts
```diff
@@ -1,151 +1,166 @@
...
+  images: [{ src: `${BASE_URL}/images/previews/reactive-trader.PNG` }],
+}
 
-export const reactiveTrader: App = {
-  appId: 'reactive-trader',
-  title: 'Reactive Trader',
-  manifestType: 'manifest',
-  description: 'Desktop Reactive Trader',
-  manifest: reactiveTraderManifestUrl,
+export const reactiveTraderCredit: App = {
```
algreasley, line 34: just added the full Credit app to start with .. the RT Credit app needs a little work to separate out the RFQ Tiles and then we can add those 3 as Views (see below)

src/workspace/src/store.ts
```diff
@@ -5,36 +5,37 @@ import {
   StorefrontNavigationSection,
```
algreasley, line 5: this is mostly formatting updates .. amongst this are small renaming, and adding RT Credit to the Apps list

Dev-Alex23 (approved):

PR #2179, opened 2023-04-13
fix(openfin): hack fix for openfin notifications - blotter highlight

algreasley:
non-urgent, but just saving progress to date .. real fix involves sorting out string/number dichotomy in handling tradeId attribute .......


SHolleworth:
Can we close this one now?

PR #2180, opened 2023-04-13
fix(workspace): correct RT addresses in Trading Workspace

algreasley:


src/workspace/config/snapshot.json
```diff
@@ -327,7 +327,7 @@
                                 },
                                 "webAPIs": ["notifications", "audio", "video"]
                               },
-                              "url": "<RT_URL>/tiles",
+                              "url": "<RT_URL>/fx-fx-tiles",
```
SHolleworth, line 330: Looks like a mistake `fx-fx`

algreasley: yes, damn!

algreasley: weird that it worked (I did test it !!) .. that section must not be active

algreasley: fixed now

SHolleworth (approved):

PR #2183, opened 2023-04-14
fix: Flash of unstyled Text (FOUT)

Dev-Alex23:
The issue where the PWA banner style is not applying immediately on load seem to be more widespread than initially thought & has been there for some time.

This PR removes the deprecated `typeface-lato, typeface-roboto & typeface-montserrat` and replaced it with Google's font CDN. The issue seems to be that when a page is first requested, the font is queued & downloaded at the tail end of the request. The default behaviour of `@font-face` is that the `font-display` is set to `swap`. This means that if the font is not available, the browser user agent will apply a font of its choice until ours is available, which explains why we see a Flash of Unstyled Text (FOUT). 

The fix is to add to preloading mechanism `<link rel="preconnect" href="https://fonts.gstatic.com/" crossorigin>` or enable non-blocking renderer `<link rel="preload" href="xxx" as="style" onload="this.rel='stylesheet'">`. With this and setting the `font-display: block` the font will not show until it is available and because of the `preconnect` & `preload` the font download is queued much earlier in the render cycle.

https://user-images.githubusercontent.com/52295529/232049957-547f88bd-fab6-46d8-afe4-96dddec7711e.mov

### DIFF
--- a/src/client/package.json
+++ b/src/client/package.json
@@ -81,10 +81,7 @@
     "react-window": "^1.8.8",
     "ress": "^3.0.0",
     "rxjs": "^7.5.4",
-    "styled-components": "^5.2.2",
-    "typeface-lato": "^1.1.13",
-    "typeface-montserrat": "^1.1.13",
-    "typeface-roboto": "^1.1.13"
+    "styled-components": "^5.2.2"
   },
   "devDependencies": {
     "@babel/core": "^7.13.16",
@@ -136,6 +133,7 @@
     "storybook": "^7.0.2",
     "text-encoding": "^0.7.0",
     "typescript": "^4.9.3",
+    "unplugin-fonts": "^1.0.0",
     "vite": "^4.2.1",
     "vite-plugin-html": "^3.2.0",
     "vite-plugin-static-copy": "^0.13.1",
--- a/src/client/vite.config.ts
+++ b/src/client/vite.config.ts
@@ -3,10 +3,12 @@ import { readdirSync, statSync } from "fs"
 import path, { resolve } from "path"
 import modulepreload from "rollup-plugin-modulepreload"
 import { injectManifest } from "rollup-plugin-workbox"
+import Unfonts from "unplugin-fonts/vite"
 import {
   ConfigEnv,
   loadEnv,
   Plugin,
+  PluginOption,
   splitVendorChunkPlugin,
   UserConfigExport,
 } from "vite"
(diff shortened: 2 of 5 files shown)


src/client/vite.config.ts
```diff
@@ -302,6 +303,32 @@ const setConfig: (env: ConfigEnv) => UserConfigExport = ({ mode }) => {
...
+          defer: false,
+          styles:
+            "ital,wght@0,100;0,300;0,400;0,500;0,700;0,900;1,100;1,300;1,400;1,500;1,700;1,900",
+        },
+        {
+          name: "Montserrat",
+          defer: false,
+          styles:
+            "ital,wght@0,100;0,200;0,300;0,400;0,500;0,600;0,700;0,800;0,900;1,100;1,200;1,300;1,400;1,500;1,600;1,700;1,800;1,900",
+        },
```
Dev-Alex23, line 326: possibility that we dont need all these weights, but include them just to make sure, the ones we are sure we don not need can be removed after, if we go with this solution

src/client/vite.config.ts
```diff
@@ -326,7 +353,7 @@ const setConfig: (env: ConfigEnv) => UserConfigExport = ({ mode }) => {
         },
       ],
     },
-    plugins: isStorybook ? [] : [...plugins],
+    plugins: isStorybook ? [fontfacePreLoad] : [...plugins, fontfacePreLoad],
```
algreasley, line 356: bump the logic up outside of `defineConfig()` now, I think
this should be a line like `plugins.push(fontfacePreLoad)` then it is much clearer that it applies in all cases
.. just leaving this
```
    plugins,
```

src/client/package.json
```diff
@@ -81,10 +81,7 @@
     "react-window": "^1.8.8",
     "ress": "^3.0.0",
     "rxjs": "^7.5.4",
-    "styled-components": "^5.2.2",
-    "typeface-lato": "^1.1.13",
-    "typeface-montserrat": "^1.1.13",
-    "typeface-roboto": "^1.1.13"
+    "styled-components": "^5.2.2"
```
algreasley, line 84: idle thought: we should look at later [styled-components](https://www.npmjs.com/package/styled-components/v/6.0.0-beta.15)

src/client/vite.config.ts
algreasley: what is the output of this plugin (in the build) ?
i.e. is it in the correct order in plugins array?

algreasley: I take it you have tried production build and `npm run serve`

algreasley (commented):
looking good for me, locally 😄

SHolleworth (approved):
Looks good to me 👍

PR #2184, opened 2023-04-14
chore(workspace): make RA URL sync up with the env in the build

algreasley:


src/workspace/.env.development
```diff
@@ -1,6 +1,4 @@
 # Reactive Trader url
 VITE_RT_URL=http://localhost:1917
-# Reactive Analytics url
```
algreasley, line 3: based on env === "local" instead .. we should be able to run locally off development or production builds!

src/workspace/.env.production
algreasley, line 1: nothing left in here after moving RA_URL processing to vite.config.ts

src/workspace/src/home/utils.ts
algreasley, line 1: this is just formatting, I think after playing with types the other day, @SHolleworth .. will be overwritten by yours anyway ..

src/workspace/src/home/utils.ts
```diff
@@ -51,18 +56,20 @@ const mapAppEntriesToSearchEntries = (apps: App[]): CLISearchResult<any>[] =>
     return entry
   })
 
-const mapPageEntriesToSearchEntries = (pages: Page[]): CLISearchResult<any>[] =>
-  pages.map(page => {
+const mapPageEntriesToSearchEntries = (
+  pages: Page[],
+): CLISearchResult<Action>[] =>
```
algreasley, line 61: ah, this is the only change, I think - it's generally beneficial, I think

src/workspace/vite.config.ts
```diff
@@ -12,6 +12,13 @@ function getBaseUrl(dev: boolean) {
 const copyOpenfinPlugin = (isDev: boolean): Plugin[] => {
   const env = process.env.ENVIRONMENT || "local"
 
+  const reactiveAnalyticsUrl =
+    env === "local"
```
algreasley, line 16: bit fiddly, but need the local address to be there for local "development" and "production" builds
the usual dance around prod -> demo makes the rest of it unnecessarily messy ..

PR #2188, opened 2023-04-20
fix: fx notification button to highlight blotter

SHolleworth:
The highlight row signal was emitting the tradeId as a string which failed against a strict equality check. Comparison was between emitted tradeId (string) and backend tradeId (number).

### DIFF
--- a/src/client/src/services/trades/__mocks__/trades.mock.ts
+++ b/src/client/src/services/trades/__mocks__/trades.mock.ts
@@ -45,7 +45,7 @@ export const mockRawTrades = {
 
 export const mockTrades: FxTrade[] = [
   {
-    tradeId: "1111111111",
+    tradeId: 1111111111,
     traderName: "LMO",
     symbol: "GBPUSD",
     notional: 1_000_000,
@@ -57,7 +57,7 @@ export const mockTrades: FxTrade[] = [
     status: TradeStatus.Done,
   },
   {
-    tradeId: "2222222222",
+    tradeId: 2222222222,
     traderName: "LMO",
     symbol: "USDJPY",
     notional: 1_000_000,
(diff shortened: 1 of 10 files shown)


src/client/src/services/trades/types.ts
```diff
@@ -49,6 +49,6 @@ export interface CreditTrade extends Trade {
 }
 
 export interface Trade {
-  tradeId: string
+  tradeId: bigint
```
algreasley, line 52: have to follow through the workings, but can we use the orig Trade from `TradingGateway.ts` now?
then we do not need to use TradeRaw with its weird capitalisation ...

algreasley: not sure why we need this loose type def when we have clearly defined types for FX and Credit trades ..

src/workspace/src/services/trades.ts
```diff
@@ -1,9 +1,11 @@
-import { map } from 'rxjs/operators'
-import { BlotterService } from '../generated/TradingGateway'
+import { map } from "rxjs/operators"
+import { BlotterService } from "../generated/TradingGateway"
```
algreasley, line 2: minor/distraction: this is surely a case for an `@/generated` path, innit?

src/client/src/services/trades/__mocks__/trades.mock.ts
```diff
@@ -45,7 +45,7 @@ export const mockRawTrades = {
 
 export const mockTrades: FxTrade[] = [
   {
-    tradeId: "1111111111",
+    tradeId: "1111111111" as unknown as bigint,
```
algreasley, line 48: eww, we need to sort out those types properly .. as I say, though, we don't have to refactor everything at once

algreasley (approved):
great stuff ...
let's raise a ticket for the discovered tech debt around the types ..

PR #2195, opened 2023-05-11
fix: OF launcher flash of white to a minimum (milliseconds)

Dev-Alex23:
OpenFIn Launcher has a white screen flash that occurs for approximately 2 seconds during the initial launch of the OpenFin application when there is no cached data available. The proposed fix enhances the user experience by setting the "autoShow" option in the launcher manifest JSON to false and utilizing the useEffect hook to display the page on first load instead.

This doesnt completely get rid of the flash as it seems to be something OpenFin tends to do, for example the `waitForPageLoad` desc is : "The user will experience a brief white flash until the content arrives and is rendered. This can be mitigated somewhat by setting the backgroundColor option to some other color compatible with the expected content.". 
The problem with this solution however is that it doesn't take into account theme switch

### DIFF
--- a/src/client/public-openfin/launcher.json
+++ b/src/client/public-openfin/launcher.json
@@ -55,7 +55,7 @@
   "snapshot": {
     "windows": [
       {
-        "autoShow": true,
+        "autoShow": false,
         "frame": false,
         "_comment": "Openfin Excel API preloaded below + added in appAssets (not included in standard OpenFin package)",
         "preload": [
@@ -72,7 +72,6 @@
         "defaultHeight": 56,
         "defaultTop": 160,
         "defaultLeft": 30,
-        "backgroundColor": "#313131",
         "cornerRounding": {
           "width": 4,
           "height": 4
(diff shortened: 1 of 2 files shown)


src/client/public-openfin/launcher.json
```diff
@@ -72,7 +72,6 @@
         "defaultHeight": 56,
         "defaultTop": 160,
         "defaultLeft": 30,
-        "backgroundColor": "#313131",
```
algreasley, line 75: I thought the `backgroundColor` was supposed to help, per OF docs extract ..
(assuming it is set to dark theme ofc) .. so how come we still get 2s or so of white flash?

src/client/public-openfin/launcher.json
```diff
@@ -55,7 +55,7 @@
   "snapshot": {
     "windows": [
       {
-        "autoShow": true,
+        "autoShow": false,
```
algreasley, line 58: I am puzzled that we have not deemed it necessary to add this in the past .. it must have got measurably worse, but I must admit, I have not noticed any marked change over the time I have been using RT

algreasley:
Close this as it's not a full solution (we get the same white flash  on RT startup) .. should engage with OF about this, as it must be a common issue - also, there may be a way of determining the appropriate bg colour prior to the .show(), to prevent the flash entirely

PR #2196, opened 2023-05-15
Feat/of notification color

Dev-Alex23:
Distinguishing the different kind of notifications by adding respective color


algreasley:
nothing wrong with the commits, just the orig branch naming (capitalised Feat) was causing havoc on checkout ..

PR #2197, opened 2023-05-15
chore: fix branch-based logic in GH Action workflows

algreasley:


.github/workflows/branch-fork.yml
```diff
@@ -0,0 +1,48 @@
...
+
+    runs-on: ubuntu-20.04
+
+    if: github.event.repository.full_name != github.repository
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+
+      - name: Set Node to v16
```
SHolleworth, line 24: Minor, Node 20 now

.github/workflows/pull-fork.yml
```diff
@@ -16,15 +16,16 @@ jobs:
...
     runs-on: ubuntu-20.04
 
-    if: "github.event.pull_request.head.repo.fork"
+    if: github.event.repository.full_name != github.repository
 
     steps:
       - name: Checkout code
         uses: actions/checkout@v3
+
       - name: Set Node to v16
```
SHolleworth, line 25: Same again

SHolleworth (approved):

.github/workflows/branch-fork.yml
algreasley: ah, the comment, yes .. I knew there would be something
as Node 20 seems to be all good to go, we should bump to this locally and test it all out then update the `engine` in package.json too

SHolleworth (approved):

PR #2200, opened 2023-05-18
5151/workspace dock

SHolleworth:
Introduced the OpenFin Workspace dock.

Screenshot 2023-05-18 at 16 47 38


algreasley (approved):
looks great (still think we should switch to the transparent bg pngs, mind

PR #2215, opened 2023-06-20
Google Analytics upgrade

algreasley:
Upgraded from UA to GA4, which for RT just means getting the Google Analytics source from a different address, including the G-xxxxx "tag" in the URL along with slightly modified script logic using conditional HTML insertion in the Vite build.
Then doing the same in the GA config setup in the main app in code.

Demonstrates switch to using a "dev" property for all but Prod, to keep stuff like local testing out of the reports that can track external usage of RT + friends

_Tested locally and on the PR deploy, per std bot comment below - if you have GA access, can be tracked in GA realtime report [here](https://analytics.google.com/analytics/web/#/p383903600/realtime/overview)_

_Deferring E2E against Prod to follow on - plan at the moment is to use a query param or something .._

rtc-4822-GA-UI--fx-trade-attempt-custom-event

### DIFF
--- a/src/client/package.json
+++ b/src/client/package.json
@@ -56,7 +56,7 @@
     "e2e:web:smoke": "playwright test --project=chrome --grep @smoke"
   },
   "dependencies": {
-    "@adaptive/hydra-platform": "^3.120.0",
+    "@adaptive/hydra-platform": "3.120.0",
     "@finos/fdc3": "^1.2.0",
     "@openfin/core": "31.75.4",
     "@openfin/workspace": "13.0.7",
@@ -66,7 +66,7 @@
     "d3": "^6.6.2",
     "date-fns": "^2.19.0",
     "downshift": "^6.1.9",
-    "lodash": "^4.17.21",
+    "lodash": "4.17.21",
     "openfin-notifications": "1.23.0",
     "polished": "^4.1.1",
     "query-string": "^7.1.3",
--- a/src/client/src/App.tsx
+++ b/src/client/src/App.tsx
@@ -1,11 +1,11 @@
 import { StrictMode } from "react"
 import { createRoot } from "react-dom/client"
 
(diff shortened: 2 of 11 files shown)


src/client/package.json
```diff
@@ -56,7 +56,7 @@
     "e2e:web:smoke": "playwright test --project=chrome --grep @smoke"
   },
   "dependencies": {
-    "@adaptive/hydra-platform": "^3.120.0",
+    "@adaptive/hy‰dra-platform": "3.120.0",
```
snony14, line 59: There is a weird character after y in `hydra`, see picture below:

snony14: image

src/client/src/App.tsx
```diff
@@ -37,19 +37,9 @@ export async function initApp() {
...
-    dimension2: gaDimension,
-    dimension3: import.meta.env,
-    page: window.location.pathname,
-  })
-
-  ga("send", "pageview")
+  window.gtag("js", new Date())
+  window.gtag(
+    "config",
+    ENVIRONMENT === "prod" ? "G-Z3PC9MRCH9" : "G-Y28QSEPEC8",
```
snony14, line 43: Should these value come from `.env` file?

src/client/package.json
algreasley: y, I saw that .. should have pushed up correction last week, sorry for the distraction

src/client/src/App.tsx
algreasley: hmm .. I was about to say yes, but then I thought about running a Prod build locally .. this ofc does not save you from that case either .. I'll think again on this, we kind of need a dynamic check .. but that is going to be difficult as it is invoked in the `index.html` .. 🤔

snony14 (approved):

SHolleworth (approved):

PR #2219, opened 2023-06-21
fix: revert TradingWS to tiles, blotter and RA

algreasley:


src/workspace/config/snapshot.json
```diff
@@ -1,183 +1,110 @@
 {
   "snapshotDetails": {
-    "timestamp": "2023-06-20T12:17:19.360Z",
-    "runtimeVersion": "29.108.73.14",
+    "timestamp": "2022-01-10T11:26:57.741Z",
+    "runtimeVersion": "23.96.67.7",
```
SHolleworth, line 4: should this have been reverted all the way back to 23?

src/workspace/config/snapshot.json
```diff
@@ -247,121 +175,45 @@
...
+        }
+      },
       "saveWindowState": false,
       "shadow": false,
       "showTaskbarIcon": true,
       "showBackgroundImages": false,
       "smallWindow": false,
       "spellCheck": false,
-      "state": "normal",
+      "state": "maximized",
```
SHolleworth, line 205: do we definitely want it to maximise when it's launched?

src/workspace/config/snapshot.json
algreasley: it doesn't matter .. that was working before I updated the snapshot

src/workspace/config/snapshot.json
algreasley: again, just restoring previous

SHolleworth (approved):

PR #2220, opened 2023-06-21
improvement: updated OF WS, OF notifications, OF core and runtimes latest

SHolleworth:


snony14 (approved):

src/client/src/OpenFin/Snapshots/Snapshots.tsx
```diff
@@ -1,3 +1,4 @@
+import OpenFin from "@openfin/core"
```
algreasley, line 1: funny we did not (seem to) need this before ...
correct tho, we shouldn't be relying on floating "globals" (I know it's "only" typing)
while you are at it, can you just check `fin.d.ts` in client and workspace - I would call that a _tactical_ setting, and would prefer this kind of import-driven approach

src/workspace/src/browser.ts
```diff
@@ -1,11 +1,12 @@
 import {
-  BrowserOverrideCallback,
   getCurrentSync,
   GlobalContextMenuOptionType,
   OpenGlobalContextMenuPayload,
-  Page
-} from '@openfin/workspace-platform'
-import { getUserToSwitch, switchUser } from './user'
+  Page,
```
algreasley, line 5: minor niggle this, but let's do a one-off pass of prettying on the codebase, we have had a series of PRs against /workspace now and it's hard to see the _real_ diffs (I am as guilty of this as anyone, btw, not a criticism of this PR alone)

algreasley:
consider this pre-approved - address what comments you can above, but happy with it generally, nice work

src/client/vite.config.ts
```diff
@@ -20,6 +20,8 @@ type BuildTarget = "web" | "openfin" | "finsemble"
 
 const localPort = Number(process.env.PORT) || 1917
 
+const openfinRuntime = "31.112.75.4"
```
snony14, line 23: Minor: change to `OPENFIN_RUN_TIME` since this is a constant.

snony14 (approved):

snony14 (approved):

PR #2225, opened 2023-06-27
open chart iq with intent

snony14:


src/client/src/App/Trades/CoreFxTrades.tsx
```diff
@@ -17,6 +17,7 @@ const FxTrades = () => {
     }
     if (window.fdc3) {
       broadcast(context)
+      window.fdc3.raiseIntent("ViewChart", context)
```
algreasley, line 20: per offline discussion, I would like to see the ViewChart intent as the _only_ FDC3 thing going on here .. broadcast is fine for general info-sharing, but what we are doing here is invoking something, so an Intent is the most appropriate FDC3 construct

algreasley: in fact, this should start RA if not already running .. that's how Intents are supposed to work, and are probably working fine in Finsemble, to invoke the ChartIQ app, right?

PR #2230, opened 2023-07-06
chore: added an npm script to run client pointed to local backend

SHolleworth:


src/client/src/services/connection.ts
```diff
@@ -46,8 +46,12 @@ const connectionExists = (status: ConnectionStatus, dispose: Disposable) => {
 
 // Connect to Hydra gateway and store the disposable
 export const initConnection = async () => {
+  const url = import.meta.env.VITE_LOCAL_HYDRA_URL
```
algreasley, line 49: should not need to change the app code ... see https://github.com/AdaptiveConsulting/ReactiveTraderCloud/blob/master/src/client/vite.config.ts#L338

src/client/package.json
```diff
@@ -32,6 +32,7 @@
     "test": "vitest",
     "verify": "npm run typecheck && npm run lint && npm run format:check && npm run test",
     "start": "cross-env TARGET=web vite",
+    "start:local": "cross-env VITE_LOCAL_HYDRA_URL=ws://localhost:8929 TARGET=web vite",
```
algreasley, line 35: these should be fine with VITE_HYDRA_URL=ws://localhost:8929 and no changes in `connection.ts` .. pretty sure I used this approach with recent local testing on prev laptop .. but yeah, not sure why I/we did not add a local script 😄

algreasley (approved):

PR #2231, opened 2023-07-17
feat: workspace nlp rfq

SHolleworth:
Feature for raising an RFQ from the Workspace home input bar.

image

After raising the rfq the user can open the credit rfqs view

image

### DIFF
--- a/src/workspace/src/home/index.ts
+++ b/src/workspace/src/home/index.ts
@@ -18,6 +18,8 @@ import {
 import { execute } from "@/services/executions"
 import { getUserResult, getUserToSwitch, switchUser } from "@/user"
 import { getNlpResults } from "./nlpProvider"
+import { createCreditRfq } from "@/services/creditRfqRequests"
+import { ExitCode } from "@openfin/core/src/OpenFin"
 
 const PROVIDER_ID = "adaptive-home-provider"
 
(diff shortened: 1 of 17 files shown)


src/workspace/src/services/connection.ts
SHolleworth, line 1: I had to copy a few pieces of our client services over to workspace

src/workspace/src/services/creditInstruments.ts
SHolleworth, line 1: As well as write some of my own modified versions of services

src/workspace/src/home/index.ts
```diff
@@ -80,55 +82,90 @@ export async function registerHome(): Promise<HomeRegistration> {
...
+            direction,
+            notional,
+            dealtCurrency,
+          })
+        }
+
+        break
+      }
+
+      case "rfq-execution": {
```
SHolleworth, line 129: The new case

src/workspace/src/services/connection.ts
algreasley: I'd like to converge client/workspace if at all possible .. just something to keep in mind, not an immediate goal

TobyLi96 (approved):

src/workspace/src/services/connection.ts
```diff
@@ -0,0 +1,172 @@
+import {
```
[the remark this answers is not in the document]
SHolleworth: Good catch, I think we should actually be using initConnection so I switched that out. I've gotten rid of withIsStaleData

src/workspace/src/home/nlpProvider.ts
```diff
@@ -5,24 +5,40 @@ import {
...
   CurrencyPair,
   Direction,
+  InstrumentBody,
+  NackCreateRfqResponse,
+  RfqBody,
   Trade,
   TradeStatus,
 } from "@/generated/TradingGateway"
+import { creditInstruments$ } from "@/services/creditInstruments"
+import { createdRfqInstrumentReponse$ as createdRfqWithInstrumentWithReponse$ } from "@/services/creditRfqRequests"
```
TobyLi96, line 31: minor: typo

TobyLi96 (approved):

TobyLi96 (approved):

PR #2249, opened 2023-08-07
5486/nlp credit maturity implementation

tammyk7:


src/services/credit/creditInstruments.ts
```diff
@@ -9,7 +9,8 @@ import {
   RemovedInstrumentUpdate,
   START_OF_STATE_OF_THE_WORLD_INSTRUMENT_UPDATE,
 } from "generated/TradingGateway"
-import { map, scan } from "rxjs/operators"
+import { of } from "rxjs"
```
algreasley, line 12: looks unnecessary ..

src/client/OpenFin/apps/Launcher/NlpSuggestions/executions/RfqExecution/state.ts
```diff
@@ -75,7 +76,7 @@ const rfqExecutionState$: Observable<RfqNlpExecutionState> = nlpIntent$.pipe(
         exhaustMap(() =>
           createCreditRfq$({
             instrumentId: instrument.id,
-            quantity: notional,
+            quantity: notional * 1000,
```
algreasley, line 79: just cross-check with the non-NLP behaviour, please .. ideally, we should not be applying this x 1000 willy nilly wherever we kick off an RFQ .. if we can converge all points to one place where the conversion is done, then I'd be happier (not always possible, ofc)
UI treatments will probably be distinct e.g. NLP Are you sure? bit above vs RFQ Form where it has the (1000s) label as a reminder

src/client/OpenFin/apps/Launcher/NlpSuggestions/executions/RfqExecution/RfqExecution.tsx
```diff
@@ -9,7 +9,8 @@ import { onNext, useRfqExecutionState } from "./state"
 
 const Usage = () => (
   <HelpText>
-    Usage: <Pill>buy/sell</Pill> <Pill>quantity</Pill> <Pill>bond</Pill>
+    Usage: <Pill>buy/sell</Pill> <Pill>quantity</Pill> <Pill>bond</Pill>{" "}
+    <Pill>maturity</Pill>
```
algreasley, line 13: we discussed making this optional .. I don't see any code for that
and we should show that it is optional here .. and need to state that it is 2-digit year, don't we? or will it work with 2 or 4?

src/client/OpenFin/apps/Launcher/NlpSuggestions/executions/RfqExecution/state.ts
```diff
@@ -48,10 +48,11 @@ const rfqExecutionState$: Observable<RfqNlpExecutionState> = nlpIntent$.pipe(
 
     const requestData =
       intent.payload as RfqNlpExecutionDataReady["payload"]["requestData"]
-    const { symbol, direction, notional } = requestData
+    const { symbol, direction, notional, maturity } = requestData
 
     const instrument = instruments.find(
-      (instrument) => instrument.ticker === symbol,
+      (instrument) =>
+        instrument.ticker === symbol && instrument.maturity.includes(maturity),
```
algreasley, line 55: so if I type "20" as maturity what happens?

src/client/OpenFin/apps/Launcher/NlpSuggestions/executions/RfqExecution/RfqExecution.tsx
tammyk7: You make it optional in the dialogflow (so wont be seen in the code) and yes ill add in the helper text that it is an optional parameter (2-digit)

PR #2246, opened 2023-08-01
feat: Launch apps as platforms and inter-platform interop

TobyLi96:
- Launch FX/Credit/Limit Checker/Reactive Analytics as platforms (enables snapshot)
- Add inter-platform interop to keep interop between FX and Reactive Analytics/Limit Checker working

PR #2250, opened 2023-08-08
chore: downgrade OF runtime to 29 for FX and credit

TobyLi96:
Based off of #2246. In this way, the launcher can launch as v31, and FX/Credit launches as v29.

### DIFF
--- a/public-openfin/rt-credit.json
+++ b/public-openfin/rt-credit.json
@@ -17,7 +17,7 @@
     }
   },
   "runtime": {
-    "version": "<OPENFIN_RUNTIME>",
+    "version": "29.108.73.14",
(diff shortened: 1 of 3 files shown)


algreasley (commented):
this seems to work fine - I would just be happier if we synced runtime across "our" stuff .. 
we can leave Workspace to its own devices as that is "their" UI .. and I would be scared to deviate much from what they specify on the [Workspace version page](https://developer.openfin.co/versions/?product=Runtime#/?product=Workspace).

algreasley (approved):
looks good .. please sync RA too (keen to get these co-located but we need to sort out infra first)

PR #2253, opened 2023-08-16
chore: add default maturity to confirm message when selecting the wro…

tammyk7:
…ng year

### DIFF
--- a/packages/client/src/client/OpenFin/apps/Launcher/NlpSuggestions/executions/RfqExecution/RfqExecutionWorkflow.tsx
+++ b/packages/client/src/client/OpenFin/apps/Launcher/NlpSuggestions/executions/RfqExecution/RfqExecutionWorkflow.tsx
@@ -1,7 +1,7 @@
 import { Card as RfqCard } from "client/App/Credit/CreditRfqs/CreditRfqCards/CreditRfqCard"
 import { formatNumber } from "client/utils"
 import { ACK_CREATE_RFQ_RESPONSE, Direction } from "generated/TradingGateway"
-import { creditInstruments$ } from "services/credit"
+import { useCreditInstruments } from "services/credit"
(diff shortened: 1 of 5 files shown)


packages/client/src/client/OpenFin/apps/Launcher/NlpSuggestions/executions/RfqExecution/RfqExecutionWorkflow.tsx
```diff
@@ -22,10 +23,20 @@ const ConfirmContent = ({
...
   useMoveNextOnEnter(onNext)
   const directionStr = direction === Direction.Buy ? "buying" : "selling"
   const notionalStr = formatNumber(notional * 1000)
+  const creditInstrument = creditInstruments$
+    .getValue()
+    .filter((bond) => bond.ticker === symbol)
+
+  const selectedMaturity =
+    maturity === creditInstrument[0].maturity.slice(0, 4) ||
+    maturity === creditInstrument[1].maturity.slice(0, 4)
```
TobyLi96, line 37: Should this be written to accommodate for cases where there are only one maturity or more than 2 maturities per instrument?

TobyLi96: I know currently we always have two maturities per instrument, but we could potentially add more maturities/instruments in the future.

algreasley: yeah, this is too brittle .. we are in control of the test data, so we have 2 of each currently, but someone could helpfully amend that data at any point so we have more instruments to play with ..

PR #2255, opened 2023-08-17
chore: re-arrange client and server as packages

algreasley:
Following on from the excellent baseline by @SHolleworth which brought client and workspace together, this is a minor re-arrangement to push client and server into a packages/* structure (that will be familiar to anyone who has worked on monorepos - RT is not a monorepo atm, but may be, if we can sort out separate deployments).

There is more to come, hence draft .. main things are combining the openfin (=== container) & openfin workspace builds and then the deployments, which should simplify our Actions and speed up every build.

It is mostly 100s or file renames, so I'll flag the important bits ...


packages/client/package-lock.json
algreasley, line 1: I notice that this still has the backward compatible "dependencies" block in it (as we are using Node 16 === npm 8)
see https://docs.npmjs.com/cli/v9/configuring-npm/package-lock-json#lockfileversion

TobyLi96 (approved):
I find `packages/client/src/client` a bit weird with the two clients. Maybe `packages/rt/sr/client` or something like this would sound better? Just my personal feeling, feel free to ignore.

PR #2259, opened 2023-08-21
deployment(maintenance): update weekly e2e to run against dev

stephan-thibodeau:
As per latest communication, updated env to be tested on a weekly basis as dev is in sync with latest e2e scripts


algreasley (approved):
thanks @stephan-thibodeau

PR #2260, opened 2023-08-22
feat: launcher nlp returns instrument with default maturity when matu…

TobyLi96:
…rity is invalid

### DIFF
--- a/packages/client/src/client/OpenFin/apps/Launcher/NlpSuggestions/executions/RfqExecution/state.ts
+++ b/packages/client/src/client/OpenFin/apps/Launcher/NlpSuggestions/executions/RfqExecution/state.ts
@@ -51,15 +51,14 @@ const rfqExecutionState$: Observable<RfqNlpExecutionState> = nlpIntent$.pipe(
       intent.payload as RfqNlpExecutionDataReady["payload"]["requestData"]
     const { symbol, direction, notional, maturity } = requestData
 
-    const instrument = instruments.find((instrument) => {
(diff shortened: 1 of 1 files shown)


packages/client/src/client/OpenFin/apps/Launcher/NlpSuggestions/executions/RfqExecution/state.ts
```diff
@@ -51,15 +51,14 @@ const rfqExecutionState$: Observable<RfqNlpExecutionState> = nlpIntent$.pipe(
...
-      return instrument.ticker === symbol && instrumentDate === maturity
-    })
+    const symbolInstruments = instruments.filter(
+      (instrument) => instrument.ticker === symbol,
+    )
+    const instrument =
+      symbolInstruments.find(
+        (instrument) =>
+          !maturity || maturity === instrument.maturity.slice(0, 4),
+      ) ?? symbolInstruments[0]
```
algreasley, line 61: logic's lookin' good .. 
but `symbolInstruments` can be `[]`?
and if poss, handling lack of maturity (lol) outside the find would be better as we don't actually need to iterate over anything in that case - minor, obvs, but might work out nicely, if dealing with empty instruments anyway?

TobyLi96: I've updated the logic to avoid `find` if maturity isn't specified

algreasley (approved):

PR #2261, opened 2023-08-24
vite config logic and logging

algreasley:


packages/client/vite.config.ts
```diff
@@ -70,11 +70,11 @@ function targetBuildPlugin(dev: boolean, target: string): Plugin {
         const mockPath = `${file.dir}/${file.name}.${target}.${extension}`
         return this.resolve(mockPath, importer)
       } else {
-        const rootPrefix = "src/client/"
+        const sourceRootPrefix = "/src"
```
algreasley, line 73: with the new folder structure, this was cutting out import paths like executions.ts

packages/client/vite.config.ts
```diff
@@ -70,11 +70,11 @@ function targetBuildPlugin(dev: boolean, target: string): Plugin {
...
         const thisImporter = (importer || "").replace(/\\/g, "/")
         if (
           !importer ||
-          !thisImporter.includes(rootPrefix) ||
+          // !thisImporter.includes(rootPrefix) ||
```
algreasley, line 77: need to check, but don't think there can be any non-src importers ?

packages/client/vite.config.ts
```diff
@@ -87,11 +87,10 @@ function targetBuildPlugin(dev: boolean, target: string): Plugin {
         const candidatePath = path.join(
           // If imported file starts with /src we can not append it to importer dir
           // so we need to strip the path by the rootPrefix first
-          importedFile.dir.startsWith("/src") &&
-            importerFile.dir.includes(rootPrefix)
-            ? `${importerFile.dir.split(rootPrefix)[0]}/`
+          importedFile.dir.startsWith(sourceRootPrefix) &&
```
algreasley, line 90: this can go away, ofc, but it's instructive for now, to see how the candidate is derived

packages/client/vite.config.ts
algreasley: wrong .. this tracks everything, node_modules files as importer _and_ source

packages/client/vite.config.ts
```diff
@@ -70,51 +70,48 @@ function targetBuildPlugin(dev: boolean, target: string): Plugin {
         const mockPath = `${file.dir}/${file.name}.${target}.${extension}`
         return this.resolve(mockPath, importer)
       } else {
-        const rootPrefix = "src/client/"
-        const thisImporter = (importer || "").replace(/\\/g, "/")
         if (
           !importer ||
-          !thisImporter.includes(rootPrefix) ||
+          importer.includes("node_modules") ||
```
algreasley, line 75: cuts out resolutions relating to external libs like `d3` etc ..

packages/client/vite.config.ts
```diff
@@ -70,51 +70,48 @@ function targetBuildPlugin(dev: boolean, target: string): Plugin {
...
           source === "./main"
         ) {
           return null
         }
 
-        const importedFile = path.parse(source)
+        if (!source.startsWith(".") && !source.startsWith("/")) {
```
algreasley, line 81: our imports should be relative (./ or ../) or aliased (/src/...) .. ignoring `~/` as that is an anomaly used once in storybook code that should not really be there, but SB stuff going away with introduction of UISK I hope

packages/client/vite.config.ts
```diff
@@ -70,51 +70,48 @@ function targetBuildPlugin(dev: boolean, target: string): Plugin {
...
+          return null
+        }
+
+        const sourcePath = path.parse(source)
```
algreasley, line 85: by the time we get here, we are down to 865 genuine cases of imports that could be switched to target-specific replacements .. we may be able to filter further, but this seems reasonable now

packages/client/vite.config.ts
```diff
@@ -70,51 +70,48 @@ function targetBuildPlugin(dev: boolean, target: string): Plugin {
...
-            console.log("candidate good", candidate)
-          } catch (e) {
-            // console.log("Error with candidate", candidate, e)
-            candidate = null
-          }
-
-          if (candidate) {
-            break
-          }
+        const candidateTs = `${candidatePath}.ts`
```
algreasley, line 105: for ref, down to 18 candidate imports, with 7 distinct replacements (for openfin target)
if we ever amend this logic, we need to ensure that we don't cut any out

packages/client/vite.config.ts
```diff
@@ -70,51 +70,48 @@ function targetBuildPlugin(dev: boolean, target: string): Plugin {
...
+        if (existsSync(candidateTs)) {
```
algreasley, line 106: `existsSync` is a nicer API call to use, no nasty exceptions and we don't need all the stats, just "is it there?"

TobyLi96 (approved):

PR #2263, opened 2023-08-29
fix: removed some styling which was cutting off the bottom of the blotter

SHolleworth:


packages/client/src/client/App/Trades/TradesGrid/styled.ts
```diff
@@ -9,7 +9,6 @@ import { LimitCheckStatus, TradeStatus } from "services/trades/types"
 import styled, { css } from "styled-components"
 
 export const Table = styled.div`
-  border-bottom: 15px solid ${({ theme }) => theme.core.darkBackground};
```
algreasley, line 12: this came from LC table mods .. is the LC instance still ok?

algreasley: oh, looks like this was just moved around, not introduced for LC .. still worth checking all is well, tho

SHolleworth: LC still looks okay, but a good idea to check

algreasley (approved):

PR #2266, opened 2023-08-30
feat: enable LC app in workspace

algreasley:


packages/client/src/workspace/apps.ts
```diff
@@ -1,20 +1,32 @@
...
+export function getViews(): App[] {
+  return getCurrentUser() === USER_TRADER
+    ? [
+        reactiveTraderFxLiveRatesView,
+        reactiveTraderFxTradesView,
+        reactiveTraderFxAnalyticsView,
+        reactiveAnalyticsView,
+      ]
+    : [
+        reactiveTraderFxTradesView,
```
SHolleworth, line 21: is there a reason that live rates had been excluded from here?

packages/client/src/workspace/apps.ts
```diff
@@ -1,20 +1,32 @@
...
-    limitCheckerView,
-    reactiveTraderFxLiveRatesView,
-    reactiveTraderFxTradesView,
-    reactiveTraderFxAnalyticsView,
-    reactiveAnalyticsView,
-  ]
+export function getApps(): App[] {
+  return getCurrentUser() === USER_TRADER
+    ? [reactiveTraderFx, reactiveTraderCredit, reactiveAnalytics]
+    : [reactiveAnalytics, limitChecker]
```
TobyLi96, line 9: This might just be me not understanding limit checker properly, but wouldn't a user want access to both limit checker and FX?

algreasley: no, this is where I fell down earlier creating a workspace with Live Trades and LC .. the whole point of the Limit Checker is that it is being used to control a Trader, so Trader should not have access (makes the interop difficult, I know!) .. anyway, we can demo by opening LC as Operator and trading as Trader .. smoke and mirrors but it's always going to be difficult demoing as a single user (ref also Credit demo, with pretend "sell side" popup ..)

packages/client/src/workspace/apps.ts
algreasley: yes, see previous categorisation in the store - the Home search was not categorised previously, but should have been if Store was restricted (obviously) .. we will need to verify the current state of the Workspace config with stakeholders/demoers but that's what we'll use UAT deployment for ..

SHolleworth (approved):

PR #2267, opened 2023-08-30
fix: vertical scroll bar no longer cutoff

SHolleworth:


algreasley (approved):
tested locally .. looks good to me

PR #2269, opened 2023-08-30
fix: added limit column, notional column showing trade notional

SHolleworth:
image

### DIFF
--- a/packages/client/e2e/limit-checker.spec.ts
+++ b/packages/client/e2e/limit-checker.spec.ts
@@ -38,7 +38,9 @@ test.describe("Limit Checker", () => {
       .locator("input")
 
     const lastTradeIdString = await tradeBlotterFirstRowCells.nth(1).innerText()
-    const lastTradeId = isNaN(Number(lastTradeIdString)) ? 0 : Number(lastTradeIdString)
+    const lastTradeId = isNaN(Number(lastTradeIdString))
+      ? 0
+      : Number(lastTradeIdString)
 
     const limitIdString = await limitTableFirstRowCells.nth(1).innerText()
     const limitId = isNaN(Number(limitIdString)) ? -1 : Number(limitIdString)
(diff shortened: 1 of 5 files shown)


packages/client/src/client/OpenFin/apps/LimitChecker/state.ts
```diff
@@ -40,7 +40,8 @@ const limitResult$ = limitCheckRequest$.pipe(
     limitBySymbol$(request.tradedCurrencyPair).pipe(
       take(1),
       map(([value, formattedValue]) => ({
-        notional: formattedValue,
+        notional: formatNotional(request.notional, ["k", "m"])[1],
```
algreasley, line 43: don't like this array dereferencing .. arbitrary "second value" .. but not a now thing

packages/client/e2e/limit-checker.spec.ts
```diff
@@ -50,26 +52,35 @@ test.describe("Limit Checker", () => {
...
+      ],
+      firstCellToAssert: 1,
+      lastCellToAssert: 6,
+    })
+
+    await assertGridRow({
+      row: tradeBlotterFirstRowCells,
+      assertions: ["Buy", "EURUSD", "EUR", "1,999,999"],
+      firstCellToAssert: 4,
+      lastCellToAssert: 8,
```
algreasley, line 77: little bit confusing that this means the previous cell is the last one to actually assert on .. right?

packages/client/e2e/limit-checker.spec.ts
```diff
@@ -38,7 +38,9 @@ test.describe("Limit Checker", () => {
       .locator("input")
 
     const lastTradeIdString = await tradeBlotterFirstRowCells.nth(1).innerText()
-    const lastTradeId = isNaN(Number(lastTradeIdString)) ? 0 : Number(lastTradeIdString)
+    const lastTradeId = isNaN(Number(lastTradeIdString))
+      ? 0
+      : Number(lastTradeIdString)
 
     const limitIdString = await limitTableFirstRowCells.nth(1).innerText()
     const limitId = isNaN(Number(limitIdString)) ? -1 : Number(limitIdString)
```
algreasley, line 46: aside: why aren't we locating based on the exact ticket name (ccy pair) rather than nth-ing it out of all "EUR" tiles ?

packages/client/e2e/limit-checker.spec.ts
```diff
@@ -50,26 +52,35 @@ test.describe("Limit Checker", () => {
...
-
-    await assertGridRow(
-      tradeBlotterFirstRowCells,
-      ["Buy", "EURUSD", "EUR", "1,999,999"],
-      4,
-      8,
-    )
+    await tradeBlotterFirstRowCells
+      .nth(1)
+      .getByText(lastTradeId + 1 + "")
```
algreasley, line 57: oof, let's come back to this ...

algreasley (approved):

PR #2274, opened 2023-08-31
fix: point ws icons to correct folder

SHolleworth:
Workspace icon urls were pointing to the wrong place in both envs and local. To fix I added some filtering to the contructUrl function to remove workspace/ dir when in an env, and change port from 2017 to 1917 when in local


packages/client/src/client/OpenFin/apps/LimitChecker/LimitCheckResults.tsx
```diff
@@ -8,7 +8,6 @@ import { LimitCheckStatus, LimitCheckTrade } from "services/trades/types"
 import styled from "styled-components"
 
 import { limitResult$ } from "./state"
-
```
algreasley, line 11: triv: restore whitespace

packages/client/src/client/utils/constructUrl.ts
```diff
@@ -0,0 +1,19 @@
...
+  path: string,
+  base = window.location.origin,
+): string => {
+  //removing workspace dir from workspace base path
+  const basePathSections = BASE_PATH.split("/")
+  const lastSection = basePathSections.pop()
+  const basePath =
+    lastSection === "workspace" ? basePathSections.join("/") : BASE_PATH
+
+  //replace workspace port 2017 with openfin port 1917
```
algreasley, line 13: as disc, hopefully simplify (remove) this bit soon

algreasley (dismissed):

PR #2277, opened 2023-09-08
Failing to activate on new release

snony14:


packages/client/src/client/Web/sw.js
```diff
@@ -83,3 +84,27 @@ self.addEventListener("message", (event) => {
     self.skipWaiting()
   }
 })
+
+// from https://web.dev/service-worker-lifecycle/#skip-the-waiting-phase
+self.addEventListener("install", (event) => {
+  console.debug("SW: install", event)
+  // self.skipWaiting()
```
snony14, line 91: I suppose the reason we are not calling `skipWaiting` here is that we want to show the reload dialog?

packages/client/src/client/Web/serviceWorkerRegistration.ts
```diff
@@ -53,51 +56,78 @@ export function register(config?: Config) {
...
         // Is not localhost. Just register service worker
-        registerValidSW(swUrl, config)
+        registerServiceWorker(swUrl, config)
       }
     })
   }
 }
 
-function registerValidSW(swUrl: string, config?: Config) {
+function addStateChangeHandler(
```
algreasley, line 65: pulled this out as a common handler ... we can use in the usual case when installing worker -> installed, but also when we come across a waiting SW already in "installed" -- this has happened several times in testing (seems to be exacerbated by using slow connection - Slow 3G in Network throttling)

packages/client/src/client/Web/serviceWorkerRegistration.ts
```diff
@@ -53,51 +56,78 @@ export function register(config?: Config) {
...
+    if (serviceWorkerInstance.state === "installed") {
+      if (navigator.serviceWorker.controller) {
+        // At this point, the updated precached content has been fetched,
+        // but the previous service worker will still serve the older
+        // content until all client tabs are closed.
+        console.debug(
+          "New content is available and will be used when all " +
+            "tabs for this page are closed. See https://cra.link/PWA.",
+        )
+        serviceWorkerInstance.postMessage({ type: "SKIP_WAITING" })
```
algreasley, line 84: just do the skip at the point of discovering we have an "installed" and waiting SW

packages/client/src/client/Web/serviceWorkerRegistration.ts
```diff
@@ -53,51 +56,78 @@ export function register(config?: Config) {
...
-      registration.onupdatefound = () => {
+      console.debug("SW Reg: handling ..", registration)
+
+      if (registration.waiting) {
+        const waitingWorker = registration.waiting
+        console.debug(
+          "'waiting' SW found .. skip waiting in case",
+          waitingWorker,
+        )
+        waitingWorker.postMessage({ type: "SKIP_WAITING" })
```
algreasley, line 116: again, do the skip at the point of discovering we have an "installed" and waiting SW

packages/client/src/client/Web/serviceWorkerRegistration.ts
```diff
@@ -53,51 +56,78 @@ export function register(config?: Config) {
...
+}
+
+function registerServiceWorker(swUrl: string, config?: Config) {
   navigator.serviceWorker
     .register(swUrl)
     .then((registration) => {
-      registration.onupdatefound = () => {
+      console.debug("SW Reg: handling ..", registration)
+
+      if (registration.waiting) {
```
algreasley, line 110: if we refreshed early or there is a pending SW from an earlier failed activation, then we will have a SW that is stuck in waiting state (installed but unable to proceed until all tabs are closed .. a refresh does not count, per SW guides)

packages/client/src/client/Web/serviceWorkerRegistration.ts
```diff
@@ -53,51 +56,78 @@ export function register(config?: Config) {
...
+        const waitingWorker = registration.waiting
+        console.debug(
+          "'waiting' SW found .. skip waiting in case",
+          waitingWorker,
+        )
+        waitingWorker.postMessage({ type: "SKIP_WAITING" })
+        addStateChangeHandler(waitingWorker, registration, config)
+      }
+
+      registration.addEventListener("updatefound", (event) => {
```
algreasley, line 120: on register, the usual case will be that a new SW is found remotely, and this is pulled down with its friends, the new hash-coded JS modules and when that is all done this "installing" SW will transition to "installed" .. this doesn't mean it is actively running on (and controlling) the page - that's what "activated" is for

packages/client/src/client/Web/serviceWorkerRegistration.ts
```diff
@@ -53,51 +56,78 @@ export function register(config?: Config) {
...
+        console.debug("Content is cached for offline use.")
+
+        // Execute callback
+        if (registrationConfig && registrationConfig.onSuccess) {
+          registrationConfig.onSuccess(registration)
+        }
+      }
+    } else if (serviceWorkerInstance.state === "activated") {
+      if (registrationConfig && registrationConfig.onUpdate) {
+        registrationConfig.onUpdate(registration)
```
algreasley, line 98: the real "done" state is activated .. at this point the new SW is controlling the current page - or it is capable of taking control of the latest modules in the event of a refresh .. any other state, and the previous activated SW will be in charge .. 
There has been a lot more logging around here to verify all this .. what is left now is the bare minimum for us to see if there are any other edge/corner cases around this problematic lifecycle !

packages/client/src/client/Web/serviceWorkerRegistration.ts
```diff
@@ -53,51 +56,78 @@ export function register(config?: Config) {
...
+        waitingWorker.postMessage({ type: "SKIP_WAITING" })
+        addStateChangeHandler(waitingWorker, registration, config)
+      }
+
+      registration.addEventListener("updatefound", (event) => {
+        console.debug("SW Reg: updatefound fired", event)
+
         const installingWorker = registration.installing
         if (installingWorker == null) {
+          console.debug("No 'installing' SW found")
```
algreasley, line 125: this has never happened (yet)

packages/client/src/client/Web/serviceWorkerRegistration.ts
```diff
@@ -53,51 +56,78 @@ export function register(config?: Config) {
...
+        )
+        serviceWorkerInstance.postMessage({ type: "SKIP_WAITING" })
+      } else {
+        // At this point, everything has been precached.
+        // It's the perfect time to display a
+        // "Content is cached for offline use." message.
+        console.debug("Content is cached for offline use.")
+
+        // Execute callback
+        if (registrationConfig && registrationConfig.onSuccess) {
```
snony14, line 92: is the `onSuccess`  redundant? I don't see where the callback is defined.

packages/client/src/client/Web/serviceWorkerRegistration.ts
snony14: This might be the second time we are calling `SKIP_WAITING` , please see line 110 to 118 of this file, though I know this is redundant in terms of SW behavior if we call `SKIP_WAITING`.

packages/client/src/client/Web/serviceWorkerRegistration.ts
algreasley: yes, I think it is .. some of this I think is inherited from CRA .. I am not worried about it, I kind of like the fact that is obviously CRA so we can use that as a reference, until we completely rewrite it as a custom thing .. which may never happen

packages/client/src/client/Web/serviceWorkerRegistration.ts
snony14, line 84: This might be the second time we are calling SKIP_WAITING , please see line 110 to 118 of this file, though I know this is redundant in terms of SW behavior if we call SKIP_WAITING.

algreasley (approved):
Approved as Moise initiated .. Moise's review is approval from his side of my updates

packages/client/src/client/Web/serviceWorkerRegistration.ts
algreasley: as discussed, this is not dup .. comments now added to clarify both cases

PR #2286, opened 2023-09-20
Bugfix: Create RFQ confirmation pill in openfin and Web

snony14:
image


packages/client/src/client/App/Credit/CreditRfqForm/CreditRfqFormCore.tsx
```diff
@@ -65,7 +67,8 @@ const CreditRfqFormCore = ({ children }: WithChildren) => {
   }, [])
   return (
     <CreditRfqFormCoreWrapper>
-      <Subscribe fallback={children}>
+      <Subscribe fallback={<Loader ariaLabel="Loading New RFQ Form" />}>
```
algreasley, line 70: why has this changed?
I had made this consistent with other *`Core` impls by using one version of Loader in parent and in the Core inner component ..

packages/client/src/client/App/Credit/CreditRfqs/CreditRfqConfirmation.tsx
```diff
@@ -97,13 +110,15 @@ export const CreditRfqCreatedConfirmation = () => {
...
-    <ConfirmationPill direction={direction}>
-      You have sent an {direction} RFQ for {formatter(quantity)}{" "}
-      {instrument?.name} to {dealerIds.length} dealers
-      <IconWrapper direction={direction} onClick={onDismissMessage}>
-        <FaTimes />
-      </IconWrapper>
-    </ConfirmationPill>
+    <ConfirmationPillWrapper>
+      <CreatedRFQConfirmationPill direction={direction}>
+        You have sent an {direction} RFQ for {formatter(quantity)}{" "}
```
algreasley, line 115: let's correct this while we are here .. right now, it gives us
```You have sent an Buy/Sell RFQ ...```

File: packages/client/src/client/App/Credit/CreditRfqs/CreditRfqConfirmation.tsx [hunk 3]
```diff
@@ -39,6 +37,16 @@ const ConfirmationPill = styled.div<{ direction: Direction }>`
   }
 `
 
+const CreatedRFQConfirmationPill = styled(ConfirmationPill)`
+  margin-right: 15px;
```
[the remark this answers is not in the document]
algreasley: this is an old comment thread .. will leave it now, as it was part of the discussion, but we have moved on from these now

packages/client/src/client/App/Credit/CreditRfqs/CreditRfqConfirmation.tsx
algreasley: see other util somewhere as I pestered Tammy to centralise the handling of the RFQ quantity (always in 000s, so need to format appropriately everywhere we present quantity to user (after they have entered the quantity in the smaller form, ofc)

packages/client/src/client/App/Credit/CreditRfqForm/CreditRfqFormCore.tsx
snony14: I haven't changed this, maybe due to merging with master, I accepted the wrong changes will revert.

packages/client/src/client/App.tsx
```diff
@@ -9,6 +9,7 @@ import { getMainApp } from "./main"
 import {
   registerCreditAcceptedNotifications,
   registerFxNotifications,
+  registerRFQRequestConfirmationNotifications,
```
algreasley, line 12: can we call it `registerCredit...` to keep with the asset class naming .. e.g. `registerCreditCreatedNotifications`
(or if you want to keep the RFQ in there - not a bad idea - add to both ..Created.. and ..Accepted..)

algreasley: We could just have `registerFXNotifications` and `registerCreditNotifications` methods, unless we need to call them individually somewhere?

packages/client/src/client/notificationsUtils.ts
```diff
@@ -40,3 +44,15 @@ export const processCreditQuote = (quote: PricedQuoteDetails) => {
...
 }
+
+export const processRFQRequestConfirmation = ({
+  request,
+  rfqId,
+}: ConfirmCreatedCreditRfq) => {
+  return {
+    title: `RFQ Created: RFQ ID ${rfqId}`,
+    rfqDetails: `You have sent a ${request.direction} RFQ for ${formatNumber(
+      request.quantity * 1000,
```
algreasley, line 55: I asked Tammy to provide some centralised utility to manage RFQ quantity but I don't think that was done
.. as it is easy to miss a case where we are not adjusting by a factor of 1000 - please add a utility, or at least a constant everywhere we are adjusting quantity, which is entered by user as integer (I think) multiple of 000s

packages/client/src/client/App/Credit/CreditRfqs/CreditRfqConfirmation.tsx
algreasley: see below, there is no other util .. must not have been created 🤦‍♀️

algreasley: this bit goes away, anyhow, right?

packages/client/src/services/credit/creditRfqRequests.ts
```diff
@@ -43,6 +53,24 @@ export const createCreditRfq$ = (request: CreateRfqRequest) => {
   )
 }
 
+export const rfqRequestConfirmation$ = createdCreditRfq$.pipe(
```
algreasley, line 56: I was going to say, we must have had this set up before for the banner/pill notification .. that's [here](https://github.com/AdaptiveConsulting/ReactiveTraderCloud/pull/2286/files#diff-8f97e13b4d29ee18eff60a545dc7079aadd9a0bd3f6add44fafa529adf81f8c7R75)
.. I guess that whole file is going away now

algreasley: makes much more sense in the services area, btw 👍🏻

packages/client/src/client/App.tsx
```diff
@@ -23,6 +24,7 @@ export async function initApp() {
 
   registerFxNotifications()
```
algreasley, line 25: please move `registerFxNotifications` to `LiveRatesCore.tsx`
and squash Credit methods to one, and replace the quote reg in `CreditRfqsCore`

packages/client/src/client/notifications.openfin.ts
```diff
@@ -215,6 +218,48 @@ const sendCreditQuoteNotification = (quote: PricedQuoteDetails) => {
...
+    },
+
+    icon: creditIconUrl,
+    category: "RFQ Created",
+    title,
+    indicator: {
+      text: "new rfq",
+      color: IndicatorColor.GRAY,
+    },
+    buttons: [
```
algreasley, line 246: you have added the View RFQ button here .. but does it work? 🤔
there doesn't seem to be the associated logic in `registerCreditCreatedNotifications` as there is in `registerCreditQuoteNotifications`
this is another potential advantage to aggregating the Credit registrations ...

packages/client/src/client/notifications.web.ts
```diff
@@ -139,3 +154,15 @@ export async function registerCreditAcceptedNotifications() {
     console.error(e)
   }
 }
+
+export async function registerCreditCreatedNotifications() {
```
algreasley, line 158: I think they should both (FX and Credit) have unregister functions .. with lazy loading and app switching, it will probably show all the wrong notifications

packages/client/src/client/notificationsUtils.ts
```diff
@@ -40,3 +44,15 @@ export const processCreditQuote = (quote: PricedQuoteDetails) => {
     )} @ $${formatNumber(quote.state.payload)}`,
   }
 }
+
+export const processRFQRequestConfirmation = ({
```
algreasley, line 48: let's call this `processCreditRfqCreated` ... and please amend my earlier `processCreditAccepted` to `processCreditRfqAccepted`

packages/client/src/client/notifications.openfin.ts
snony14: Yes, it works, it works just like the other notification when we accept or reject, it highlights the newly created RFQ.

algreasley: ah, it was a side effect of running the other registration .. cannot rely on that in case we modify the other registration

packages/client/src/client/App.tsx
```diff
@@ -21,8 +18,7 @@ export async function initApp() {
 
   checkTradingGatewayCompatibility()
 
-  registerFxNotifications()
-  registerCreditAcceptedNotifications()
+  registerCreditNotifications()
```
algreasley, line 21: this is where your dup notifications are coming from, btw .. this is run in all the windows

packages/client/src/client/OpenFin/apps/Launcher/index.tsx
```diff
@@ -98,10 +98,10 @@ function Launcher() {
   useEffect(() => {
     setOverlay(overlayRef.current)
     getCurrentWindowBounds().then(setInitialBounds).catch(console.error)
-    registerCreditQuoteNotifications()
+    registerCreditNotifications()
```
algreasley, line 101: still need this (and Workspace one) for NLP stuff, so hard to de-dup this final additional registration on top of the `CreditRfqsCore` one

packages/client/src/client/notifications.web.ts
```diff
@@ -99,11 +114,12 @@ export async function registerFxNotifications() {
 }
 
 let quotesReceivedSubscription: Subscription | null = null
```
algreasley, line 116: we are missing a `unregisterFxNotifications` here
.. we should use a type somehow to keep our variant modules honest ... 🤔

packages/client/src/services/credit/creditRfqRequests.ts
```diff
@@ -1,27 +1,37 @@
 import { createSignal } from "@react-rxjs/utils"
 import { filter, map, tap, withLatestFrom } from "rxjs/operators"
 
-import { showRfqInSellSide } from "@/client/utils"
+import { multiplyBy1000, showRfqInSellSide } from "@/client/utils"
```
algreasley, line 4: hehe .. bit literal, that name 😉
I was thinking of maybe something like `adjustUserCreditQuantity`, which is not a great name 😄 but along the lines of that tells the code reader what we are up to

packages/client/src/client/notifications.web.ts
```diff
@@ -99,11 +114,12 @@ export async function registerFxNotifications() {
...
+let acceptedRfqWithQuoteSubscription: Subscription | null = null
+let createdCreditSubscription: Subscription | null = null
 
-export async function registerCreditQuoteNotifications() {
+export async function registerCreditNotifications() {
```
algreasley, line 120: would have to be here where we check `fin.Application.getApplications` or whatever it is and do the logic to see if Launcher or Workspace is running when RT Credit starts up ... prob need to do for FX as well, with moving to `LiveRatesCore`

packages/client/src/client/notifications.finsemble.ts
```diff
@@ -43,3 +43,7 @@ export function registerCreditQuoteNotifications() {
 export function unregisterCreditQuoteNotifications() {
   // no-op
 }
+
```
algreasley, line 46: this file needs updating now .. _really_ need that type for the "variant modules"

packages/client/src/client/OpenFin/apps/Launcher/index.tsx
algreasley: this should really have the corresponding FX reg/unreg too, to be consistent

packages/client/src/client/notifications.openfin.ts
```diff
@@ -335,15 +388,49 @@ export const registerCreditAcceptedNotifications = (
...
-    })
+    acceptedRfqWithQuoteSubscription = acceptedRfqWithQuote$.subscribe(
+      (rfq) => {
+        sendQuoteAcceptedNotification(rfq)
+      },
+    )
+  }
+}
+
+let areCreatedCreditNotificationsRegistered = false
```
algreasley, line 399: minor: should not need individual guard flags for the three Credit registrations

packages/client/src/client/notifications.web.ts
```diff
@@ -44,6 +46,19 @@ const sendQuoteAcceptedNotification = ({ rfq, quote }: RfqWithPricedQuote) => {
   new Notification(title, options)
 }
 
+const sendRFQCreatedConfirmationNotification = (
```
algreasley, line 49: minor: rename as others e.g. `sendCredit..`

packages/client/src/client/notifications.openfin.ts
```diff
@@ -335,15 +388,49 @@ export const registerCreditAcceptedNotifications = (
...
+    createdCreditSubscription = createdCreditConfirmation$.subscribe(
+      (rfqRquest) => {
+        sendRFQCreatedConfirmationNotification(rfqRquest)
+      },
+    )
   }
 }
 
-export const unregisterCreditQuoteNotifications = () => {
+export const registerCreditNotifications = (
```
algreasley, line 413: _(previously posted in the wrong place 🤦‍♂️)_
Here is where we will need to check `fin.Application.getCurrentSync().getXxx`
or whatever it is and do the logic to see if Launcher or Workspace is running when RT Credit starts up
..
Will need to do for FX as well, with moving to `LiveRatesCore`

packages/client/src/workspace/provider.ts
```diff
@@ -1,8 +1,7 @@
 import { init as workspacePlatformInit } from "@openfin/workspace-platform"
 
 import {
-  registerCreditAcceptedNotifications,
-  registerCreditQuoteNotifications,
+  registerCreditNotifications,
```
algreasley, line 4: note (x-ref with `Launcher/index.tsx`) .. both registrations .. so I think we need to our level Launcher up
(like the dup notifications issue, tho, we could do this as a next step ..)

packages/client/src/client/OpenFin/apps/Launcher/index.tsx
snony14: Adding them increases the number of FX notification by a lot.

algreasley: just tested and do not get _any_ FX dups .. probs because it is triggering the notification send based off the FX Trade ACK, like Credit create/accept! ..
Yet again, it is the `quotes` that are dragging us back ...!

stephan-thibodeau:
@algreasley the recurring openfin e2e test failures against this PR are mitigated by https://github.com/AdaptiveConsulting/ReactiveTraderCloud/pull/2288 which is on PR review stage at the moment

packages/client/src/client/notifications.openfin.ts
algreasley: not so .. all notifications apart from Credit Quote derive from the ACKs to Hydra API calls, so will only occur once in the context that the call was initiated .. so as it stands we will just have the dup quote notifications as before

packages/client/src/client/notifications.openfin.ts
```diff
@@ -299,51 +396,40 @@ export const registerCreditQuoteNotifications = (
...
-    addEventListener(
-      "notification-action",
-      handler || handleHighlightCreditBlotterAction,
-    )
-
-    acceptedRfqWithQuote$.subscribe((rfq) => {
-      sendQuoteAcceptedNotification(rfq)
-    })
+  if (!areCreditNotificationsCreated) {
+    registerCreditAcceptedNotifications(handler)
```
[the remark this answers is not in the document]
algreasley: they are now all separate - makes no sense to have an "all-in-one" registration call any more

algreasley (approved):
Finally, a litany of glitches resolved, time this little fella went in

PR #2287, opened 2023-09-21
Bugfix: Blotter top row flash on refresh.

snony14:
blotter_top_row_flash

### DIFF
--- a/packages/client/src/client/App/Trades/TradesState/tableTrades.ts
+++ b/packages/client/src/client/App/Trades/TradesState/tableTrades.ts
@@ -361,21 +361,16 @@ const newTradeId$ = merge(trades$, creditTrades$).pipe(
   scan(
     (acc, trades) => {
       return {
-        stateOfWorld: acc.stateOfWorld && acc.trades.length === 0,
         trades: trades,
-        skip: acc.trades.length === trades.length,
+        count: acc.count + 1,
       }
     },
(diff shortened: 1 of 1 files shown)


packages/client/src/client/App/Trades/TradesState/tableTrades.ts
```diff
@@ -361,21 +361,16 @@ const newTradeId$ = merge(trades$, creditTrades$).pipe(
   scan(
```
algreasley, line 361: I think the old code should work - makes more logical sense
From running the old code briefly in local production build, it appears to be failing because we are combining the FX and Credit trades data, so the logic is being run twice, subverting up the booleans ...

algreasley: My suggestion would be to keep trade highlighting completely separate for the asset classes (like the hooks below)
As the code is going to be identical, I would use a function to create the final streams (or hooks)

algreasley (changes requested):

packages/client/src/client/App/Trades/TradesState/tableTrades.ts
snony14: Cool

algreasley:
will be superseded by more wide-ranging blotter highlighting fixes
.. found several more issues, and also dragged in ticket 5521 as well

PR #2288, opened 2023-09-21
Test: 5581 Improve "New RFQ/Create RFQ for GOOGL @smoke" test resiliency

stephan-thibodeau:
Latest investigated  PR and scheduled e2e test failure show high failure occurrences in regards to `New RFQ/Create RFQ for GOOGL @smoke` test case. 

**Facts**
1. Failures are happening more frequently on CI, while it was rarely happening when test being ran on local
2. Playwright report output is frequently about test timing out after default 30 seconds. No specific step causing the failure are reported
3. Screenshots are showing some quotes being available, but the test didn't select one for execution
4. At the moment of interacting with the UI, it may have other quotes rendering as selectable, potentially causing erratic behaviour with hoving and clicking on accept button

**Troubleshooting**
1. I've set the test to run in loop 10 times to find a frequency and to have few occurrences to better troubleshoot
2. I've temporary set a test workflow file to run the test in loop on CI for both web and openfin

**Observations**
1. Test flakiness is observable on CI with an average ratio of 1/5 (~20%)
2. Test flakiness ratio is lower on local (~5%)
4. test execution time is higher on ci than when ran locally. On CI it get > than 30 seconds on many failures
5. Locally test execution time average is around 25 seconds

**Action taken**
1. Since the total test execution time is close to 30 sec, then I increased test timeout to 60 seconds
2. To avoid lengthy failure in case it fail to click on the accept button, I added a 5 seconds timeout on quote click action. It should be there as the previous step wait for the button to be visible
3. Added retry logic in case the test failed to click on the  `accept` button.

### DIFF
--- a/packages/client/e2e/credit.spec.ts
+++ b/packages/client/e2e/credit.spec.ts
@@ -1,15 +1,15 @@
 import { expect, Page } from "@playwright/test"
 
 import { test } from "./fixtures"
-import { ElementTimeout, OPENFIN_PROJECT_NAME, TestTimeout } from "./utils"
+import { ElementTimeout, isOpenFin, TestTimeout } from "./utils"
 
 test.describe("Credit", () => {
   let newRfqPage: Page
   let rfqsPage: Page
   let rfqBlotterPage: Page
 
-  test.beforeAll(async ({ context, creditPagesRec }, testInfo) => {
-    if (testInfo.project.name === OPENFIN_PROJECT_NAME) {
+  test.beforeAll(async ({ context, creditPagesRec }, workerInfo) => {
+    if (isOpenFin(workerInfo)) {
       const mainWindow = creditPagesRec["mainWindow"]
 
       await mainWindow.evaluate(async () => {
@@ -19,6 +19,7 @@ test.describe("Credit", () => {
       newRfqPage = creditPagesRec["credit-new-rfq"]
       rfqsPage = creditPagesRec["credit-rfqs"]
       rfqBlotterPage = creditPagesRec["credit-blotter"]
+
       newRfqPage.setViewportSize({ width: 1280, height: 1024 })
       rfqsPage.setViewportSize({ width: 1280, height: 1024 })
       rfqBlotterPage.setViewportSize({ width: 1280, height: 1024 })
@@ -34,8 +35,8 @@ test.describe("Credit", () => {
     }
   })
 
-  test.afterEach(async ({ context }, testInfo) => {
-    if (testInfo.project.name === OPENFIN_PROJECT_NAME) {
+  test.afterEach(async ({ context }, workerInfo) => {
+    if (isOpenFin(workerInfo)) {
       const subWindowFrame = context
         .pages()
         .find((page) => page.url().includes("openfin-sub-window-frame"))
@@ -52,40 +53,8 @@ test.describe("Credit", () => {
     test.setTimeout(TestTimeout.EXTENDED)
 
     test("Create RFQ for GOOGL @smoke", async () => {
-      const MAX_ATTEMPT = 2
-
-      const clickAcceptButton = async () => {
-        const firstQuote = rfqsPage.getByTestId("quotes").locator("div").first()
-        const acceptButton = firstQuote.getByText(/Accept/)
-
-        await firstQuote.hover()
-        // Wait for first quote response
-        await expect(firstQuote).not.toContainText("Awaiting response", {
-          timeout: ElementTimeout.LONG,
-        })
-        await acceptButton.click({ timeout: ElementTimeout.AGGRESSIVE })
-      }
-
-      // retry logic to circumvent intermittent failures at clicking on the accept button
-      const acceptQuote = async () => {
-        let attempt = 0
-        let isQuoteAccepted = false
-        while (!isQuoteAccepted) {
-          // fail test if max attempt threshold is reach
-          expect(
-            attempt,
-            "Check if max attempts to click on accept button is reach",
-          ).toBeLessThan(MAX_ATTEMPT)
-          try {
-            await clickAcceptButton()
-            isQuoteAccepted = true
-          } catch (exception) {
-            console.warn(`Failed to click on the 'accept' button, retrying ...`)
-            isQuoteAccepted = false
-          }
-          attempt++
-        }
-      }
+      const firstQuote = rfqsPage.getByTestId("quotes").locator("div").first()
+      const acceptButton = firstQuote.getByText(/Accept/)
 
       await newRfqPage.getByPlaceholder(/Enter a CUSIP/).click()
       await newRfqPage
@@ -95,7 +64,7 @@ test.describe("Credit", () => {
         .click()
 
       const quantity = newRfqPage.getByTestId("quantity")
-      await quantity.type("2")
+      await quantity.pressSequentially("2")
       await quantity.blur()
 
       await newRfqPage.locator("span").getByText(/All/).click()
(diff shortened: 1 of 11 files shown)


packages/client/e2e/credit.spec.ts
```diff
@@ -49,6 +49,8 @@ test.describe("Credit", () => {
   })
 
   test.describe("New RFQ", () => {
+    test.describe.configure({ retries: 1 })
```
stephan-thibodeau, line 52: To retry once in case of failure

packages/client/e2e/credit.spec.ts
```diff
@@ -49,6 +49,8 @@ test.describe("Credit", () => {
...
+    test.setTimeout(60_000)
```
stephan-thibodeau, line 53: Extend test timeout only for this test

packages/client/e2e/credit.spec.ts
```diff
@@ -81,15 +83,15 @@ test.describe("Credit", () => {
...
       await expect(firstQuote).not.toContainText("Awaiting response", {
         timeout: Timeout.LONG,
       })
 
       await firstQuote.hover()
-
-      await firstQuote.getByText(/Accept/).waitFor({state: "visible"})
-      await firstQuote.getByText(/Accept/).click()
+      await firstQuote.getByText(/Accept/).waitFor({ state: "visible" })
+      await firstQuote.getByText(/Accept/).click({ timeout: Timeout.AGGRESSIVE })
```
stephan-thibodeau, line 94: Prior we wait for the accept button to be visible so we make the test fail after 5 seconds in case the click action didn't succeed

packages/client/e2e/credit.spec.ts
algreasley: you know what I am going to say 😉 
please add this to Timeout enum, with appropriate name - `EXTENDED` ? or `NEW_RFQ`, if it is really specialised just for this case .. this is an overall timeout for the whole test, right?

packages/client/e2e/credit.spec.ts
```diff
@@ -81,15 +82,41 @@ test.describe("Credit", () => {
...
-
-      await firstQuote.hover()
 
-      await firstQuote.getByText(/Accept/).waitFor({state: "visible"})
-      await firstQuote.getByText(/Accept/).click()
+      // retry logic to circumvent intermittent failures at clicking on the accept button
+      const MAX_ATTEMPT = 2
+      let isAcceptBtnClicked = false
+
+      const acceptQuote = async () => {
```
stephan-thibodeau, line 90: added logic to catch any error while clicking on accept button and capture state in a variable to check for retry

packages/client/e2e/credit.spec.ts
```diff
@@ -81,15 +82,41 @@ test.describe("Credit", () => {
...
+        }
+      }
+
+      await expect(firstQuote)
+        .not.toContainText("Awaiting response", {
+          timeout: Timeout.LONG,
+        })
+        .then(async () => {
+          let attempt = 1
+          while (!isAcceptBtnClicked) {
```
stephan-thibodeau, line 110: retry until `MAX_ATTEMPT` is reached, otherwise test fail

packages/client/e2e/credit.spec.ts
```diff
@@ -172,7 +172,7 @@ test.describe("Credit", () => {
 
       await expect(rfqsPage.getByTestId("quotes").first()).toContainText(
         "$100",
-        { timeout: ElementTimeout.NORMAL },
+        { timeout: ElementTimeout.LONG },
```
algreasley, line 175: 30s seems like a heck of a long time to wait for the quote ..

algreasley: especially as we are providing it above .. so within 5 seconds at the outside we should be done (really)

stephan-thibodeau: agree, will update to 5s `Element.Timeout.AGGRESSIVE`

stephan-thibodeau:
@algreasley  I have updated the retry logic based on our latest conversation.  it is ready for another review 👀

packages/client/e2e/utils.ts
```diff
@@ -1,10 +1,17 @@
-import { expect, Locator } from "@playwright/test"
+import { expect, Locator, WorkerInfo } from "@playwright/test"
 
-export const OPENFIN_PROJECT_NAME = "openfin"
-export enum Timeout {
+export const IS_OPENFIN = (workerInfo: WorkerInfo): boolean => {
```
algreasley, line 3: call it `isOpenFin` .. only use that identifier syntax for constants (a carry-over from the likes of Java etc.)
minor .. can drop the { return ..}  (simplified arrow method syntax)

packages/client/e2e/credit.spec.ts
```diff
@@ -49,7 +50,44 @@ test.describe("Credit", () => {
...
+
+      // retry logic to circumvent intermittent failures at clicking on the accept button
+      const acceptQuote = async () => {
+        let attempt = 0
+        let isQuoteAccepted = false
+        while (!isQuoteAccepted) {
+          // fail test if max attempt threshold is reach
+          expect(
+            attempt,
+            "Check if max attempts to click on accept button is reach",
```
algreasley, line 78: this is the message to [print on error](https://playwright.dev/docs/test-assertions#custom-expect-message), right?
so let's go with something like "Max accept attempts exceeded" or "Exceeded maximum attempts to click on quote Accept button" which is probably better

packages/client/e2e/credit.spec.ts
```diff
@@ -49,7 +50,44 @@ test.describe("Credit", () => {
...
+          timeout: ElementTimeout.LONG,
+        })
+        await acceptButton.click({ timeout: ElementTimeout.AGGRESSIVE })
+      }
+
+      // retry logic to circumvent intermittent failures at clicking on the accept button
+      const acceptQuote = async () => {
+        let attempt = 0
+        let isQuoteAccepted = false
+        while (!isQuoteAccepted) {
```
algreasley, line 74: is [this](https://playwright.dev/docs/test-assertions#expecttopass) any better for handling the retry attempts?
I know the example is for timed retries, but we might be able to use it as a more recognisable Playwright pattern .. just pondering

packages/client/e2e/utils.ts
stephan-thibodeau: thanks Alan, I've applied both changes 👍

packages/client/e2e/credit.spec.ts
stephan-thibodeau: you are right, I've updated the message accordingly

packages/client/e2e/credit.spec.ts
```diff
@@ -58,7 +64,7 @@ test.describe("Credit", () => {
         .click()
 
       const quantity = newRfqPage.getByTestId("quantity")
-      await quantity.type("2")
+      await quantity.fill("2")
```
algreasley, line 67: a-ha - I heard about this on the Playwright update video .. need to be careful we use this in the right context, as it bypasses character input events - we do want to `type` in this case (not `fill`) as we have custom functionality which changes the formatting of the value as we type

packages/client/e2e/limit-checker.spec.ts
```diff
@@ -78,8 +83,9 @@ test.describe("Limit Checker", () => {
     })
   })
 
-  test("Trade is blocked if notional is above limit", async ({}, testInfo) => {
-    test.skip(testInfo.project.name !== OPENFIN_PROJECT_NAME, "Openfin Only")
+ // eslint-disable-next-line no-empty-pattern
+  test("Trade is blocked if notional is above limit", async ({}, workerInfo) => {
+    if (!isOpenFin(workerInfo)) test.skip()
```
algreasley, line 88: just noticed this ... what we had before - [conditional skip](https://playwright.dev/docs/test-annotations#conditionally-skip-a-test) - is better (conventions again ..)

packages/client/e2e/spot-tile.spec.ts
```diff
@@ -85,9 +85,11 @@ test.describe("Spot Tile", () => {
       await tilePage.locator("[data-testid='rfqButton']").click()
 
       await expect(tilePage.getByTestId("rfqTimer")).toBeVisible()
-      await tilePage.getByTestId("rfqTimer").waitFor({state: "hidden", timeout: 10500})
+      await tilePage
+        .getByTestId("rfqTimer")
+        .waitFor({ state: "hidden", timeout: 10500 })
```
algreasley, line 90: hard-coded timeouts

packages/client/e2e/credit.spec.ts
stephan-thibodeau: good point, I've updated all deprecated `type` to `pressSequentially` throughout the tests suite 👍

packages/client/e2e/limit-checker.spec.ts
stephan-thibodeau: I moved all `test.skip` occurrences to the former syntax 👍

algreasley (approved):

PR #2289, opened 2023-09-26
chore: Update playwright from 1.37 > 1.38 

stephan-thibodeau:
Updated playwright from `1.37.1` > `1.38.1`

- `1.38.1` requires to explicitly install required browsers.  I've updated the specific ci workflows to download `chromium` prior to run the test
- Removing `playwright` dev-dependency and migrate import usage to `@playwright/test`
- Updated `README` to explicitly mention the need to install `chromium` if it never been installed

Tested locally for both web and openfin ✅


packages/client/playwright.config.ts
```diff
@@ -4,7 +4,7 @@ import { devices } from "@playwright/test"
 const config: PlaywrightTestConfig = {
   testDir: "./e2e",
   /* Maximum time one test can run for. */
-  timeout: 30_000,
+  timeout: 60_000,
```
stephan-thibodeau, line 7: Set temporarily to 60 seconds as 30 is too aggressive.  Permanent timeout changes to be done as part of https://github.com/AdaptiveConsulting/ReactiveTraderCloud/pull/2288

packages/client/package.json
```diff
@@ -134,7 +134,6 @@
     "jest-styled-components": "^7.1.1",
     "jsdom": "^21.1.1",
     "openfin-cli": "^4.0.0",
-    "playwright": "1.37.1",
```
stephan-thibodeau, line 137: No need to `Playwright` anymore

.github/workflows/e2e-full.yml
```diff
@@ -25,7 +25,9 @@ jobs:
           node-version: 20
 
       - name: Install
-        run: npm ci
+        run: |
+          npm ci 
+          npx playwright install --with-deps chromium
```
stephan-thibodeau, line 30: installing chromium only when

packages/client/package.json
algreasley: Something has changed since we discussed this and checked `package-lock.json` diffs locally ..
Now the `playwright` dependency is installed regardless as a _transitive_ dependency of `@playwright/test`
Check the diff - it is tiny compared to the last submission of this PR.
(which was when you replaced the `package-lock.json` rather than updating it)
Might need a closer look ..

.github/workflows/e2e-full.yml
algreasley: see https://playwright.dev/docs/browsers#google-chrome--microsoft-edge
we can run against the Chrome that is surely available on CI .. let's try it at least
certainly should not be installing _every_ time we run a job (surely?)

.github/workflows/pull.yml
```diff
@@ -188,8 +190,10 @@ jobs:
         with:
           node-version: 20.5.1
 
-      - name: Install dependencies
-        run: npm ci
+      - name: Install
```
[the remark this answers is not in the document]
stephan-thibodeau: @algreasley I tried running both web and openfin e2e tests using already installed chrome version and it works so I removed chromium install 👍

.github/workflows/pull.yml
```diff
@@ -218,8 +222,10 @@ jobs:
...
```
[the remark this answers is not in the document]
stephan-thibodeau: We currently run openfin `fx` and `credit` on 2 different instances due to the performance issue we encountered while having both `fx` and `credit` windows opened on the same CI instance

.github/workflows/pull.yml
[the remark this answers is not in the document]
algreasley: I still see the installs in all workflows ?

packages/client/package.json
algreasley: ok, I think this is fine, now you have the minimal updates to lock file based on removing the explicit `playwright` dependency

.github/workflows/pull.yml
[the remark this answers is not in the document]
algreasley: yes, but we do not need **browser** install for OpenFin tests

algreasley (approved):
nice and neat and up-to-date 😄

PR #2290, opened 2023-09-27
fix: blotter highlighting wrong row or not at all

algreasley:


packages/client/src/client/App/Trades/TradesState/tableTrades.ts
```diff
@@ -357,33 +359,23 @@ export const [creditTradeRowHighlight$, setCreditTradeRowHighlight] =
...
-    { stateOfWorld: true, trades: [], skip: false } as {
-      stateOfWorld: boolean
-      trades: TradeType[]
-      skip: boolean
-    },
-  ),
-  filter(
-    ({ stateOfWorld, trades, skip }) =>
-      !stateOfWorld && trades.length > 0 && !skip,
+const newTradeId$ = tradesStream$.pipe(
```
algreasley, line 362: simplified new trade ID capture streams, going back to the more fundamental source streams, which retain the "state of the world" flag for FX and the appropriate quoteAccepted update for Credit

packages/client/src/services/credit/creditRfqs.ts
```diff
@@ -72,18 +71,24 @@ export const creditRfqsById$ = creditRfqUpdates$.pipe(
     [boolean, Record<number, RfqDetails>]
   >(
     (acc, [update, instruments, dealers]) => {
-      const rec = acc[1]
+      // need to know when we can release updates from this stream
```
algreasley, line 74: looks like a lot of changes here, but just a bit of renaming and simplification of each case
(it used to be a bigger change, but then I discovered I could just use the source stream `creditRfqUpdates$` to drive the new ID for a "trade" or accepted quote)

packages/client/src/services/credit/creditRfqs.ts
```diff
@@ -166,40 +171,32 @@ export const creditRfqsById$ = creditRfqUpdates$.pipe(
...
-              {
-                ...rec,
-                [previousRfq.id]: {
-                  ...previousRfq,
-                  quotes: previousRfq.quotes.map((quote) => ({
-                    ...quote,
-                    state: getQuoteStateOnAccept(quote, update.payload),
-                  })),
-                },
+          const previousRfq = rfqs[update.payload.rfqId]
```
algreasley, line 175: like all the other quote updates .. not sure why that slightly convoluted way of grabbing the prev RFQ was used just for the Accept

snony14 (dismissed):

snony14 (approved):

PR #2291, opened 2023-09-28
fix: credit start sow, restore reset on re-connect

algreasley:
"cleanup" during blotter highlighting work introduced a subtle but lethal bug in the reconnect flow, where the RFQ store was not reset ... this has been restored
(for future ref, this hit us in an Error thrown in `useQuoteState` .. might want to make this a bit more lenient in future)


packages/client/src/services/credit/creditRfqs.ts
```diff
@@ -79,7 +79,7 @@ export const creditRfqsById$ = creditRfqUpdates$.pipe(
       switch (update.type) {
         case START_OF_STATE_OF_THE_WORLD_RFQ_UPDATE:
           // init state of scan is correct for start of SoW
-          return acc
+          return [false, {}]
```
algreasley, line 82: turns out this is _quite_ important .. 😬

snony14 (approved):

PR #2292, opened 2023-10-03
Test: 5590 Improve "when user buys a currency, the new row should flash briefly" test resiliency

stephan-thibodeau:
Mitigation for low frequency but periodical false negative on both scheduled e2e or pull request e2e.  Since the failure occurence is erratic and is tend to be related to timing issue the countermeasure is to retry until it times out

Ran tests on both web and openfin ✅

### DIFF
--- a/packages/client/e2e/blotter.spec.ts
+++ b/packages/client/e2e/blotter.spec.ts
@@ -4,7 +4,7 @@ import { expect, Page } from "@playwright/test"
 import fs from "fs"
 
 import { test } from "./fixtures"
-import { isOpenFin } from "./utils"
+import { ElementTimeout, isOpenFin } from "./utils"
 
 const getTradeIDColIndex = () => {
   // const tradeIndex = fxColFields.indexOf(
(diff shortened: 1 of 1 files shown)


packages/client/e2e/blotter.spec.ts
```diff
@@ -55,8 +55,8 @@ test.describe("Trade Blotter", () => {
 
   // eslint-disable-next-line no-empty-pattern
   test.afterAll(async ({}, testInfo) => {
-    if (testInfo.project.name === OPENFIN_PROJECT_NAME) {
-      await blotterPage.getByTestId("filter-button").click()
+    if (await blotterPage.isVisible("[data-testid='filter-button']")) {
```
stephan-thibodeau, line 58: this `afterAll` is used to clean last test state where it add a filter. I made it conditional as the test framework execute the `afterAll` also upon test failure so the filter button may not be present in this case causing false negative.  Also made it executed for both openfin and web

From playwright documentation
https://playwright.dev/docs/api/class-test#test-after-all-1

> Details
>
> Note that worker process is restarted on test failures, and afterAll hook runs again in the new worker. Learn more about [workers and failures](https://playwright.dev/docs/test-retries).

packages/client/e2e/blotter.spec.ts
```diff
@@ -72,17 +72,45 @@ test.describe("Trade Blotter", () => {
...
+      try {
+        await tilePage.locator('[data-testid="Buy-EURUSD"]').nth(0).click()
+
+        const tradeID = await tilePage
+          .locator("[data-testid='trade-id']")
+          .innerText()
+
+        const newRow = blotterPage.getByTestId(`trades-grid-row-${tradeID}`)
+
+        await expect(newRow).toHaveCSS("animation", /1s ease-in-out/, {
```
stephan-thibodeau, line 88: will throw exception if not true

packages/client/e2e/blotter.spec.ts
```diff
@@ -72,17 +72,45 @@ test.describe("Trade Blotter", () => {
...
+          timeout: Timeout.AGGRESSIVE,
```
stephan-thibodeau, line 89: Make the assertion fail faster as it should be flashing upon trade being placed

stephan-thibodeau:
@algreasley I've rebased and updated the retry logic with latest finding.  Ready for review, thanks

algreasley (approved):

PR #2293, opened 2023-10-06
Test: 4625 Added e2e tests for analytics panel  

stephan-thibodeau:
Adding tests for analytics panel

1. Added dedicated `analytics.spec` file
2. Added tests
3. Added execution as part of openfin-fx e2e tests

Tested on both openfin and web ✅


packages/client/e2e/analytics.spec.ts
```diff
@@ -0,0 +1,145 @@
...
+          initiallastPositionAmount,
+          { timeout: ElementTimeout.NORMAL },
+        )
+    })
+
+    test("Correct text color is displayed based on position value", async () => {
+      const initialpositionLocator = analyticsPage.locator(
+        "[data-testid='lastPosition']",
+      )
+      const amount = Number(
```
algreasley, line 62: not sure what's possible .. but be careful re amount being `NaN` (non-numeric text) or `0` (for empty string)

packages/client/e2e/analytics.spec.ts
```diff
@@ -0,0 +1,145 @@
...
+
+    test.skip("Graph display coherent values", async () => {
+      // TODO
+      // more complex. may requires datatest-id to reliably assert graph content
+    })
+  })
+
+  test.describe("Positions section", () => {
+    test("Position nodes are showing tooltip information for each currencies", async () => {
+      for (const currency of currencies) {
```
algreasley, line 84: minor/thought: I don't think we need to check _all_ of them, do we?

packages/client/e2e/analytics.spec.ts
```diff
@@ -0,0 +1,145 @@
...
+        })
+        const regexp = RegExp(`${currency} [-+,.0-9]`, "g")
+        expect(
+          await currencyTooltip.textContent(),
+          `tooltip for ${currency} doesn't match expected pattern`,
+        ).toMatch(regexp)
+      }
+    })
+
+    test("Position nodes can be moved", async () => {
```
algreasley, line 105: cool - forgot about this (think it just comes with the graph tech) .. fairly useless, tho, innit 😆

packages/client/e2e/analytics.spec.ts
```diff
@@ -0,0 +1,145 @@
...
+      const nzdInitialPosition = await nzdNode.boundingBox()
+      await nzdNode.dragTo(jpyNode)
+      expect(nzdInitialPosition).not.toEqual(await nzdNode.boundingBox())
+    })
+  })
+
+  test.describe("PnL section", () => {
+    test("PnL value is displayed for each currencies", async () => {
+      const pnlSection = analyticsPage
+        .locator("div:last-of-type")
```
algreasley, line 125: do we need all the "last-of-type" / last() stuff? (trying to KISS as always)

packages/client/e2e/analytics.spec.ts
```diff
@@ -0,0 +1,145 @@
...
+          .locator("div", {
+            hasText: currencypair,
+          })
+          .first()
+          .getByTestId("priceLabel")
+          .textContent()
+        const regexp = RegExp(`[-,.0-9km]`, "g")
+        expect(
+          amountString,
+          `amount for ${currencypair} doesn't match expected pattern`,
```
algreasley, line 140: minor: could do bit more accurate label for this - "expected" -> "abbreviated numerical"

.github/workflows/pull.yml
```diff
@@ -197,7 +197,7 @@ jobs:
       - name: Run tests
         env:
           E2E_RTC_WEB_ROOT_URL: https://openfin.env.reactivetrader.com/pull/${{ github.event.number }}
-        run: npm run e2e:openfin spot-tile.spec blotter.spec
+        run: npm run e2e:openfin spot-tile.spec blotter.spec analytics.spec
```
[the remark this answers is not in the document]
stephan-thibodeau: Agree, updated to `fx-analytics.spec`

packages/client/e2e/analytics.spec.ts
```diff
@@ -0,0 +1,145 @@
+import { expect, Page } from "@playwright/test"
+
+import { test } from "./fixtures"
+import { ElementTimeout, isOpenFin } from "./utils"
+
+const currencyPairs = [
```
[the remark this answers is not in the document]
stephan-thibodeau: good advise.

packages/client/e2e/analytics.spec.ts
```diff
@@ -0,0 +1,145 @@
...
+        expect(await initialpositionLocator.getAttribute("color")).toEqual(
+          "negative",
+        )
+      } else
+        expect(await initialpositionLocator.getAttribute("color")).toEqual(
+          "positive",
+        )
+    })
+
+    test.skip("Graph display coherent values", async () => {
```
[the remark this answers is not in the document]
stephan-thibodeau: +1.  I created a dedicated ticket and added the reference in there.

packages/client/e2e/analytics.spec.ts
```diff
@@ -0,0 +1,145 @@
...
+      expect(lastPositionAmount).toMatch(regexp)
+    })
+
+    test("Last position amount is updated periodically", async () => {
+      const lastpositionLocator = analyticsPage.locator(
+        "[data-testid='lastPosition']",
+      )
+      const initiallastPositionAmount = await lastpositionLocator.textContent()
+
+      if (initiallastPositionAmount)
```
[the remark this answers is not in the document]
stephan-thibodeau: Yes the test is fully reliable. From my observation the update takes 1<8 seconds so this is probably every ~ 10 seconds. Having `NORMAL` 15 seconds make the tests fail in a reasonable time if it doesn't update

packages/client/e2e/analytics.spec.ts
stephan-thibodeau: Good point, I added a check to make the test fail if the value is not truthy

packages/client/e2e/analytics.spec.ts
stephan-thibodeau: Since it is not time costly my thought is it worth checking them all to cover the case where we get data corruption for a single currency

packages/client/e2e/analytics.spec.ts
stephan-thibodeau: thank for raising, I've updated with a better locator strategy that avoid using css pseudo-class

packages/client/e2e/analytics.spec.ts
stephan-thibodeau: sure! I,ve updated the error msg as per suggested

.github/workflows/pull.yml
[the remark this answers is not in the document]
algreasley: thanks .. I meant all of them, tho 😉

packages/client/e2e/analytics.spec.ts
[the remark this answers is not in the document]
algreasley: maybe 😄  ..  I think later, tho, given my note about impracticality

packages/client/e2e/analytics.spec.ts
algreasley: ah, these are numbers we are talking about ... so their boolean equivalent is not relevant (and a bit fuzzy/dangerous, tbh)
0 is perfectly valid (falsy)
I think what I want here is a guard/expect to handle NaN (may seem unlikely but that is what we had the other day)
also, I am thinking we need to check for blank text .. as that will convert happily to 0 but is definitely a failure, whereas 0 itself is not.

algreasley: I know you are only reducing to +ve or -ve, but we should still try to use code that renders the actual number - it might be copied later on! So, as we don't do locale variations, we can just strip the comma to get a parseable representation of the number (for fun, I'll try and come up with a locale-sensitive version, in case we ever extend the UI to show French, German etc. number formats ..

packages/client/e2e/analytics.spec.ts
algreasley: I'm curious now .. what does the Playwright UI say if you select that element?

algreasley: .. and are we going to transition to Page Objects, as I understand them .. we could do this progressively
e.g. this could be replaced by PO.getPnLSection(), I think ?

stephan-thibodeau: Here is the `pnlSection` selection in Playwright UI
Screenshot 2023-10-17 at 11 31 12
when it is chained the locator looks like 
`locator('div').filter({ has: locator('div').filter({ hasText: 'PnL' }) }).last().locator('div').filter({ hasText: 'EURCAD' }).first().getByTestId('priceLabel')`
Screenshot 2023-10-17 at 11 35 16

stephan-thibodeau: I agree that we are lacking implementation of Page object pattern.  I suggest that we tackle the whole PO implementation as part of the dedicated ticket to be prioritised shortly

packages/client/e2e/analytics.spec.ts
stephan-thibodeau: I updated the test based on your feedback so we handle all cases gracefully, thanks

.github/workflows/pull.yml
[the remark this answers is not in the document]
stephan-thibodeau: Make sense, done for all fx spec files ✅

algreasley (approved):
looks great @stephan-thibodeau - thanks for filling this gap in our checks

PR #2295, opened 2023-10-12
Chore: added condition to avoid build/deploy/test against draft pr

stephan-thibodeau:
Avoid running build/deploy/test on draft PR.  The skipped steps are then executed once the PR is moved to `review` state

### DIFF
--- a/.github/workflows/pull.yml
+++ b/.github/workflows/pull.yml
@@ -5,6 +5,8 @@ on:
     types:
       - opened
       - synchronize
+      - reopened
+      - ready_for_review
 
 defaults:
   run:
@@ -21,8 +23,7 @@ jobs:
     name: Build & deploy
 
     runs-on: ubuntu-20.04
-
-    if: "!github.event.pull_request.head.repo.fork"
+    if: ${{ !github.event.pull_request.head.repo.fork && !github.event.pull_request.draft }}
 
     steps:
       - name: Checkout code
@@ -47,8 +48,7 @@ jobs:
         run: npm ci
 
       - name: Verify Quality (type checking, linting, format checking, tests)
-        run: |
-          npm run verify
+        run: npm run verify
 
       - name: Build (web)
         env:
@@ -67,7 +67,6 @@ jobs:
             gs://$BUCKET_WEB/pull/${{ github.event.number }}
 
       # No need to deploy legacy Launcher as branch/tag workflows
-
       - name: Build (OpenFin)
         env:
           DOMAIN: https://openfin.env.reactivetrader.com
@@ -87,8 +86,7 @@ jobs:
           DOMAIN: https://openfin.env.reactivetrader.com
           URL_PATH: /pull/${{ github.event.number }}/workspace
           VITE_BUILD_VERSION: ${{ github.sha }}
-        run: |
-          npm run workspace:build
+        run: npm run workspace:build
 
       - name: Deploy (OpenFin - Workspace)
         run: |
@@ -147,7 +145,6 @@ jobs:
 
   web-end-to-end-test:
     name: Web e2e test - All
-
     needs: build
     runs-on: ubuntu-20.04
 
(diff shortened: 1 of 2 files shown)


.github/workflows/pull.yml
```diff
@@ -33,24 +34,27 @@ jobs:
...
 
+      - name: Install
+        run: npm ci
+
+      - name: Verify Quality (type checking, linting, format checking, tests)
+        run: |
+          npm run verify
+  
       - name: "Authenticate to Google Cloud"
+        if: ${{ github.event.pull_request.draft == false }}
```
algreasley, line 45: you should be able to put it above, with the other `if` on L26 .. which, per [docs](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idif), should be like
```
if: ${{ ! github.event.pull_request.head.repo.fork }}
```

algreasley: e.g. (not sure if this will work)
```
if: ${{ !github.event.pull_request.head.repo.fork && !github.event.pull_request.draft }}
```

.github/workflows/pull.yml
```diff
@@ -5,6 +5,7 @@ on:
     types:
       - opened
       - synchronize
+      - ready_for_review
```
algreasley, line 8: should not need - synchronize (last push) will have run the workflow already
in fact, remove all types .. [per first note in docs](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request), the default is already opened/synchronize,reopened

.github/workflows/pull.yml
algreasley: can you change the ordering back too .. so that the setup (GCloud stuff is out of the way before we start doing npm tasks ... while you are here, we don't need to split lines for single commands (old stuff) e.g.
```
run: npm run verify
```

stephan-thibodeau: @algreasley sure.  It make sense to set the condition at the jobs level if we want to skip all the jobs. my initial assumption was that we wanted to still perform npm install and npm verify but these checks are also done on branch level so we are covered ✅

.github/workflows/pull.yml
stephan-thibodeau: I will give it a try.  in my early testing the workflow was not triggered when PR is moved from `draft` -> `ready for review`.

stephan-thibodeau: @algreasley based on my latest tests if we are not explicitly setting activity type `ready_for_review` then we are not covered for the scenario where we push the latest as a Draft, then move PR to Ready for review as nothing trigger the `synchronize` activity.

stephan-thibodeau: I kept the PR without the activity types for further testing together, I will add them if we conclude that they are required

algreasley: ahh, I see .. as the usual "synchronize" will be bypassed by the "PR draft exclusion" ...

algreasley: so go with default, as my initial comment, and add the `ready_for_review` .. my mistake, sorry

.github/workflows/pull.yml
```diff
@@ -120,7 +114,6 @@ jobs:
           body-includes: "(auto-deploy)"
 
       - name: Create comment (if not exists)
-        if: ${{ steps.comment.outputs.comment-id == 0 }}
```
algreasley, line 123: need to re-instate this

.github/workflows/pull.yml
stephan-thibodeau: I explicitly added the defaults one too as per [doc](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#using-activity-types)

.github/workflows/pull.yml
stephan-thibodeau: corrected, thanks

PR #2296, opened 2023-10-17
chore: Serialise web and openfin e2e tests execution

stephan-thibodeau:
Serializing `web` -> `openfin` e2e tests on a PR basis to prevent intermittent concurrency issues that cause false negative results

Screenshot 2023-10-17 at 14 44 58

### DIFF
--- a/packages/client/e2e/credit.spec.ts
+++ b/packages/client/e2e/credit.spec.ts
@@ -148,7 +148,7 @@ test.describe("Credit", () => {
 
       await sellSidePage.waitForSelector("text=New RFQ")
 
-      await sellSidePage.getByTestId("price-input").fill("100")
+      await sellSidePage.getByTestId("price-input").pressSequentially("100")
 
       await sellSidePage.keyboard.press("Enter")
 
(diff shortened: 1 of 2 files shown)


packages/client/e2e/credit.spec.ts
```diff
@@ -148,7 +148,7 @@ test.describe("Credit", () => {
 
       await sellSidePage.waitForSelector("text=New RFQ")
 
-      await sellSidePage.getByTestId("price-input").fill("100")
+      await sellSidePage.getByTestId("price-input").pressSequentially("100")
```
algreasley, line 151: nice - are there any more lying around?
in some cases a `.fill` will be appropriate, but maybe as a parallel test, to check what happens if people paste values in - can't think of one off-hand in RT, mind you .. RA perhaps?
(I am referring to the Playwright 1.38 update video)

algreasley (approved):
Like you, I'd prefer if we could do these tests in parallel, especially as they are in a Pull Request workflow .. if this becomes onerous, we could consider bypassing the OpenFin tests, but make sure we run them on master (this would require a bit of gnarly logic in the branch workflow tho, so far from ideal)

packages/client/e2e/credit.spec.ts
stephan-thibodeau: There is no explicit test validating how pasting a value in a text field behave, either in RT or RA. There are still few `.fill` being used in RT.  I will update them to `.pressSequentially` in a subsequent PR

PR #2302, opened 2023-11-13
Final Version for Spot Tile Tests 

fabio-weareadaptive:
This is final version for the Spot Tile Tests written on the revamped Playwright framework.

### DIFF
--- a/packages/client/playwright/components/spotTile.component.ts
+++ b/packages/client/playwright/components/spotTile.component.ts
@@ -1,6 +1,7 @@
 import { Locator, Page } from "@playwright/test"
 
 import { BaseComponent } from "../Base.component"
+import { GrReturn } from "react-icons/all"
 
 export enum CurrencyPair {
   EURUSD = "EUR/USD",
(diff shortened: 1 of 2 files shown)


packages/client/playwright/components/spotTile.component.ts
```diff
@@ -24,10 +25,19 @@ export default class SpotTileComponent extends BaseComponent {
...
   }
 
+  public getToggle() {
+    return this.page.getByTestId("toggleButton")
+  }
+
+  public async getTileSate() {
+    const tileState = await this.page.evaluate(() => {
+      return window.localStorage.getItem("selectedView")
+    })
```
stephan-thibodeau, line 35: I may be wrong but I think you only need to return `window.window.localStorage.getItem("selectedView")`

packages/client/playwright/tests/spot-tile.spec.ts
```diff
@@ -133,4 +133,34 @@ test.describe("FX purchase", () => {
       await expect(eurusdTile.getMaxExceedError).not.toBeVisible()
     })
   })
+  test.describe("Toggle between prices and graph views", () => {
+    test("When I click the graph icon on the Live Rates bar then I should toggle from graph to price views", async ({}) => {
```
stephan-thibodeau, line 137: 💯

packages/client/playwright/components/spotTile.component.ts
```diff
@@ -1,6 +1,7 @@
 import { Locator, Page } from "@playwright/test"
 
 import { BaseComponent } from "../Base.component"
+import { GrReturn } from "react-icons/all"
```
stephan-thibodeau, line 4: This import is not needed

stephan-thibodeau (changes requested):

packages/client/playwright/components/spotTile.component.ts
fabio-weareadaptive: Removed ✅

packages/client/playwright/components/spotTile.component.ts
fabio-weareadaptive: You're right, I've simplified it accordingly to keep it cleaner 👌🏼

stephan-thibodeau (approved):

PR #2305, opened 2024-06-04
chore: retire Reactive Analytics

algreasley:
Remove all traces of Reactive Analytics from the OpenFin Launcher and Workspace variants
Plus some extras:
- highlight remaining Interop case (Live Rates <-> Limit Checker) - tested in Launcher and Workspace
 (where we need to switch user to open the other view
 e.g. launch Live Rates view as Trader then `/switch user Operations` in Home, and then launch Limit Checker view if you started out on the default Trader user)
- add in various missing images and correct ones like Limit Checker and RT Credit which had old/incorrect images

Tested locally and through my "test installers" (deployments to `test/openfin` branch)


packages/client/src/workspace/store.ts
```diff
@@ -175,79 +166,86 @@ async function getLandingPage(): Promise<StorefrontLandingPage> {
...
+      apps:
+        currentUser === USER_TRADER
+          ? [
+              {
+                ...reactiveTraderFxTradesView,
+                description:
+                  "FX Trades view - combine with Limit Checker view (Operations user)",
+                icons: [
+                  {
+                    src: `${WS_BASE_URL}/images/previews/interop-with-views-color-linking.png`,
```
algreasley, line 217: only way to get a "preview" on the middle row (don't ask me why) is to push something into the icon space .. works, but a bit squashed

packages/client/src/workspace/utils.ts
```diff
@@ -4,25 +4,39 @@ import { firstValueFrom } from "rxjs"
...
+    title: `${symbol} Spot Tile`,
+    description: `Reactive Trader - Live rates and trade ${symbol} in this view`,
     manifestType: "url",
     manifest: `${VITE_RT_URL}/fx-spot/${symbol}`,
-    title: `${symbol} Spot Tile`,
     icons: [{ src: ADAPTIVE_LOGO }],
     publisher: "Adaptive Financial Consulting",
-    description: `View ${symbol} live rates`,
+    tags: ["FX", "Trading", "Market Data"],
+    images: [{ src: `${WS_BASE_URL}/images/previews/GBPUSD.png` }],
```
algreasley, line 40: overdue a preview image for the ccy pair views

packages/client/src/workspace/utils.ts
```diff
@@ -4,25 +4,39 @@ import { firstValueFrom } from "rxjs"
...
+  if (currentUser !== USER_TRADER) {
+    return []
+  }
   const currencyPairs = await firstValueFrom(currencyPairSymbols$)
 
   return currencyPairs.map((symbol) => ({
     appId: `reactive-trader-${symbol}`,
+    title: `${symbol} Spot Tile`,
+    description: `Reactive Trader - Live rates and trade ${symbol} in this view`,
     manifestType: "url",
```
algreasley, line 35: bit of an anomaly, launched as a view but this is not a `manifestType` known to the OpenFin types, but need this to open a view without a manifest (as the ccypairs are dynamic) .. guessing some of this as this work predated my time on RT

packages/client/vite.config.ts
```diff
@@ -220,12 +216,12 @@ const copyPlugin = (
               transform,
             },
             {
-              src: "public-workspace/images/icons/*",
-              dest: "images/icons",
+              src: "public-workspace/images/*",
```
algreasley, line 219: not sure why this was arranged like this .. probably another gaffe of mine from the client/workspace convergence .. should all be ship-shape with this mod

SHolleworth (approved):

algreasley:
Merging per @SHolleworth's earlier approval (just modified OpenFin testing command/dependency version so E2E now passes

PR #2306, opened 2024-06-11
chore: update hydra-platform to 4.57.0, rework connection

algreasley:
- Update to latest hydra web libs
- Use direct Hydra address (wss://...) rather than Nginx or local proxy - that part of deployment to be retired soon
- Separate codegen as dev dependency
- re-run codegen
- re-work connection to use new connectToTradingGateway
See web deployment [here](https://web.env.reactivetrader.com/branch/chore_update-hydra)

### DIFF
--- a/packages/client/package.json
+++ b/packages/client/package.json
@@ -58,17 +58,17 @@
     "workspace:start:preview": "npm:openfin:build && concurrently \"npm:serve\" \"npm:workspace:run\"",
     "finsemble:dev": "cross-env TARGET=finsemble vite",
     "finsemble:build": "cross-env TARGET=finsemble vite build",
-    "generateCod": "hydra-web-codegen -i trading-gateway.hyer -o src/generated",
     "storybook": "storybook dev -p 6006",
     "storybook:build": "cross-env STORYBOOK=true storybook build -o ./dist/storybook",
     "_e2e:openfin:run": "cross-env-shell of-automation $npm_config_manifest_url ./e2e/openfin.spec.js --chromeDriverOverride=122 --devToolsPort=9091 --closeRuntime=never",
     "e2e:openfin": "playwright test --project=openfin",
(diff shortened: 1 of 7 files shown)


packages/client/src/services/connection.ts
```diff
@@ -36,37 +35,39 @@ export enum ConnectionStatus {
...
 }
 
-// Connect to Hydra gateway and store the disposable
-export const initConnection = async () => {
-  const dispose = await connectToGateway({
-    url: `${window.location.origin}/ws`,
+// Connect to Hydra gateway and store the disconnect for idle timeout handling
+export const initConnection = () => {
+  const { connection } = connectToTradingGateway({
+    url: import.meta.env.VITE_HYDRA_URL,
```
algreasley, line 47: this is the main change really ... use direct comms to Hydra trading gateway

packages/client/package.json
```diff
@@ -58,17 +58,17 @@
...
     "e2e:openfin": "playwright test --project=openfin",
     "e2e:openfin:smoke": "playwright test --project=openfin --grep @smoke",
     "e2e:web": "playwright test --project=chrome",
-    "e2e:web:smoke": "playwright test --project=chrome --grep @smoke"
+    "e2e:web:smoke": "playwright test --project=chrome --grep @smoke",
+    "generateCod": "hydra-web-codegen -i ../../trading-gateway.hyer -o src/generated"
   },
   "dependencies": {
-    "@adaptive/hydra-platform": "4.12.0",
+    "@adaptive/hydra-platform": "4.57.0",
```
algreasley, line 71: this is latest .. 4.54 introduced the codegen separation .. the major differences in `TradingGateway.ts` came from many more versions back - PaulS mods to services generation.

TobyLi96 (approved):

PR #2317, opened 2024-07-23
chore: fix base path for prod as reactivetrader.com

algreasley:


packages/client/src/client/utils/constructUrl.ts
```diff
@@ -1,8 +1,11 @@
-import { BASE_URL } from "../constants"
+import { BASE_PATH } from "../constants"
 
 export const constructUrl = (path: string): string => {
-  if (BASE_URL.endsWith("/") && path.startsWith("/")) {
-    return `${BASE_URL}${path.slice(1)}`
+  const baseUrl = `${window.location.origin}${BASE_PATH}`
```
algreasley, line 4: Prod does not like the base as "https://web.prod.reactivetrader.com" ..

algreasley: I could rework BASE_URL entirely, as it is a bit wacky setting that then overriding it completely here with `window.location.origin` .. this is as per orig code prior to build "tidy up"

rikoe (approved):

PR #2319, opened 2024-08-08
OpenFin Workspace latest beta

algreasley:
Tests Ok from installers.

In dev workflow, runtime 36+ - apps run up fine first time on Mac - just fails with EPERM on second and subsequent attempts to launch from OF CLI - OF know about this, something to do with extra controls on directory perms in later OSXes (something like that) - they suggested launching the app directly through the RVM, which we could try I suppose ...
Everything works as normal on Windows (11)

_Note: Latest WS is v20 / Runtime 40.x , so would not push to UAT/Prod until they flagged that as stable._

From RVM 14 / Runtime 36, see [release notes](https://cdn.openfin.co/versions/?product=RVM#14.0.0.152)

**_Universal macOS RVM Binary_**

_Starting in RVM 14, a universal RVM binary will be available for macOS. The binary runs natively on both Apple Silicon and Intel-based Mac computers, because it contains executable code for both architectures._

**_Native macOS ARM64 Runtime Support_**

_Starting in OpenFin Runtime 36.122.80.10 we are producing native Apple Silicon (ARM64) builds and starting with RVM 14 we will select the appropriate binaries for each architecture, ensuring optimal execution on Apple Silicon devices._

### DIFF
--- a/packages/client/install/README.md
+++ b/packages/client/install/README.md
@@ -1,75 +1,61 @@
 # Installers
 
-Reactive Trader® supports multiple desktop platforms, including [OpenFin] and [Finsemble].
+Reactive Trader® supports multiple desktop platforms, including [OpenFin](https://www.openfin.co/), now known by the brand name **Here&trade;** and [Finsemble](https://documentation.finsemble.com/) now known under the product name [io.connect](https://interop.io/products/io-connect/).
 
 Below you can find links to the installers for the various platforms.
 
-[openfin]: https://openfin.co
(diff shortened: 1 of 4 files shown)


algreasley:
Just not working with playwright / chromedriver / that version of chromium -- tried lots of things, but it just will not detect the Sell Side window, so the only solution would be to force the Sell Side to open in a previous test, then subsequent tests will work - I am just not prepared to hack tests like that and force them to be serialised just because of a deficiency in the platform.

PR #2320, opened 2024-08-09
chore: eslint v9

algreasley:


packages/client/e2e/fx-blotter.spec.ts
```diff
@@ -157,7 +157,7 @@ test.describe("Trade Blotter", () => {
         .readFileSync("e2e/test-data/blotter-data.csv", "utf8")
         .toString()
         .split("\n") as Array<string>
-    } catch (err) {
+    } catch (_err) {
```
grancalavera, line 160:
```suggestion
    } catch {
```

grancalavera (dismissed):

packages/client/e2e/fixtures.ts
```diff
@@ -81,7 +81,7 @@ export const test = base.extend<IPlaywrightFixtures>({
         )
       }
       await use(contexts[0])
-    } catch (e) {
+    } catch (_e) {
```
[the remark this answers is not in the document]
algreasley: hah, yes, that would have been a more sensible option, wouldn't it?

packages/client/e2e/limit-checker.spec.ts
```diff
@@ -1,30 +1,23 @@
...
   let limitCheckerPage: Page
 
   test.beforeAll(async ({ fxPagesRec, limitCheckerPageRec }, workerInfo) => {
-    if (isOpenFin(workerInfo)) {
-      tilePage = fxPagesRec["fx-tiles"]
-      blotterPage = fxPagesRec["fx-blotter"]
-      limitCheckerPage = limitCheckerPageRec
-    }
-  })
+    test.fixme(!isOpenFin(workerInfo), "Openfin Only")
```
algreasley, line 13: this whole thing is Openfin Only .. not sure it will ever be otherwise ...

packages/client/e2e/limit-checker.spec.ts
```diff
@@ -57,15 +48,16 @@ test.describe("Limit Checker", () => {
...
+      tradeBlotterFirstRowCells
+        .nth(1)
+        .getByText(String(isNaN(lastTradeId) ? 1 : lastTradeId + 1)),
+    ).toBeVisible()
 
     await assertGridRow({
       row: limitTableFirstRowCells,
       assertions: [
-        limitId + 1 + "",
+        String(isNaN(limitId) ? 0 : limitId + 1),
```
algreasley, line 60: cure the playwright lint conditional warnings .. this is not "conditional testing", just adjusting expected values according to circumstances .. which can't be a wrong thing to do

packages/client/eslint.config.js
```diff
@@ -1,11 +1,17 @@
+import { fixupPluginRules } from "@eslint/compat"
 import js from "@eslint/js"
+import pluginVitest from "@vitest/eslint-plugin"
```
algreasley, line 3: adds vitest and playwright plugins, applied only in the restricted config blocks towards the end of the config

packages/client/src/workspace/home/index.ts
```diff
@@ -85,22 +86,34 @@ export async function registerHome(): Promise<HomeRegistration> {
...
-      await launchPage(pageToLaunch)
+      if (pageToLaunch) {
+        await launchPage(pageToLaunch)
+      } else {
+        console.warn(`No page found for ${result.data.pageId}`)
+      }
     }
   }
 
+  const isPageData = (
```
algreasley, line 109: bit of "long hand" typing .. to tie the page data with its source in `utils.ts` and give a narrowed type for the page search result, as it is totally different from app/view result, at the "data" level

PR #2326, opened 2024-10-29
chore: npm auth for uisk package install

algreasley:


algreasley:
no need .. UISK package can just be added in "optionalDependencies"

PR #2332, opened 2024-11-27
chore: ci limits on building finsemble, and clean up

algreasley:


.github/workflows/tag.yml
```diff
@@ -83,13 +83,6 @@ jobs:
             rsync -d -r dist \
             gs://$BUCKET_OPENFIN/tag/${{ steps.vars.outputs.TAG }}
 
-      - name: Deploy (Launcher)
```
algreasley, line 86: about time we retired this .. very old installations used to point to `launcher.prod.reactivetrader.com` ..

.github/workflows/pull.yml
```diff
@@ -156,15 +140,13 @@ jobs:
...
         with:
-          name: playwright-report
+          name: playwright-report-web
           path: packages/client/playwright-report/
           retention-days: 7
 
   openfin-end-to-end-test-fx:
     name: Openfin e2e test - FX
 
-    needs: web-end-to-end-test
```
algreasley, line 166: might as well kick them all off in parallel, no real dependency here

.github/workflows/pull.yml
```diff
@@ -191,15 +173,13 @@ jobs:
         uses: actions/upload-artifact@v4
         if: failure()
         with:
-          name: playwright-report
```
algreasley, line 194: this would have caused a clash in the very unlikely event that we had failures in both at once .. hasn't happened, but good practice, esp as I have unleashed all E2Es at once 👿

.github/workflows/e2e-full.yml
```diff
@@ -55,17 +55,3 @@ jobs:
           name: playwright-report-uat
           path: packages/client/playwright-report/
           retention-days: 7
-
```
algreasley, line 58: start to wind back the relentless full E2E-ing we are doing atm (pushed directly to master), looks like the web credit tests are working consistently now (although I have done nothing to fix them .. anyway the UAT deployment is the same commit as Prod ..

.github/workflows/branch.yml
```diff
@@ -90,6 +90,7 @@ jobs:
             gs://$BUCKET_OPENFIN/branch/${{ steps.vars.outputs.BRANCH }}
 
       - name: Build (Finsemble)
+        if: ${{ steps.vars.outputs.ENVIRONMENT == 'dev' || steps.vars.outputs.ENVIRONMENT == 'uat' }}
```
algreasley, line 93: only build Finsemble for Dev, UAT and Prod .. and consider dropping it entirely in the new year .. if we pick it up again (unlikely), we would probably look at using the [io.Connect](https://docs.interop.io/desktop/getting-started/how-to/migrate-to-io-connect/index.html) (i.e. rebadged Glue 42) product which has superseded Finsemble now (there is a migration path, I am told).

.github/workflows/pull.yml
algreasley: vague chance that the OpenFin tests could interfere with Web ones as they are a bunch of dups, just running against OpenFin deployment (same BE ofc) .. it's working as it is, so I'd like to keep it like this in the interests of speed (takes a while to boot up OpenFin anyway) .. but if we get clashes, we can drop back to waiting for the Web E2E to finish, as before

PR #2343, opened 2025-01-20
chore: run E2E from Prod code against Prod deployment

algreasley:
use static tag to identify Prod current commit - and check this code out in the Prod E2E workflow


.github/workflows/pull.yml
algreasley, line 1: just temp stuff in PR workflow .. to test

PR #2359, opened 2025-03-10
Updated functions for selecting Live and All rates for responsiveNav

umarkayani-adaptive:
* Added new 'project' for responsiveNav with set viewport
* Added data-testid for TabBar to easily access via Playwright


packages/client/e2e/utils.ts
```diff
@@ -3,6 +3,9 @@ import { expect, Locator, WorkerInfo } from "@playwright/test"
 export const isOpenFin = (workerInfo: WorkerInfo) =>
   workerInfo.project.name === "openfin"
 
+export const isResponsive = (workerInfo: WorkerInfo) =>
```
algreasley, line 6: ok, for now .. but we should add `openfinResponsive` project too .. and then `isOpenFin` will need to change as well, obviously .. this is not an academic point, I ran up OpenFin without maximise and the viewport settings, as I was trying to remove that kind of thing (and use of `workerInfo`) from the actual spec file .. it failed as the smaller window triggered the responsive Tab -> Dropdown 😐

packages/client/playwright.config.ts
```diff
@@ -24,6 +24,15 @@ const config: PlaywrightTestConfig = {
         trace: "retain-on-failure",
       },
     },
+    {
+      name: "webResponsive",
+      use: {
+        viewport: {width: 900, height: 1024},
+        //Artifacts
```
algreasley, line 31: at some point, need to extract this as common stuff ...

packages/client/src/client/components/TabBar/TabBar.tsx
```diff
@@ -36,7 +36,10 @@ export const TabBar = <T extends string>({
   doNotShowDropdown,
 }: TabBarProps<T>) => (
   <Background>
-    <LeftSection doNotShowDropdown={items.length < 2 || !!doNotShowDropdown}>
+    <LeftSection
+      data-testid="tab-bar"
```
algreasley, line 40: should be able to remove the `data-testid` on the individual tabs now?
when making changes like the above, we need to cast around to see if that affects things, and whether we can clean up code in the near vicinity ..

packages/client/src/client/components/TabBar/TabBar.tsx
```diff
@@ -75,7 +78,7 @@ export const TabBar = <T extends string>({
             </Action>
           ))}
       {handleItemOnClick && !doNotShowDropdown && (
-        <DropdownWrapper>
+        <DropdownWrapper data-testid="nav-dropdown">
```
algreasley, line 81: this test ID is too vague .. needs to be `rfq-filter-dropdown` or similar .. and like I was suggesting, we can start to use Locators, per alpaca-reference-ui .. plenty of refactoring to do around E2E (you should be happy, refactoring is fun 😁)

packages/client/e2e/fixtures.ts
```diff
@@ -143,7 +143,7 @@ export const test = base.extend<Fixtures>({
         mainPage: mainWindow,
         blotterPO: new CreditBlotterPageObject(mainWindow),
         newRfqPO: new CreditNewRfqPageObject(mainWindow),
-        rfqsPO: new CreditRfqTilesPageObject(mainWindow),
+        rfqsPO: new CreditRfqTilesPageObject(mainWindow, workerInfo),
```
algreasley, line 146: I am strongly inclined to pass `workerInfo` into POs as a standard .. we will need this kind of thing elsewhere (FX and mobile, I'd think)

packages/client/e2e/fixtures.ts
```diff
@@ -141,9 +141,9 @@ export const test = base.extend<Fixtures>({
 
       pages = {
```
algreasley, line 142: If you replace the whole block with:

```suggestion
    let mainPage, blotterPage, newRfqPage, rfqsPage
    const contextPages = context.pages()

    if (isOpenFin(workerInfo)) {
      Object.keys(creditOpenfinUrlSuffixes).forEach((urlPath) => {
        const page = contextPages.find(
          (p) => p.url() === `${process.env.E2E_RTC_WEB_ROOT_URL}/${urlPath}`,
        )

        if (!page) throw Error(`Openfin page at ${urlPath} was not found`)

        switch (urlPath) {
          case "openfin-window-frame?app=CREDIT":
            mainPage = page
            break
          case "credit-blotter":
            blotterPage = page
            break
          case "credit-new-rfq":
            newRfqPage = page
            break
          case "credit-rfqs":
            rfqsPage = page
            break
          default:
            throw Error(`Unknown Openfin page URL - ${urlPath}`)
        }
      })
    } else {
      mainPage =
        blotterPage =
        newRfqPage =
        rfqsPage =
          contextPages.length > 0 ? contextPages[0] : await context.newPage()
    }

    use({
      mainPage,
      blotterPO: new CreditBlotterPageObject(blotterPage, workerInfo),
      newRfqPO: new CreditNewRfqPageObject(newRfqPage, workerInfo),
      rfqsPO: new CreditRfqTilesPageObject(rfqsPage, workerInfo),
    })
```

packages/client/src/client/components/TabBar/TabBar.tsx
algreasley: done .. but a bunch of knock-on effects in unit tests and FX specs .. shows it was a bit brittle, relying on tabs always being there!

algreasley (approved):

PR #2362, opened 2025-03-17
fix: stop bubble graph duplicating on reconnect

SHolleworth:
Seemed to just need tear down logic for deleting the graph


algreasley (approved):
I am happy it works, but confused about why this is affecting us now .. was the theming anything to do with it? or does it go back to the `d3` upgrade I did months ago??

PR #2368, opened 2025-03-19
Chore/fx e2e test updates

umarkayani-adaptive:


packages/client/e2e/pages/FxNewRfqPageObject.ts
algreasley, line 1: rename to FxTilesPageObject

packages/client/e2e/pages/FxNewRfqPageObject.ts
```diff
@@ -0,0 +1,31 @@
+import { Page, TestInfo } from "@playwright/test"
+
+export class FxNewRfqPageObject {
+  constructor(
+    readonly page: Page,
+    readonly testInfo: TestInfo,
+  ) {}
+
+  async clickCurrency(currency: string) {
```
algreasley, line 9: as credit, 
```suggestion
  async selectFilter(filter: string) {  
```

packages/client/e2e/pages/FxNewRfqPageObject.ts
```diff
@@ -0,0 +1,31 @@
...
+    return await this.page
+      .getByTestId("tab-bar-tabs")
+      .getByText(currency)
+      .click()
+  }
+
+  get spotTileNotionalInput() {
+    return this.page.locator("input[id='notional-input-EURUSD']")
```
algreasley, line 17: this includes a specific currency pair
pass this in perhaps as a `currencyPair` param?

algreasley: as you have done below .. but please use `currencyPair`, no need to abbreviate to `pair`

packages/client/e2e/pages/FxNewRfqPageObject.ts
```diff
@@ -0,0 +1,31 @@
...
+  }
+
+  async buyCurrencyPair(pair: string) {
```
algreasley, line 20: no need for the `CurrencyPair` suffix here, we cannot buy anything else! 😄

algreasley: and having the parameter named as `currencyPair` seals the deal!

packages/client/e2e/pages/FxNewRfqPageObject.ts
```diff
@@ -0,0 +1,31 @@
...
+    return await this.page.getByTestId(`Buy-${pair}`).click()
+  }
+
+  async sellCurrencyPair(pair: string) {
+    return await this.page.getByTestId(`Sell-${pair}`).click()
+  }
+
+  get tradeId() {
```
algreasley, line 28: I am leaning towards the pattern of having all the getters at the top .. happy to discuss this pattern, tho

packages/client/e2e/pages/index.ts
```diff
@@ -2,3 +2,6 @@ export * from "./CreditBlotterPageObject"
 export * from "./CreditNewRfqPageObject"
 export * from "./CreditRfqTilesPageObject"
 export * from "./CreditSellSidePageObject"
+// FX pages
```
algreasley, line 5: no need for this comment .. the names say it all

packages/client/e2e/utils.ts
```diff
@@ -15,6 +15,11 @@ export enum TestTimeout {
   EXTENDED = 90000,
 }
 
+export enum RfqTimeout {
```
algreasley, line 18: we have Credit, with RFQs .. this enum name is potentially confusing
also, no need to move these here .. they are not general timeouts .. I specifically moved/kept them in the spot tile spec so they are located in the relevant scope

packages/client/e2e/pages/FxNewRfqPageObject.ts
```diff
@@ -0,0 +1,31 @@
...
+
+export class FxNewRfqPageObject {
+  constructor(
+    readonly page: Page,
+    readonly testInfo: TestInfo,
+  ) {}
+
+  async clickCurrency(currency: string) {
+    return await this.page
+      .getByTestId("tab-bar-tabs")
```
algreasley, line 11: this is where a Component would be useful, to tie in with reused UI elements - ref our conversation about the [RTC-QA-BUG-VERSION branch](https://github.com/AdaptiveConsulting/ReactiveTraderCloud/compare/master...RTC-QA-BUG-VERSION)

packages/client/e2e/pages/FxNewRfqPageObject.ts
algreasley: but should not be marked `async` - neither should the originals in the Credit POs - did we discuss this already?

algreasley (dismissed):

packages/client/e2e/pages/FxNewRfqPageObject.ts
algreasley: done

packages/client/e2e/pages/FxNewRfqPageObject.ts
algreasley: defer adding Components until later

packages/client/e2e/fx-blotter.spec.ts
```diff
@@ -13,10 +12,8 @@ const getTradeIDFromCSV = (csvRows: string | string[]): string => {
 }
 
 test.describe("Trade Blotter", () => {
-  test.afterAll(async ({ fxPages: { blotterPO } }, workerInfo) => {
-    if (isOpenFin(workerInfo)) {
-      await blotterPO.filterButton.click()
-    }
+  test.afterAll(async ({ fxPages: { blotterPO } }) => {
+    await blotterPO.clearFilters()
```
SHolleworth, line 16: Didn't seem to need to be openfin specific to me

packages/client/e2e/pages/FxBlotterPageObject.ts
```diff
@@ -38,6 +34,15 @@ export class FxBlotterPageObject {
...
   }
 
+  async clearFilters() {
+    for (const button of await this.page
+      .getByTestId("clear-filter-button")
+      .all()) {
+      await button.click()
+    }
+    return
+  }
```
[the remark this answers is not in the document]
algreasley: ok, I am always wary of helpful catch-all things, and any conditionals (which this is .. in a way) .. as that is an anti-pattern of E2E - we should _always_ be deliberately setting up tests and tearing down in a controlled way - there should be no room for doubt, variance etc.
.. unless our UI is designed to behave randomly, such as waiting for the quotes .. we don't know which one is coming first, so we have to be flexible in that case ...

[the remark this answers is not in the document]
SHolleworth: Hmm, so just a single click then?

packages/client/e2e/fx-blotter.spec.ts
```diff
@@ -183,5 +145,7 @@ test.describe("Trade Blotter", () => {
 
     expect(typeof filteredCSVRows).not.toBe("string")
     expect(getTradeIDFromCSV(filteredCSVRows)).toBe(tradeIDToSearch)
+
+    await blotterPO.clearFilter()
```
SHolleworth, line 149: It looks like afterAll (sometimes?) refreshed the blotter, resetting the blotter state before clearFilter, causing the hang and timeout. So I think (hope) moving it to the end of the test should fix that issue

packages/client/e2e/pages/FxBlotterPageObject.ts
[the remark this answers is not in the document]
algreasley: if we only setup for and expect a single filter button, yes

packages/client/e2e/fx-blotter.spec.ts
algreasley: that's odd .. I really don't like cleanup in the tests, in general .. perhaps we have to swallow it this time .. it worked before tho?

algreasley: restored to original usage ..

SHolleworth (approved):

PR #2369, opened 2025-03-20
Replace Gap and FlexBox with new spacing components

SHolleworth:


packages/client/src/client/App/Credit/CreditRfqs/CreditRfqCards/styled.tsx
```diff
@@ -76,11 +76,7 @@ export const NoRfqsWrapper = styled.div`
 
 // Card Footer
```
algreasley, line 77: started off taking against this pointless-looking comment .. 😂 
.. then found `AcceptedCardState` is not used, and the final item `QuantityTypography` is not part of the footer .. 
I used to have a mode set in VS Code where it annotated the usage count of each identifier .. got annoying, but might be useful when refactoring, like this ..?

packages/client/src/client/App/Footer/ContactUsButton/ContactUs.tsx
```diff
@@ -28,10 +29,10 @@ export const ContactUs = () => (
       onClick={() => {
         window.open(`mailto:${EMAIL}`)
       }}
+      paddingBottom="xs"
     >
-      <Typography>{EMAIL}</Typography>
+      {EMAIL}
```
algreasley, line 34: minor point - what is the point of `Link` here .. I'd rather just have the plain Typography element inline - like this, it makes me think that Link has some "special powers" 🙂

packages/client/src/client/App/Footer/ContactUsButton/FollowUs.tsx
```diff
@@ -17,10 +16,10 @@ const Row = ({
       variant="Text sm/Regular underlined"
       color="Component colors/Utility/Blue dark/utility-blue-dark-500"
       onClick={onClick}
+      paddingBottom="md"
```
algreasley, line 19: ^ ^ can we get some typing on the color, do you think?
I know it has to allow "red", "white" and "inherit" and wotnot .. so not that easy .. but perhaps "csstype"'s Color | UISK theme Color ?
.. seems the base `div` component's def of `color` takes precedence - may not be worth the effort, but it would be  nice to have completion, I would think ..

